PR #30, opened 2019-04-26
Better Presence - WIP

tschellenbach:


src/channel.js
```diff
@@ -723,6 +723,22 @@ export class Channel {
 	}
 
 	_initializeState(state) {
+		// add the Users
+		console.log('state', state);
+		if (state.members) {
+			for (const m of state.members) {
+				console.log('m', m);
+				this._client.state.updateUserReference(m.user, this.cid);
```
tbarbugli: poor man encapsulation here, please use `this.getClient()` instead of `this._client` (the getter is used to do sanity checks eg. check that the channel is not dandling after a client.disconnect)

src/client_state.js
```diff
@@ -8,11 +8,29 @@ export class ClientState {
...
 			this.users = this.users.set(user.id, Immutable(user));
 		}
 	}
+
+	updateUserReference(user, channelID) {
+		this.updateUser(user);
+		if (!this.userChannelReferences[user.id]) {
+			this.userChannelReferences[user.id] = [];
+		}
+		this.userChannelReferences[user.id].push(channelID);
```
tbarbugli: I might be paranoid here but perhaps we should err on safe side and go for a set instead of array (eg. Object.Keys)

```
if (!this.userChannelReferences[user.id]) {
	this.userChannelReferences[user.id] = {};
}
this.userChannelReferences[user.id][channelID] = true;
```

tbarbugli: looking at `_initializeState` this might actually be a good idea (updateUserReference is called on each member and watcher)

src/client_state.js
```diff
@@ -8,11 +8,29 @@ export class ClientState {
 		// show the status for a certain user...
 		// ie online, offline etc
 		this.users = Immutable({});
+		// store which channels contain references to the specified user...
+		this.userChannelReferences = {};
+	}
+
+	updateUsers(users) {
+		for (const user of users) {
```
tbarbugli: bit more consistent to re-use updateUser here (the user null check)

```users.map(u => this.updateUser(u)```

tschellenbach:
Sounds good

src/channel.js
tschellenbach: done, but the disconnect flow for channels seems broken

src/client_state.js
tschellenbach: done

src/client_state.js
tschellenbach: done, but.... i dont get why we get null users, is this a backend bug?

PR #32, opened 2019-05-01
[CHAT-275]ip based ban

thesyncim:
- [x]  restore array match query tests
- [x]  ensure user_id is present when banning server side
- [x] add basic test to ensure that we don't ban our own IP


Bogdan-D:
outdated pr

PR #40, opened 2019-05-10
fix reactions in threads

jaapbakker88:
# Submit a pull request
## CLA

Reactions in threads were not managed in threads. This PR should fix that.

### DIFF
--- a/src/channel_state.js
+++ b/src/channel_state.js
@@ -94,28 +94,29 @@ export class ChannelState {
 		}
 	}
 
-	addReaction(reaction, reaction_counts, message) {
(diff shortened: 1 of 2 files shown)


src/channel_state.js
```diff
@@ -94,28 +94,29 @@ export class ChannelState {
 		}
 	}
 
-	addReaction(reaction, reaction_counts, message) {
+	addReaction(reaction, message) {
 		const { messages } = this;
+		if (!message) return;
```
jaapbakker88: Somehow it doesn't work without this line. Seems the function is called twice but only the second time it has `message`

PR #109, opened 2019-09-06
Fix type check for readable stream

ferhatelmas:
# Submit a pull request

## CLA

Check was always true due to expression is evaluated first then `typeof` was returning `boolean` which true for the main boolean expression.

Added a guard for null check too. Since `typeof null === object` will return true which would break on property checks.

Other change is irrelevant - comes from prettier. 

I didn't add a test since it wasn't exported, if you guide me how to test it in terms of code layout, I'm happy to add more confidence.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -24,9 +24,10 @@ import pkg from '../package.json';
 
 function isReadableStream(obj) {
 	return (
+		obj !== null &&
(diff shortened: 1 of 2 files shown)


tbarbugli:
that's a trick one, thank you for the PR!

PR #140, opened 2019-10-07
add custom commands into client and crud test

ravlio:
# Submit a pull request

## CLA


src/client.js
```diff
@@ -1202,6 +1202,26 @@ export class StreamChat {
 		return this.get(this.baseURL + `/channeltypes`);
 	}
 
+	createCommand(data) {
```
vishalnarkhede: We need to update typescript [file](https://github.com/GetStream/stream-chat-js/blob/master/types/stream-chat/index.d.ts) for these changes. If you familiar with typescript, please do it ... or let me know. I will update it once this PR is approved.

test/http_mock/index.js
```diff
@@ -0,0 +1,70 @@
+const express = require('express');
+const app = express();
+app.use(express.json());
+module.exports = {
+	start() {
```
vishalnarkhede: What do we need this server for exactly? (just want to know)

ravlio: I use it for the actionHandler stub (our new feature for custom commands). client sendMessage -> chat server actionHandler->customer Server->chat server. So I'm mocking *customer Server*

vishalnarkhede: Ok can you please add that description as comment somewhere in file? Also we should stop the server after corresponding tests are executed. I don't see `stop()` getting called anywhere (unless I am missing something)

PR #160, opened 2019-10-17
Add test for roles

Bogdan-D:
# Submit a pull request

## CLA

### DIFF
--- a/test/serverside.js
+++ b/test/serverside.js
@@ -1963,6 +1963,46 @@ describe('Channel types', function() {
 				descending: true,
 			});
 		});
+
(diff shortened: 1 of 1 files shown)


test/serverside.js
```diff
@@ -1963,6 +1963,46 @@ describe('Channel types', function() {
 				descending: true,
 			});
 		});
+
```
thesyncim: a nice test is copy the default messaging permissions to a new custom channel type and expect the same behaviour

Bogdan-D: the test is validates exactly it should: when roles are empty it should fail, when owner is present it's ok

PR #188, opened 2019-11-21
Chat 879 system messages 2

ravlio:
Allow passing a message when:

promoting  / demoting moderators
ban / unban a user (API call only, not for /ban /unban commands)
invite user
accept invite
reject invite


vishalnarkhede:
Is there any update on this PR? Is it blocked because of something?

PR #190, opened 2019-11-21
add test for deleteUser

ravlio:
Validation error is missing; we should also have validate tags for those fields

Sentry Issue: CHAT-FS

Stacktrace (most recent call first):

  File "github.com/GetStream/chat/state/pg_redis_state.go", line 1043, in (*PGRedisState).DeleteUser
    return nil, stack.Error("Can't do a hard delete without specifying mark_messages_deleted")
  File "github.com/GetStream/chat/controllers/delete_user.go", line 60, in (*DeleteUser).Handle
    removed, err := state.DeleteUser(r.Context(), chat.Auth().App(), userToRemove, request.MarkMessagesDeleted, request.HardDelete)
  File "github.com/GetStream/chat/server/server.go", line 474, in (*Server).MakeHandler.func1
    response, err := handler.Handle(ctx, w, r.WithContext(handleCtx))
  File "github.com/GetStream/chat/server/cors.go", line 42, in CorsPreFlightMiddleware.func1
    })
  File "github.com/GetStream/chat/utils/panic_recovery.go", line 23, in PanicRecoverHandler.func1
    }
  File "github.com/GetStream/chat/server/wrapper.go", line 28, in (*netHTTPResponseWriter).StatusCode
    if w.statusCode == 0 {
  File "github.com/GetStream/chat/server/server.go", line 994, in (*Server).serveConn
    atomic.AddUint32(&s.httpConns, ^uint32(0))
  File "github.com/GetStream/chat/server/server.go", line 1027, in (*Server).serveHTTP
    s.httpServer.ServeConn(conn)
...
(11 additional frame(s) were not displayed)

Can't do a hard delete without specifying mark_messages_deleted

### DIFF
--- a/test/gdpr.js
+++ b/test/gdpr.js
@@ -398,7 +398,7 @@ describe('GDPR endpoints', function() {
 			expect(state.messages[0].text).to.equal('hi');
 		});
 
-		it('hard delete a user without specifying mark_messages_deleted shoud fail', async function() {
+		it('hard delete a user without specifying mark_messages_deleted should fail', async function() {
 			// setup
 			const userID = uuidv4();
 			const creatorID = uuidv4();
--- a/test/gdpr.js
+++ b/test/gdpr.js
@@ -14,6 +14,7 @@ import {
 	createUsers,
 	runAndLogPromise,
 	sleep,
+	expectHTTPErrorCode,
 } from './utils';
 import uuidv4 from 'uuid/v4';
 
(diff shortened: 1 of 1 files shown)


test/gdpr.js
```diff
@@ -397,6 +398,27 @@ describe('GDPR endpoints', function() {
 			expect(state.messages[0].text).to.equal('hi');
 		});
 
+		it('hard delete a user without specifying mark_messages_deleted shoud fail', async function() {
```
ferhatelmas: typo `should`

PR #194, opened 2019-11-28
support client.channel with arity 2 and with ID null and undef

tbarbugli:


test/channels.js
```diff
@@ -23,6 +27,72 @@ Promise.config({
...
+		expect(channel.cid).to.eql('messaging:123');
+		expect(channel.id).to.eql('123');
+		expect(channel.data).to.eql({ cool: true });
+		done();
+	});
+
+	it('default options', function(done) {
+		const channel = client.channel('messaging', '123');
+		expect(channel.cid).to.eql('messaging:123');
+		expect(channel.id).to.eql('123');
```
ferhatelmas: why not check data here too ?

test/channels.js
```diff
@@ -23,6 +27,72 @@ Promise.config({
...
+		done();
+	});
+
+	it('null ID no options', function(done) {
+		const channel = client.channel('messaging', null);
+		expect(channel.id).to.eq(undefined);
```
ferhatelmas: and here too ?

PR #195, opened 2019-11-28
CHAT-832 raise clear errors when trying to build tokens without secret or on browser

tbarbugli:


src/client.js
```diff
@@ -374,6 +374,9 @@ export class StreamChat {
 	 * @return {string} Returns a token
 	 */
 	createToken(userID, exp) {
+		if (this.secret == null) {
```
ferhatelmas: isn't it better to check falsy values?

src/signing.js
```diff
@@ -22,6 +22,13 @@ export function JWTUserToken(apiSecret, userId, extraData = {}, jwtOptions = {})
 		...extraData,
 	};
 
+	// make sure we return a clear error when jwt is shimmed (ie. browser build)
+	if (jwt == null || jwt.sign == null) {
+		throw Error(
+			`Unable to find jwt crypto, if you are getting this error is probably because you are trying to generate tokens on browser or React Native (or other environment where crypto functions are not available). Please Note: token should only be generate server-side.`,
```
ferhatelmas: would be nice if wrapped.

ferhatelmas:
```diff
- should only be generate
+ should only be generated
```

ferhatelmas: `if you are getting this error is` could be dropped, IMHO

src/client.js
tbarbugli: not needed in this case (you can see that in the client constructor)

PR #196, opened 2019-11-28
CHAT-888 add test for orphaned channels

tbarbugli:


test/serverside.js
```diff
@@ -2351,8 +2351,21 @@ describe('Channel types', function() {
 				await tClient.channel(typeName, 'general').watch();
 			});
 
-			it('create a channel of the new type', async function() {
-				await expectHTTPErrorCode(400, client.deleteChannelType(typeName));
+			it('delete a channel type with existing channels should fail', async function() {
+				await client.channel(typeName, 'general').delete();
```
Bogdan-D: how does it check it's fail?

tbarbugli: no idea what you mean :)

Bogdan-D: I mean that after this change the test doesn't actually testing that deleting is failing

PR #201, opened 2019-12-03
Return user.banned with the user payload

ravlio:
This is for the queryUser endpoint only. Feature request from Imgur (#49 from their issue tracker)


test/test.js
```diff
@@ -1581,6 +1582,7 @@ describe('Chat', function() {
 				};
 
 				users[i] = (await serverAuthClient.updateUser(user)).users[username(i)];
+				users[i].banned = false;
```
Bogdan-D: this line looks wrong

PR #207, opened 2019-12-17
Remove some unnecessary

ferhatelmas:
* unused import (imported but not used)
* constant condition (variable is already type checked)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -698,7 +698,7 @@ export class StreamChat {
 			},
 		);
 		this.connectionID = this.wsConnection.connectionID;
-		const cids = Object.keys(this.activeChannels || {});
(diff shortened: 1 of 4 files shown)


src/client.js
```diff
@@ -698,7 +698,7 @@ export class StreamChat {
 			},
 		);
 		this.connectionID = this.wsConnection.connectionID;
-		const cids = Object.keys(this.activeChannels || {});
+		const cids = Object.keys(this.activeChannels);
```
Bogdan-D: what if `activeChannels` is null? It will throw exception in such case and client side code will stop working

ferhatelmas: If it's missing, it will crash two lines below anyway (line 703).

According to the call path, it can't be null - always initialized.

PR #211, opened 2019-12-31
Modify return type of setUser to be consistent with the vanilla js

setzamora:
Modify return type of setUser to be consistent with the vanilla js and documentation.

Unless this is the way forward considering it's a setter, the setUser function is inconsistent with the vanilla js one.

### DIFF
--- a/types/stream-chat/api-response-tests/data.ts
+++ b/types/stream-chat/api-response-tests/data.ts
@@ -1,5 +1,5 @@
 import {
-  ConnectAPIReponse,
+  ConnectAPIResponse,
   SearchAPIResponse,
   UpdateUsersAPIResponse,
   BanUserAPIResponse,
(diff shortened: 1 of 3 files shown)


tbarbugli:
@vishalnarkhede `client.setUser` returns a promise

tbarbugli:
@vishalnarkhede mapping the response type instead of just using `Object` would be awesome (the response has some very useful information)

setzamora:
Source updated to use the `ConnectAPIResponse` type

tbarbugli:
@vishalnarkhede this is a very good occasion to update the Event type since things changed (we added new events and more fields).

I also do not think it makes sense to say setUser returns a `Event` object; we should create a new response type ad-hoc (Event has way too many optional fields)

ferhatelmas:
@setzamora we dropped auto-generated data file from git in #222, could you update your PR? If any help is needed, let me know for a small bump.

ferhatelmas:
@setzamora we will need these changes so I will rebase and merge. Thanks for your contribution.

ferhatelmas:
Merging via #223.

PR #213, opened 2020-01-03
CRNS-4: Updating dev token method (for compatibility with RN)

vishalnarkhede:
# Submit a pull request

## CLA

If devToken is used on React Native, then `window.btoa` is not available (for obvious reasons). Thus we need some alternative ... hence adding a common method compatible with all clients.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   "dependencies": {
     "@babel/runtime": "^7.3.1",
     "axios": "^0.18.1",
+    "base64-js": "^1.3.1",
     "chai-arrays": "^2.0.0",
     "cross-fetch": "^3.0.0",
     "form-data": "^2.3.3",
--- a/src/base64.js
+++ b/src/base64.js
@@ -1,0 +1,57 @@
+const base64 = require('base64-js');
+
+// source - https://github.com/beatgammit/base64-js/blob/master/test/convert.js#L72
+const map = (arr, callback) => {
+	const res = [];
+	let kValue, mappedValue;
+
+	for (let k = 0, len = arr.length; k < len; k++) {
+		if (typeof arr === 'string' && !!arr.charAt(k)) {
+			kValue = arr.charAt(k);
+			mappedValue = callback(kValue, k, arr);
+			res[k] = mappedValue;
+		} else if (typeof arr !== 'string' && k in arr) {
+			kValue = arr[k];
+			mappedValue = callback(kValue, k, arr);
+			res[k] = mappedValue;
+		}
+	}
+	return res;
+};
+
+export function encodeBase64(data) {
+	return base64.fromByteArray(
+		map(data, function(char) {
+			return char.charCodeAt(0);
+		}),
+	);
+}
+
+// base-64 decoder throws exception if encoded string is not padded by '=' to make string length
+// in multiples of 4. So gonna use our own method for this purpose to keep backwards compatibility
+// https://github.com/beatgammit/base64-js/blob/master/index.js#L26
+export function decodeBase64(s) {
+	const e = {},
+		w = String.fromCharCode,
+		L = s.length;
+	let i,
+		b = 0,
+		c,
+		x,
+		l = 0,
+		a,
+		r = '';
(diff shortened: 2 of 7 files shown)


ferhatelmas:
@vishalnarkhede I think copying code is bad since we won't get changes to the copied code. 

Isn't it better to add `buffer` as a dep into RN since the previous implementation is cleaner. Unless this is possible, adding `base64` as a dep here.

vishalnarkhede:
> @vishalnarkhede I think copying code is bad since we won't get changes to the copied code.
> 
> Isn't it better to add `buffer` as a dep into RN since the previous implementation is cleaner. Unless this is possible, adding `base64` as a dep here.

@ferhatelmas My thoughts behind it!!

1. I don't think we are expecting regular changes in encoding.
2. We can definitely add dependency to our RN sdk, But people don't necessarily use our RN sdk. They can have their own react-native app (without using our sdk)
3. `Buffer` is a core node api, so installing it will come with loads of other dependencies, which i don't think we really need. These are quite heavy apis.

ferhatelmas:
> @ferhatelmas My thoughts behind it!!

Thanks for the detailed explanation.

> I don't think we are expecting regular changes in encoding.

I wish so but the package is at v0.1.0 so not stable, I believe. Also, there are better packages like [universal-base64](https://www.npmjs.com/package/universal-base64).

> We can definitely add dependency to our RN sdk, But people don't necessarily use our RN sdk. They can have their own react-native app (without using our sdk)

Then, it should be their responsibility to sign the token as they want.

> Buffer is a core node api, so installing it will come with loads of other dependencies, which i don't think we really need. These are quite heavy apis.

Yes, (~7KB gzipped), that's why it's the less preferred solution. My suggestion is adding a dependency and using its api, instead of copying its source. 

See the above suggested package, for example but bynens' base64 or something else doesn't matter much as soon as we can update easily on changes.

vishalnarkhede:
@ferhatelmas [universal-base64](https://www.npmjs.com/package/universal-base64) won't work for our purpose. If you check the source code, it's doing exactly what we were doing so far - `btoa` for browser and `Buffer` for rest.

And the reason I went for copying the function is because of this performance issue issue raised on github repo - https://github.com/mathiasbynens/base64/issues/19

This issue is fixed in this PR (which is not merged yet) - https://github.com/mathiasbynens/base64/pull/20

So I have used a code from this PR. I think then instead of adding it directly as dependency, we can create a fork of it and publish NPM package on our end for the same. But I am just not sure if its that necessary!!

What do you think? Also @tbarbugli

ferhatelmas:
This exchange is educating for me, thanks. We're iterating and finding the best solution.

I think it's unlikely to expect that PR to be merged in the current state because it is simply rewriting the whole source (fixes indentation, uses newer js, breaks old module syntax instead of fixing only the bug and there is an error in the implementation as you commented)

In this PR, my concern is copying. Two drawbacks of copying I see:
* harder to get updates from upstream (forks somehow better because there is no other surrounding irrelevant code at least)
* multiple copies in our projects (js and rn) and changes might diverge 

Your suggestion sounds better to me since it resolves the second point of mine. When upstream is updated, we can update one line in package.json and good to go. 

On the other hand, to resolve the first point too. You might check other packages that work with RN by default. The package I mentioned was just an example. Since it's a core piece, there should be something (or we can even have ours).

tbarbugli:
@vishalnarkhede why not https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/Libraries/Utilities/binaryToBase64.js ?

vishalnarkhede:
@tbarbugli @ferhatelmas Done!!

ferhatelmas:
@vishalnarkhede it seems this PR does multiple things in addition to base64. Can we extract them?

vishalnarkhede:
@ferhatelmas those other changes are very small.  Don't think they are worth going through trouble of separating !!

ferhatelmas:
@vishalnarkhede I think yes for better tracking.

For example, you fixed one of the typescript issues but it's actually a completely different fix and in its fix PR, it should close #216.

PR #222, opened 2020-01-13
Run dts on travis and drop autogenerated data file

ferhatelmas:
* drop data file since it's autogenerated
* run correct dts line and drop duplicated script
* silence tests until it's clean


ferhatelmas:
@jaapbakker88 Failure from dts will be fixed separately on a follow up.

PR #224, opened 2020-01-14
More type test cleanup

ferhatelmas:
* drop no lint that prevents prettier run on test file
* also run dtslint on test file
* drop unused code


ferhatelmas:
@jaapbakker88 good to create a new release now. There will be two more improvements but we can release again when they are done.

PR #225, opened 2020-01-14
Prevent travis test auto run

ferhatelmas:
Change install to script so that travis does only what we want.


ferhatelmas:
@jaapbakker88 merging this since minor.

PR #236, opened 2020-01-20
add tests for message.id using client/server side auth.

thesyncim:
fix old tests

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1433,9 +1433,29 @@ describe('Chat', function() {
 				const message = {
 					id,
 					text: 'yo',
+					user_id: uuidv4(),
 				};
-				await channel.sendMessage(message);
-				await expectHTTPErrorCode(400, channel.sendMessage(message));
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -1433,9 +1433,29 @@ describe('Chat', function() {
...
+				let serverChannel = client.channel(channel.type, channel.id);
+				await serverChannel.sendMessage(message);
+				await expectHTTPErrorCode(
+					400,
+					serverChannel.sendMessage(message),
+					`StreamChat error code 4: SendMessage failed with error: "a message with ID ${id} already exists"`,
+				);
+			});
+
+			it('Add a chat message with ID client side should fail', async function() {
```
vishalnarkhede: @thesyncim we generate id on client side in react and RN, before sending a message. Are you sure it should fail?

https://github.com/GetStream/stream-chat-react/blob/master/src/components/Channel.js#L335

PR #245, opened 2020-01-27
fix types for channel.sendReaction

jaapbakker88:
# Submit a pull request

## CLA

### DIFF
--- a/types/stream-chat/index.d.ts
+++ b/types/stream-chat/index.d.ts
@@ -113,6 +113,9 @@ export interface Reaction {
   [propName: string]: any;
 }
 
+export interface ReactionRequest {
+  type: string;
+}
 export interface ReactionResponse extends Reaction {
   created_at: string;
   updated_at: string;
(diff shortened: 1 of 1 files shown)


types/stream-chat/index.d.ts
```diff
@@ -113,6 +113,9 @@ export interface Reaction {
   [propName: string]: any;
 }
 
+export interface ReactionRequest {
+  type: string;
```
tbarbugli: reactions have more fields, @Bogdan-D please help here :)

jaapbakker88: This is just for sending a reaction though. I was working on the typescript stuff and it was failing because we reused the Reaction response object for this.

vishalnarkhede: @jaapbakker88  Which properties of `Reaction` were breaking? I can see `message_id`, `user` as required properties on `Reaction` type. They are used for request using server side token. So maybe we can make them optional!!

jaapbakker88: @vishalnarkhede updated to use only the `type` as a required property. It was indeed breaking on the rest.

jaapbakker88:
@vishalnarkhede updated this

vishalnarkhede:
Looks good now :)

PR #251, opened 2020-02-04
Adding some more logs for investigation

vishalnarkhede:
# Submit a pull request

## CLA

This is part of investigation of support ticket #1805

For some reason, our client tried to reconnect even though we just disconnected our ws connection. My suspicion is that call to ws.close results in error causing disconnect to fail. But to confirm this, we need some logs.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -138,9 +138,25 @@ export class StableWSConnection {
 			const { ws } = this;
 			isClosedPromise = new Promise(resolve => {
 				ws.onclose = () => {
+					this.logger(
+						'info',
+						`connection:disconnect() - resolving isClosedPromise`,
+						{
+							tags: ['connection'],
+						},
+					);
 					resolve();
 				};
 			});
+
+			this.logger(
+				'info',
+				`connection:disconnect() - Manually closed connection by calling client.disconnect()`,
+				{
+					tags: ['connection'],
+				},
+			);
+
 			ws.close(1000, 'Manually closed connection by calling client.disconnect()');
 		}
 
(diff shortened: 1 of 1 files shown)


src/connection.js
```diff
@@ -138,9 +138,25 @@ export class StableWSConnection {
 			const { ws } = this;
 			isClosedPromise = new Promise(resolve => {
 				ws.onclose = () => {
+					this.logger(
+						'info',
+						`connection:disconnect() - resolving isClosedPromise`,
```
ferhatelmas: first 2 logs aren't consistent with the rest.

ferhatelmas:
What is the drawback of adding all these logs? Do we always need to have them on?

vishalnarkhede:
@ferhatelmas No, one should'nt have them on all the time. These logs just help us our clients to provide us with some information, which in return helps us debug the issue.

src/connection.js
vishalnarkhede: which once? If you can point me to exact log please...

ferhatelmas: Maybe I am familiar with it since I see only the diff.

This one, see separator `-` vs `:`

src/connection.js
```diff
@@ -138,9 +138,25 @@ export class StableWSConnection {
...
+							tags: ['connection'],
+						},
+					);
 					resolve();
 				};
 			});
+
+			this.logger(
+				'info',
+				`connection:disconnect() - Manually closed connection by calling client.disconnect()`,
```
ferhatelmas: and this one

ferhatelmas:
> These logs just help us our clients to provide us with some information, which in return helps us debug the issue.

So could you confirm if flow is like following:
* user enables logger
* runs their code
* passed the console output to us

vishalnarkhede:
> > These logs just help us our clients to provide us with some information, which in return helps us debug the issue.
> 
> So could you confirm if flow is like following:
> 
> * user enables logger
> * runs their code
> * passed the console output to us

Yes, correct!!

src/connection.js
vishalnarkhede: You are right. Fixed it now :)

PR #265, opened 2020-02-28
Increment wsId when ws is disconnected manually

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -121,8 +121,7 @@ export class StableWSConnection {
 
 		this._removeConnectionListeners();
 
-		// reset the wsID;
-		this.wsID = 1;
+		this.wsID += 1;
 		this.isHealthy = false;
 
 		// remove ws handlers...
--- a/src/connection.js
+++ b/src/connection.js
@@ -450,10 +450,14 @@ export class StableWSConnection {
 	 *
 	 */
 	_errorFromWSEvent = event => {
-		this.logger('error', `connection:onclose() - WS failed with code ${event.code}`, {
-			tags: ['connection'],
-			event,
-		});
+		this.logger(
+			'error',
+			`connection:_errorFromWSEvent() - WS failed with code ${event.code}`,
+			{
+				tags: ['connection'],
+				event,
+			},
+		);
 
 		const error = new Error(`WS failed with code ${event.code}`);
 		error.code = event.code;
--- a/src/connection.js
+++ b/src/connection.js
@@ -111,6 +111,8 @@ export class StableWSConnection {
 			},
 		);
 
+		this.wsID += 1;
+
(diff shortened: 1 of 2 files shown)


src/connection.js
```diff
@@ -121,8 +121,7 @@ export class StableWSConnection {
 
 		this._removeConnectionListeners();
 
-		// reset the wsID;
-		this.wsID = 1;
+		this.wsID += 1;
```
vishalnarkhede: For some wierd reason, when client disconnect ws connection manually,

vishalnarkhede: This is a disconnect function, which one uses when he wants to manually disconnect the websocket connection e.g. when mobile app goes to background.

For some weird reason, ws hits the original [onclose](https://github.com/GetStream/stream-chat-js/pull/265/files#diff-875da84119066ec244742e95d1209f0cL357) callback (with code `1001`) after `ws.close()` (line 159), which should not happen ideally since we override onclose callback on line 140. This is **NOT** a reproducible issue on our side. But some customer is facing this issue. The logs that he provided didn't help figure out why is this happening.

The main problem is - when original onclose callback gets called, it triggers the reconnection logic ... which defeats the whole purpose manually disconnecting the client. The moment user calls `disconnect`, our code immediately triggers reconnect.

To avoid this, instead of reseting the this.wsID to 1, we can increment it, so that original callback function [will not respond anymore](https://github.com/GetStream/stream-chat-js/pull/265/files#diff-875da84119066ec244742e95d1209f0cL364)

tschellenbach: hmm, i think you should also move it up, it should be the first thing in the function in my opinion

src/connection.js
```diff
@@ -111,6 +111,8 @@ export class StableWSConnection {
 			},
 		);
 
+		this.wsID += 1;
```
vishalnarkhede: Repeating the comment (since last one got removed due to updated commit)
===================================

This is a disconnect function, which one uses when he wants to manually disconnect the websocket connection e.g. when mobile app goes to background.

For some weird reason, ws hits the original onclose (Line 357) callback (with code 1001) after ws.close() (line 159), which should not happen ideally since we override onclose callback on line 140. This is NOT a reproducible issue on our side. But some customer is facing this issue. The logs that he provided didn't help figure out why is this happening.

The main problem is - when original onclose callback gets called, it triggers the reconnection logic ... which defeats the whole purpose manually disconnecting the client. The moment user calls disconnect, our code immediately triggers reconnect.

To avoid this, instead of reseting the this.wsID to 1, we can increment it, so that original callback function will not respond anymore( line 364)

ZD ticket - https://getstream.zendesk.com/agent/tickets/1805

PR #276, opened 2020-03-20
Updating removeMessage function to handle threads

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/channel_state.js
+++ b/src/channel_state.js
@@ -252,7 +252,7 @@ export class ChannelState {
 	 */
 	removeMessage(messageToRemove) {
 		let isRemoved = false;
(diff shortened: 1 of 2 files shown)


src/channel_state.js
```diff
@@ -251,23 +251,43 @@ export class ChannelState {
 	 * @return {boolean} Returns if the message was removed
 	 */
 	removeMessage(messageToRemove) {
-		let removed = false;
-		const messages = this.messages.flatMap(message => {
```
vishalnarkhede: We didn't handle the case of threads here. Because of which "cancel" action on giphy command didn't work in our web and mobile sdk.

PR #288, opened 2020-03-31
Populate channel state with the new membership field

tbarbugli:
Add support for upcoming new field from channel responses. @vishalnarkhede if you are happy with this can you take care of typescript defs?

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -981,6 +981,8 @@ export class Channel {
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


vishalnarkhede:
> Add support for upcoming new field from channel responses. @vishalnarkhede if you are happy with this can you take care of typescript defs?

Yup, I will !!

PR #293, opened 2020-04-06
queryChannels fix - Removing watch if connectionId is absent

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -903,6 +903,7 @@ export class StreamChat {
 
 		if (!this._hasConnectionID()) {
 			payload.watch = false;
+			payload.presence = false;
 		}
 
 		const data = await this.get(this.baseURL + '/channels', {
--- a/src/client.js
+++ b/src/client.js
@@ -892,11 +892,6 @@ export class StreamChat {
 
 		// Make sure we wait for the connect promise if there is a pending one
 		await this.wsPromise;
-
-		if (!this._hasConnectionID()) {
-			defaultOptions.watch = false;
-		}
-
 		// Return a list of channels
 		const payload = {
 			filter_conditions: filterConditions,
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -892,11 +892,6 @@ export class StreamChat {
 
 		// Make sure we wait for the connect promise if there is a pending one
 		await this.wsPromise;
-
-		if (!this._hasConnectionID()) {
-			defaultOptions.watch = false;
```
vishalnarkhede, line 897: Customers are  getting plenty of query channels related issue. In our logs, it seems like there are lot of queryChannel calls being made with empty connectionId.  Error for this case on js-client look something  like this:

```js
QueryChannels failed with error: "Watch or Presence require you to specify a user id"
```

Error is a bit misleading. But as I said, real reason is absence of connectionId in request. 

If there is no connectionId, which mean `setUser()` didn't succeed, then we used to remove watch from defaultOptions. But if there is a `watch: true` in options (provided in parameters), then final payload will have `watch: true` as well,  resulting in api failure ... which in turn fails the application itself:

So instead we should simply remove the watch from final payload itself and continue with queryChannels call. We already have reconnection logic in place, once reconnection succeeds, queryChannels will run again with `watch: true`

src/client.js
```diff
@@ -906,6 +901,10 @@ export class StreamChat {
 			...options,
 		};
 
+		if (!this._hasConnectionID()) {
+			payload.watch = false;
```
thesyncim, line 905: we also need to handle
payload.presence = false

thesyncim: presence also requires connection_id

vishalnarkhede: Done!

PR #318, opened 2020-05-04
Forcefully closing the ws connection, if close frame gets delayed

vishalnarkhede:
# Submit a pull request

## CLA

Feeld is facing some issue where ws just keeps waiting for close frame after manually disconnecting the ws. Its something we definitely need to fix on backend (as discussed with @thesyncim ). But we should not infinitely wait for it, in case there are delays. 

So forcefully resolving disconnection promise, if we don't receive close frame from backend within 2 seconds!

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -178,17 +178,26 @@ export class StableWSConnection {
 		const { ws } = this;
 		if (ws && ws.close && ws.readyState === ws.OPEN) {
 			isClosedPromise = new Promise(resolve => {
-				ws.onclose = () => {
+				const onclose = event => {
 					this.logger(
 						'info',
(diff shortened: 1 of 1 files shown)


src/connection.js
```diff
@@ -137,17 +137,26 @@ export class StableWSConnection {
...
+							event,
 						},
 					);
 					resolve();
 				};
+
+				ws.onclose = onclose;
+				// In case we don't receive close frame websocket server in time,
+				// lets not wait for more than 2 seconds.
+				setTimeout(onclose, 2000);
```
vishalnarkhede, line 157: Or maybe just 1 sec. What do you think?

PR #319, opened 2020-05-07
Handling token expiration

vishalnarkhede:
# Submit a pull request

### Static token

```js
client.setUser({id: 'vishal'}, 'token');
```

### Token provider

```js
client.setUser({id: 'vishal'}, async () => {
    // const token = await api.fetchToken();
    return token;
});
```

## Description

- New class - `TokenManager`, which will be a common place to access token from for client and connection class.
- When ws connection fails with error code `40`, the token will be refreshed using tokenProvider function (given to `setUser` function) and connection request will be retried.
- When http request fails with error code `40`, the token will be refreshed using tokenProvider function (given to `setUser` function) and connection request will be retried.

## TODO

- [x] Handle error from ws connection about expired token
- [x] Handle error from http request about expired token
- [x] Handle the case of guest users
- [x] Handle the case of anonymous users
- [x] tests
- [ ] docs

## CLA

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -93,6 +93,14 @@ export class StreamChat {
 		this.configs = {};
 		this.anonymous = false;
 
+		// If its a server-side client, then lets initialize the tokenManager, since token will be
+		// generated from secret.
+		if (this.secret) {
+			this.tokenManager = new TokenManager({
+				secret: this.secret,
+			});
+		}
+
 		/**
 		 * logger function should accept 3 parameters:
 		 * @param logLevel string
@@ -375,6 +383,7 @@ export class StreamChat {
 		if (exp != null) {
 			extra.exp = exp;
 		}
+
 		return JWTUserToken(this.secret, userID, extra, {});
 	}
 
--- a/src/client.js
+++ b/src/client.js
@@ -22,6 +22,8 @@ import fetch, { Headers } from 'cross-fetch';
 import FormData from 'form-data';
 import pkg from '../package.json';
 import Immutable from 'seamless-immutable';
+import { TokenManager } from './token_manager';
+import { isFunction } from './utils';
 
 function isReadableStream(obj) {
 	return (
@@ -29,15 +31,6 @@ function isReadableStream(obj) {
 		typeof obj === 'object' &&
 		typeof obj._read === 'function' &&
 		typeof obj._readableState === 'object'
-	);
-}
-
-function isFunction(value) {
-	return (
-		value &&
-		(Object.prototype.toString.call(value) === '[object Function]' ||
-			'function' === typeof value ||
-			value instanceof Function)
 	);
 }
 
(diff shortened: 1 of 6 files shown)


src/client.js
```diff
@@ -472,93 +468,65 @@ export class StreamChat {
 		});
 	}
 
-	async get(url, params) {
+	doAxiosRequest = async (type, url, data, params = {}) => {
```
vishalnarkhede, line 471: Keeping all the logging and response handling in a single place!

src/client.js
```diff
@@ -195,30 +194,16 @@ export class StreamChat {
...
-		}
+		await this._setToken(user, userTokenOrProvider);
 		this._setUser(user);
 		this.anonymous = false;
 
 		return this._setupConnection();
-	}
+	};
+
+	_setToken = async (user, userTokenOrProvider) => {
```
tbarbugli, line 204: ideally we add some validation for `userTokenOrProvider` just to make sure that you dont pass silly stuff to it

src/client.js
```diff
@@ -323,6 +308,8 @@ export class StreamChat {
 		this.activeChannels = {};
 		// reset client state
 		this.state = new ClientState();
+		// reset token manager
+		this.tokenManager.expire();
```
tbarbugli, line 312: why dont we call this `reset` then?

src/client.js
tbarbugli: isn't pointless to have this as an async fn just to await on `this.tokenManager.setTokenOrProvider`

src/client.js
```diff
@@ -472,93 +465,68 @@ export class StreamChat {
...
-
-			return this.handleResponse(response);
-		} catch (e) {
-			this._logApiError('post', url, e);
 			if (e.response) {
+				if (
+					e.response.data.code === chatCodes.TOKEN_EXPIRED &&
+					!this.tokenManager.isStatic()
+				) {
+					await this.tokenManager.loadToken();
```
tbarbugli, line 502: one simple optimization available here:

```js

function doAxiosRequest() {
   await this.tokenManager.tokenReady(); // return a promise from loadToken (resolved if finished, pending if running)
   // the actual request handling

   if (e.response.data.code === chatCodes.TOKEN_EXPIRED && !this.tokenManager.isStatic()) {
      return await this.doAxiosRequest(type, url, params, data);
   }
}
```

src/client.js
vishalnarkhede: > ideally we add some validation for `userTokenOrProvider` just to make sure that you dont pass silly stuff to it

We have validation in token_manager actually

ferhatelmas:
@vishalnarkhede when I say this way, you might not like much but there are 3 type of changes here:
* syntax due to new proposals
* refactoring for common error handling
* token expiration feature

It would be better if you could extract first two items into its own PR then we're left alone with the feature so that the details can reviewed better.

vishalnarkhede:
@ferhatelmas 

> syntax due to new proposals

??

ferhatelmas:
for example, https://github.com/GetStream/stream-chat-js/pull/319/files#diff-cf27c1d543e886c89cd9ac8b8aeaf05bR158 and also with some formatting changes.

vishalnarkhede:
@ferhatelmas I hear you and I will be careful in next PR. But it's not always possible to isolate every little change. All of them were done to implement token expiration functionality (except for the one that you mentioned in last comment). At this point it would be too much effort to extract them to separate PR!

ferhatelmas:
Thanks. If we slowly adapt keeping changes to be independent, it would be useful when problems are seen and some investigation is needed.

OK, don't worry about this.

PR #321, opened 2020-05-08
[CHAT-976] queryMembers

thesyncim:
# Submit a pull request

## CLA

**introduce a new endpoint that allows clients to filter and search channel members**

example queries:

query all the members for the current channel where user.name='tommaso'
```js
channel.queryMembers({'name':'tommaso'})
```
filter member by id
```js
channel.queryMembers({'id':'tommaso'})
```
query all the members for the current channel where user.name contains
```js
channel.queryMembers({'name':{"$q":'tomm'}})
```
query channel moderators
```js
channel.queryMembers({'is_moderator:true'})
```
query for banned members in channel
```js
channel.queryMembers({'banned:true'})
```

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -161,7 +161,7 @@ export class Channel {
 	/**
 	 * search - Query Members
 	 *
-	 * @param {object|string}  filterConditions object MongoDB style filters
+	 * @param {object}  filterConditions object MongoDB style filters
 	 * @param {object} sort             Sort options, for instance {created_at: -1}
 	 * @param {object} options        Option object, {limit: 10, offset:10}
 	 *
--- a/src/channel.js
+++ b/src/channel.js
@@ -159,6 +159,18 @@ export class Channel {
 	}
 
 	/**
+	 * search - Query Members
+	 *
+	 * @param {object|string}  member search query or object MongoDB style filters
+	 * @param {object} options       Option object, {limit: 10}
+	 *
+	 * @return {object} search messages response
+	 */
+	async queryMembers(query, options = {}) {
+		return await this.getClient().queryMembers(this.cid, query, options);
+	}
+
+	/**
 	 * sendReaction - Send a reaction about a message
 	 *
 	 * @param {string} messageID the message id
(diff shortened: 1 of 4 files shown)


src/channel.js
```diff
@@ -158,6 +158,38 @@ export class Channel {
...
+		if (!options) {
+			options = {};
+		}
+		let id;
+		const type = this.type;
+		let members;
+		if (this.id) {
+			id = this.id;
+		} else if (this.data && Array.isArray(this.data.members)) {
+			members = this.data.members;
```
vishalnarkhede, line 179: Why do we need this? (just curious) @thesyncim

thesyncim: this is needed for member-based channels (channels without ID)

vishalnarkhede: are you referring to one-on-one conversations? Because they also have some id - `"!members-TAMXTZLQ5V6tHNhmyYupyOwNRuWyxShsqdLhxgk7jLc"` Can't we use that?

thesyncim: you could also use only the ID, but that would not work for calls like this one.
`channel('messaging',{members:[a,b,c]}).queryMembers({})` (id is undefined at this point)

types/stream-chat/index.d.ts
```diff
@@ -427,6 +427,12 @@ export class Channel {
   delete(): Promise<DeleteChannelAPIResponse>;
   search(query: string | object, options: object): Promise<SearchAPIResponse>;
 
+  queryMembers(
+    filterConditions: object,
+    sort: object,
+    options: object,
```
vishalnarkhede, line 433:
```suggestion
    sort?: object,
    options?: object,
```

types/stream-chat/index.d.ts
```diff
@@ -671,6 +677,10 @@ export interface UsersAPIResponse extends APIResponse {
   users: UserResponse[];
 }
 
+export interface MembersAPIResponse extends APIResponse {
+  members: Member[];
```
vishalnarkhede, line 681:
```suggestion
  members: ChannelMemberResponse[];
```

PR #330, opened 2020-05-15
Revert "Fix uuid bump issues due to recent PRs"

vishalnarkhede:
Reverts GetStream/stream-chat-js#329

Given version of UUID doesn't work well with react-native - https://github.com/uuidjs/uuid#getrandomvalues-not-supported


ferhatelmas:
@vishalnarkhede why don't we handle it in native side and instead revert this? With this approach, we will stuck at old version.

/cc @mahboubii @jaapbakker88

Issue #239, opened 2020-01-24
File Upload API doesn't work in node.js


kanekv:
There are multiple problems in this piece of code: https://github.com/GetStream/stream-chat-js/blob/master/src/client.js#L564-L579.

1. Buffer check doesn't work properly for buffers created other than fs.createReadStream
2. It's using `File` which doesn't exist in node.js
3. To accept buffers `data.append('file', fileField);` should accept mimetype as 3rd argument.

kanekv:
@ferhatelmas , @jaapbakker88 suggested that you would be the best person to check these issues.

ferhatelmas:
@kanekv thanks for the great details, we will check with @jaapbakker88 soon.

PR #335, opened 2020-05-19
Fixed File interface and improve contentType

mksglu:
File interface is not part of Javascript (node.js); it belongs to Web APIs. So, the File is not available in your node.js application.`sendFile` may be used in different parts on the client-side or in some SDKs. So instead of removing this line, we should add some conditions. And this code actually does accept mime-type as 3rd argument but is in a different case. I think we can use this control a little more flexibly. #239

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -194,11 +194,11 @@ export class StreamChat {
 
 		this.setUserPromise = Promise.all([setTokenPromise, wsPromise])
 			.then(
-				result =>
+				(result) =>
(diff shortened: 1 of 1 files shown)


jaapbakker88:
@mksglu thanks for the pr, can you run `yarn lint-fix` on this to make sure code style is correct.

mahboubii:
@mksglu Thanks for the contribution, this issue is resolved as part of #448

PR #337, opened 2020-05-22
Feature/user deleted event

imtoori:


jaapbakker88:
https://github.com/GetStream/chat/pull/1060

ferhatelmas:
@jaapbakker88 how about types?

jaapbakker88:
> @jaapbakker88 how about types?

don't think types are an issue here.

src/events.js
```diff
@@ -35,6 +35,7 @@ export const EVENT_MAP = {
 	// local events
 	'connection.changed': true,
 	'connection.recovered': true,
+	'user.deleted': true,
```
ferhatelmas, line 38: is this location in the map, right ?

ferhatelmas: ordered

PR #340, opened 2020-05-29
Force resolve wsPromise upon recovery

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -737,6 +737,9 @@ export class StreamChat {
 				type: 'connection.recovered',
 			});
 		}
+
+		this.wsPromise = Promise.resolve();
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -737,6 +737,9 @@ export class StreamChat {
 				type: 'connection.recovered',
 			});
 		}
+
+		this.wsPromise = Promise.resolve();
+		this.setUserPromise = Promise.resolve();
```
vishalnarkhede, line 742: We have `await this.setUserPromise` inside queryChannels and queryUsers. If the ws connection fails the first time but then gets recovered later, we need to unblock these await by resolving there promises.

PR #343, opened 2020-06-02
Drop unnecessary/simplify in query members

ferhatelmas:
* remove unnecessary to make test a bit faster

### DIFF
--- a/test/query_members.js
+++ b/test/query_members.js
@@ -135,12 +135,12 @@ describe('Query Members', function() {
 		const { members } = await channel.queryMembers({});
 		expect(members.length).to.be.equal(8);
 		for (let i = 0; i < members.length; i++) {
-			const { members: single } = channel.queryMembers(
+			const { members: single } = await channel.queryMembers(
 				{},
 				{},
 				{ limit: 1, offset: i },
 			);
-			expect(members.length).to.be.equal(1);
+			expect(single.length).to.be.equal(1);
 			expect(members[i].user.id).to.be.equal(single[0].user.id);
 		}
 	});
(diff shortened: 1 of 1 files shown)


test/query_members.js
```diff
@@ -47,19 +46,13 @@ describe('Query Members', function() {
...
 			created_by_id: mod,
 		});
 		await channel.create();
 		await channel.addModerators([mod]);
-		await channel.addMembers([rob]);
-		await channel.addMembers([rob2]);
-		await channel.addMembers([adam]);
-		await channel.addMembers([banned]);
-
+		await channel.addMembers([rob, rob2, adam, banned]);
```
thesyncim, line 55: this is on purpose.
our created_at has microsecond precision, and created_at is assigned in a loop.
we need this to ensure consistent test results

ferhatelmas: yup, reverted it to be explicit.

test/query_members.js
```diff
@@ -59,124 +56,148 @@ describe('Query Members', function() {
...
 		const clientA = await getTestClientForUser(invited);
 		await clientA.channel('messaging', channel.id).acceptInvite();
 
-		//reject the invite
+		// reject the invite
 		const clientR = await getTestClientForUser(rejected);
 		await clientR.channel('messaging', channel.id).rejectInvite();
 	});
 
+	it('query members with multiple filters client side', async function() {
```
ferhatelmas, line 73: @thesyncim this query should pass according to my understanding but fails. How we build query with `and` and `or` might be wrong. Anything wrong on syntax ? Otherwise, I will see SQL side.

ferhatelmas: It was in backend, fixed.

PR #346, opened 2020-06-02
add channels.name autocomplete test

imtoori:
add test for https://github.com/GetStream/chat/pull/1077

### DIFF
--- a/test/channels.js
+++ b/test/channels.js
@@ -1581,14 +1581,12 @@ describe('query by $autocomplete operator on channels.name', function() {
 	});
 
 	it('return 1 result', async function() {
-		const resp = await client
-			.queryChannels({
-				name: {
-				members: [user],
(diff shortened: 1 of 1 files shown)


test/channels.js
```diff
@@ -1566,6 +1566,35 @@ describe('$ne operator', function() {
...
+		});
+		await channel.create();
+	});
+
+	it('return 1 result', async function() {
+		console.log(channel.data.name);
+		console.log(channel.data.name.substring(0, 3));
+		const resp = await client
+			.queryChannels({
+				name: {
```
thesyncim, line 1588:
```suggestion
				name: {
				members: [user],
```

thesyncim: better to be a bit more selective

ferhatelmas: better, indeed @imtoori

ferhatelmas: I pushed this change, merging.

PR #347, opened 2020-06-08
Feature/translation

tbarbugli:


types/stream-chat/index.d.ts
```diff
@@ -462,6 +462,7 @@ export class Channel {
   hide(userId?: string, clearHistory?: boolean): Promise<APIResponse>;
   show(userId?: string): Promise<APIResponse>;
   getMessagesById(messageIds: string[]): Promise<GetMultipleMessagesAPIResponse>;
+  translateMessage(messageId: string, language: string): Promise<GetMessageAPIResponse>;
```
ferhatelmas, line 465: How about included `i18n` into `GetMessageAPIResponse` explicitly ?

PR #350, opened 2020-06-09
Handling rate-limited api calls

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -16,7 +16,7 @@ import fetch, { Headers } from 'cross-fetch';
 import FormData from 'form-data';
 import pkg from '../package.json';
 import { TokenManager } from './token_manager';
-import { isFunction, chatCodes } from './utils';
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -469,6 +468,12 @@ export class StreamChat {
 
 	doAxiosRequest = async (type, url, data, params = {}) => {
 		await this.tokenManager.tokenReady();
+
+		if (this.rateLimitResetTime) {
+			await sleep(this.rateLimitResetTime * 1000 - new Date());
```
ferhatelmas, line 473: might be better to add small noise so that all doesn't come at the same time

vishalnarkhede: noise?

ferhatelmas: jitter/delay - some small random time

vishalnarkhede: Makes sense!! Will do :)

PR #351, opened 2020-06-10
Add test for channel.truncated webhook

keyneston:
Currently `channel.truncated` doesn't fire a webhook. I'm adding it and this is the related test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "prettier-fix": "npx prettier --write '**/*.{js,ts,md,css,scss,json}' .eslintrc.json .prettierrc .babelrc",
     "test-api-dts": "node types/stream-chat/api-response-tests/index.js && npx prettier --write 'types/stream-chat/api-response-tests/data.ts' .eslintrc.json .prettierrc .babelrc && tsc types/stream-chat/api-response-tests/data.ts && rm types/stream-chat/api-response-tests/data.js && dtslint types/stream-chat",
     "eslint": "npx eslint '**/*.{js,md}' --max-warnings 0",
-    "test": "NODE_ENV=test mocha --exit --bail --timeout 15000 --require @babel/register test/*.js -- --async-stack-traces",
+    "test": "NODE_ENV=test mocha --exit --bail --timeout 15000 --require @babel/register test/*.js --async-stack-traces",
     "testall": "NODE_ENV=test mocha --exit --timeout 3000 --require @babel/register test/*.js -- --async-stack-traces",
     "testwatch": "NODE_ENV=test nodemon ./node_modules/.bin/mocha --timeout 15000 --require test-entry.js test/test.js",
     "lint": "npx prettier --list-different '**/*.{js,ts,css,scss,json}' .eslintrc.json .prettierrc .babelrc && npx eslint 'src/*.js' --max-warnings 0",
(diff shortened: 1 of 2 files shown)


test/webhook.js
```diff
@@ -375,10 +375,14 @@ describe('Webhooks', function() {
 	});
 
 	it('member.added', async function() {
-		await Promise.all([
+		const [events] = await Promise.all([
```
ferhatelmas, line 378: even `[event, ]`

package.json
```diff
@@ -113,7 +113,7 @@
     "prettier-fix": "npx prettier --write '**/*.{js,ts,md,css,scss,json}' .eslintrc.json .prettierrc .babelrc",
     "test-api-dts": "node types/stream-chat/api-response-tests/index.js && npx prettier --write 'types/stream-chat/api-response-tests/data.ts' .eslintrc.json .prettierrc .babelrc && tsc types/stream-chat/api-response-tests/data.ts && rm types/stream-chat/api-response-tests/data.js && dtslint types/stream-chat",
     "eslint": "npx eslint '**/*.{js,md}' --max-warnings 0",
-    "test": "NODE_ENV=test mocha --exit --bail --timeout 15000 --require @babel/register test/*.js -- --async-stack-traces",
+    "test": "NODE_ENV=test mocha --exit --bail --timeout 15000 --require @babel/register test/*.js --async-stack-traces",
```
ferhatelmas, line 116: this change is wrong because we can specify multiple patterns for files

ferhatelmas: let's revert it

ferhatelmas: or we can reorder parameters to reduce proneness to error

keyneston: The way the `--` is written before this change prevents all additional flags from being passed. Additionally it causes the `--async-stack-traces` flag to be ignored as shown by the following error:

> Warning: Could not find any test files matching pattern: --async-stack-traces

Or if you try and run `yarn test --grep 'foo'`:

> Warning: Could not find any test files matching pattern: --async-stack-traces
> Warning: Could not find any test files matching pattern: --grep
> Warning: Could not find any test files matching pattern: foo

ferhatelmas: `--async-stack-traces -- test/*.js`

how about this one ? not tested

ferhatelmas: it seems we also need to update `testall` with whatever we're doing

keyneston: `--` says "stop parsing any additional flags from this point. If it is included you can't pass any additional flags such as `--grep`. I'm confused why having multiple patterns for files conflicts with this? That should only matter if file names include `-` where yarn would think they were flags.

ferhatelmas: in short, your change is good. I think no need to overcomplicate with my version. 

Please be ensure to update `testall` too.

keyneston: Take this as an example:
```
 "test-tabs": "NODE_ENV=test mocha --exit --bail --timeout 15000 --require @babel/register --async-stack-traces test/utils.js test/webhook.js",
```
Tests two files, no `--` involved.

ferhatelmas: it's mainly npm trick to pass flags to running command if it's run `test: "some_cmd && npm run .. -- flags"`

keyneston: Ahh that makes sense.

PR #365, opened 2020-07-01
Feature/axios uploads

mahboubii:
- cross-fetch is removed
- SDK uses a local instance of Axios instead of the global one, prevent issues like sending cookies with intercepters

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -70,7 +70,7 @@ export class StreamChat {
 			delete this.options.httpsAgent;
 		}
 
-		this.reqest = axios.create(this.options);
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -70,6 +70,8 @@ export class StreamChat {
 			delete this.options.httpsAgent;
 		}
 
+		this.request = axios.create(this.options);
```
vishalnarkhede, line 73: Can we rename this to something like `this.axiosInstance`, to not confuse it with javascript's request object?

PR #376, opened 2020-07-14
Skip some events for the stopWatching test

keyneston:
So this test works by creating an event handler. Then on each event we append that we got the event. After we get three events we then do the checking to make sure everything is what we expect.

The specific order we get events in changes as is expected for a concurrent program. Additionally we sometimes get extra events such as `notification.added_to_channel`. These events may have keys such as `channel`, but not have keys like `user`. If these events show up early enough, then we crash. By skipping these events we can increase the pass rate of tests.

This was debugged and fixed with the assistance of @gumuz and @vishalnarkhede.

### DIFF
--- a/test/presence.js
+++ b/test/presence.js
@@ -97,6 +97,13 @@ describe('Presence', function() {
 			const results = [];
 			const eventPromise = new Promise(resolve => {
 				b.on('all', e => {
+					// We sometimes get other events which don't have a watcher_count or
+					// user. In which case this crashes. Ignore these events, and just
+					// filter down to relevant ones.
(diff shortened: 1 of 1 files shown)


ferhatelmas:
Better :+1:

thesyncim:
thanks for fixing this :)

keyneston:
@thesyncim no problem, there are few things that bug me more than inconsistent test failures.

keyneston:
I just had a build fail due to this test. I'm having a fear now that this isn't actually fixed. Has anyone else seen it fail recently?

PR #389, opened 2020-07-25
Allow no data in channel update

rubenvereecken:
# Submit a pull request

## CLA

... or something along these lines. It is currently impossible to just send a system message without an update to the channel's data. If I set an empty object, all my old custom data is overwritten. I thought I managed to work around this by setting `channelData=undefined` but it turns out that just magically ignores the attached message. The only way to do it, it seems, it by not setting the `data` field altogether. Just like the `addMembers` method does.

Please consider this functionality. It's a bit crippling.

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -256,10 +256,17 @@ export class Channel {
 	 * @return {type} The server response
 	 */
 	async update(channelData, updateMessage) {
-		const data = await this.getClient().post(this._channelURL(), {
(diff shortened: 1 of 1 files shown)


tbarbugli:
@rubenvereecken that's a neat trick to get a system message without updates :) I rather not merge this, we plan to add direct support for system messages like this in the future

rubenvereecken:
Glad to hear it

PR #392, opened 2020-07-29
Setter for userAgent

vishalnarkhede:
# Submit a pull request

## CLA

Introducing setter function for userAgent, so that we can use it to set custom userAgent from Chat component of stream-chat-react and stream-chat-react-native. This will help us track the stats of which sdks and version are consumers using.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1318,10 +1318,24 @@ export class StreamChat {
 		return await this.get(this.baseURL + `/messages/${messageID}`);
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -1318,10 +1318,24 @@ export class StreamChat {
 		return await this.get(this.baseURL + `/messages/${messageID}`);
 	}
 
+	/**
+	 * @deprecated Please use getUserAgent instead
```
ferhatelmas, line 1322: isn't it internal anyway?

PR #397, opened 2020-08-10
Test types

mahboubii:
fix CI error for testing types
remove old dtslint stuff and replace it with some sanity testing for types


ts-test/unit-test.ts
```diff
@@ -0,0 +1,260 @@
...
+
+const updateRequest = {
+  id: 'vishal',
+  set: {
+    name: 'Awesome',
+  },
+  unset: ['example'],
+};
+
+// TODO: update request type seems to be wrong! @nhannah
```
mahboubii, line 91: @nhannah looks like set and unset is not supported

ts-test/unit-test.ts
```diff
@@ -0,0 +1,260 @@
...
+const updateUser: Promise<{
+  users: { [key: string]: UserResponse<UserType> };
+}> = client.partialUpdateUser(updateRequest);
+const updateUsers: Promise<{
+  users: { [key: string]: UserResponse<UserType> };
+}> = client.partialUpdateUsers([updateRequest]);
+
+const eventHandler = (event: Event) => {};
+// @ts-expect-error
```
mahboubii, line 100: following expect errors appear to be wrong too @nhannah

PR #399, opened 2020-08-12
warm up connection

thesyncim:
# Submit a pull request

## CLA


src/client.js
```diff
@@ -56,6 +56,7 @@ export class StreamChat {
 		const defaultOptions = {
 			timeout: 3000,
 			withCredentials: false, // making sure cookies are not sent
+			warmUp: true,
```
tbarbugli, line 59: lets roll with this disabled first

src/client.js
```diff
@@ -788,8 +792,17 @@ export class StreamChat {
...
+		let warmUp;
+		if (this.options.warmUp) {
+			warmUp = this.doAxiosRequest('options', this.baseURL + '/connect');
+		}
+
 		const handshake = await this.wsConnection.connect();
 		this.connectionID = this.wsConnection.connectionID;
+		if (this.options.warmUp) {
+			// wait for warm up connection to complete
+			await warmUp;
```
tbarbugli, line 804: I think a fire and forget is better for this

thesyncim: done

src/client.js
thesyncim: done

src/client.js
```diff
@@ -56,6 +56,7 @@ export class StreamChat {
 		const defaultOptions = {
 			timeout: 3000,
 			withCredentials: false, // making sure cookies are not sent
+			warmUp: false,
```
vishalnarkhede, line 59: can we add a code comment about what does this mean exactly?

test/test.js
```diff
@@ -3003,3 +3003,48 @@ describe('paginate order with id_gt{,e}', () => {
...
+
+		// first client uses warmUp
+		const warmUpClient = getTestClientWithWarmUp();
+		warmUpClient.setBaseURL(baseUrl);
+		warmUpClient.health = await warmUpClient.setUser(
+			{ id: user },
+			createUserToken(user),
+		);
+
+		let t0 = new Date().getTime();
```
ferhatelmas, line 3030: using performance would be better for more accurate numbers since uses counters for relative time than real time system

https://nodejs.org/api/perf_hooks.html

src/client.js
```diff
@@ -491,6 +492,9 @@ export class StreamChat {
 				case 'patch':
 					response = await this.axiosInstance.patch(url, data, requestConfig);
 					break;
+				case 'options':
+					response = await this.axiosInstance.options(url);
```
ruudniew, line 496:
```suggestion
					response = await this.axiosInstance.options(url, requestConfig);
```

The `api_key` request header should be included, so the preflight request doesn't fail

ruudniew (changes requested):

PR #442, opened 2020-09-25
Fix non-deterministic autocomplete queryusers test

gumuz:
# Submit a pull request

## CLA

### DIFF
--- a/test/query_users.js
+++ b/test/query_users.js
@@ -73,6 +73,16 @@ describe('Query Users', function() {
 				{ username: { $autocomplete: 'ro' } },
 			],
 		});
+
+		// the users are not always returned in the same order, sort 'em
+		response.users.sort((a, b) => {
+			if (a.name === b.name) return 0;
+			if (a.name > b.name) {
+				return -1;
+			} else {
+				return 1;
+			}
+		});
 		expect(response.users[0].name).to.equal('Roxy');
 		expect(response.users[1].name).to.equal('Roxanne');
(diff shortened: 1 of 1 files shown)


test/query_users.js
```diff
@@ -73,6 +73,16 @@ describe('Query Users', function() {
 				{ username: { $autocomplete: 'ro' } },
 			],
 		});
+
+		// the users are not always returned in the same order, sort 'em
+		response.users.sort((a, b) => {
```
ferhatelmas, line 78: This is fine for tests to pass but not the actual solution.

We should fix it adapting sort parameters so that server returns a deterministic order or fix the issue in server.

test/query_users.js
```diff
@@ -66,13 +66,17 @@ describe('Query Users', function() {
...
-		});
+		const response = await serverClient.queryUsers(
+			{
+				unique,
+				$or: [
+					{ name: { $autocomplete: 'ro' } },
+					{ username: { $autocomplete: 'ro' } },
+				],
+			},
+			{ name: 1 },
```
ferhatelmas, line 77: That's good but default order, creation time, should also have worked in the first place.

Any explanation why it doesn't, precision lose in js ?

gumuz: Even though the updateUsers() accepts a list of users, the api call sends a map. I think this makes the creation order non-deterministic (edited)

ferhatelmas: We could do separate calls and put delay between calls to ensure creation time.

However, it would make tests slower so adding another sort order is good.

PR #444, opened 2020-09-29
Add permission for using frozen channels

ruudniew:
Adds the `UseFrozenChannel` permission to default permissions, adds tests


test/channel_permissions.js
```diff
@@ -55,6 +65,13 @@ function setupUser(ctx, user, done, test) {
 		});
 }
 
+async function getUserClient(ctx, user) {
```
ferhatelmas, line 68: you have `getTestClientForUser` in test utils

test/channel_permissions.js
```diff
@@ -18,13 +28,13 @@ class Context {
...
 
 	async setup() {
-		await this.authClient.updateUser(this.adminUser);
-		await this.authClient.updateUser(this.moderator);
-		await this.authClient.updateUser(this.guestUser);
-		await this.authClient.updateUser(this.channelMember);
-		await this.authClient.updateUser(this.channelOwner);
-		await this.authClient.updateUser(this.messageOwner);
-		await this.authClient.updateUser(this.scapegoatUser);
+		await this.authClient.upsertUser(this.adminUser);
```
ferhatelmas, line 31: That's good but maybe do it separately since there might be more cases of it. Otherwise, reason isn't clear.

test/channel_permissions.js
```diff
@@ -1104,6 +1298,42 @@ describe('Commerce permissions', function() {
 		);
 	});
 
+	describe('Create message on frozen channel', function() {
```
ferhatelmas, line 1301: What is ongoing here ? duplicated ?

test/channel_permissions.js
```diff
@@ -861,6 +1019,42 @@ describe('Gaming', function() {
...
+		it("admins can't create / delete reply to messages in a frozen channel", async () => {
+			await createDeleteReactionFrozen(ctx, ctx.adminUser, true);
+		});
+
+		describe('after admins get the UseFrozenChannelPermission', () => {
+			before(async () => {
+				await addUseFrozenChannelPermissions(ctx);
+			});
+
+			it('admins can create a message', async () => {
```
ferhatelmas, line 1048: how about delete ?

test/channel_permissions.js
ferhatelmas: It is repeated for all default channel types, hidden due to diff. 

Would be nice to dry it but not the target of this PR, of course.

test/channel_permissions.js
```diff
@@ -112,6 +129,75 @@ function createMessage(ctx, user, message, responseTest) {
...
+			assert.fail('send reaction or delete reaction failed: ', e);
+		}
+	}
+}
+
+async function removeUseFrozenChannelPermissions(ctx) {
+	const { permissions } = await ctx.authClient.getChannelType(ctx.channelType);
+	permissions.forEach((v, k) => {
+		if (v.name === BuiltinPermissions.UseFrozenChannel) {
+			permissions.splice(k, 1);
```
ferhatelmas, line 179: that's a bad practice to delete while iterating. I would use filter

test/channel_permissions.js
ruudniew: Deleting is possible on frozen channels without the new permission

ferhatelmas (approved):
LGTM

test/channel_permissions.js
ferhatelmas: might note it somewhere

PR #447, opened 2020-09-30
[CHAT-1206] client/channel shadowBan() & removeShadowBan() methods

gumuz:
client implementation + tests for shadow banning, see also: https://github.com/GetStream/chat/pull/1397


gumuz:
Would appreciate your review @vishalnarkhede to make sure I got all the typescript correct. The build / test-types check fails, because the backend isn't deployed yet, right

vishalnarkhede (approved):

PR #448, opened 2020-09-30
fix: #239 sendFile accepts buffer and File

mahboubii:


jaapbakker88 (approved):
LGTM

test/test.js
```diff
@@ -2901,6 +2901,12 @@ describe('Chat', () => {
 				expect(data.file).to.be.not.empty;
 			});
 
+			it('Upload a buffer', async () => {
```
ferhatelmas, line 2904: why didn't you add the other test in stream-js?

https://github.com/GetStream/stream-js/pull/389/files#diff-b3be5f00e4539d897e017e0426fb7606R22

ferhatelmas (approved):
LGTM

test/test.js
mahboubii: pushed a similar test 👍

PR #449, opened 2020-09-30
fix #179: regenerate yarn.lock file

### LINKED ISSUE #179: Can't resolve isomorphic-ws
Apologies if this is not the right place, but I'm trying to understand why I'm getting this error when running tests (via karma/puppeteer/chromium)

```
ERROR in ./node_modules/stream-chat/dist/browser.es.js
--
153 | Module not found: Error: Can't resolve 'isomorphic-ws' in 'node_modules/stream-chat/dist'
```

The code simply import `stream-chat`.
```
import { StreamChat } from 'stream-chat';
```

This works fine when running in the browser on local dev, i.e. webpack-dev-server.

Adding `isomorphic-ws` explicitly to my `package.json` seems to fix the issue. But I fail to understand why I need to add it, since it's already included because it's specified in `stream-chat-js` dependencies ([package.json](https://github.com/GetStream/stream-chat-js/blob/923347588ea07ce1a195a77ebefdde64bbb4769f/package.json#L42) and [yarn.lock](https://github.com/GetStream/stream-chat-js/blob/47d5a40630fe96f243e8b7fc786f858002a9bba3/yarn.lock#L3271))

Is it because it's defined as an [externalPackage](https://github.com/GetStream/stream-chat-js/blob/dc91e086e8a0a9fcab9ac4763989d697668378e8/rollup.config.js#L22) in rollup? and since the browser version is basically not using the dist/compiled, it builds everything and that resolves isomorphic-ws just fine? 

That's the only reasonable explanation I have, that something is wrong in the built version of stream-chat

mahboubii:


ferhatelmas (approved):
LGTM

nothing seriously to dig here

PR #452, opened 2020-10-01
Updating countUnread function

### LINKED ISSUE #336: Get full channel's unread messages count
[Docs](https://getstream.io/chat/docs/unread_channel/?language=js#unread-messages-per-channel) suggest to use channel.countUnread() to retrieve the number of unread messages in particular channel.

But in reality this method only count unread messages from fetched ones: there may be more. 
https://github.com/GetStream/stream-chat-js/blob/fe91fdf4d89a993f9cd32da3fd6d236682431567/src/channel.js#L652-L673

Is there any way to get real channel's unread count?
Thanks.

vishalnarkhede:
fixes #336


src/channel.ts
```diff
@@ -1097,16 +1097,17 @@ export class Channel<
...
 
   /**
-   * countUnread - Count the number of messages with a date thats newer than the last read timestamp
+   * countUnread - Count of unread messages
    *
    * @param {Date | Immutable.ImmutableDate | null} [lastRead] lastRead the time that the user read a message, defaults to current user's read state
    *
    * @return {number} Unread count
    */
   countUnread(lastRead?: Date | Immutable.ImmutableDate | null) {
```
vishalnarkhede, line 1106: `countUnread` function fails to return actual count of unread messages if [number of unread messages] > [number of messages loaded/queried in channel]. E.g., if user queries only first page of channel messages and user has 100 unread messages --->  `channel.countUnread()` will return only 25.

src/channel.ts
```diff
@@ -1462,6 +1464,10 @@ export class Channel<
             event.user.id,
             Immutable({ user: { ...event.user }, last_read: event.received_at }),
           );
+
+          if (event.user?.id === this.getClient().user?.id) {
+            s.unreadCount = 0;
```
vishalnarkhede, line 1469: set to 0 `message.read` event.

DanC5 (approved):

src/channel.ts
```diff
@@ -1476,6 +1482,11 @@ export class Channel<
         }
         break;
       case 'message.new':
+        s.unreadCount = s.unreadCount + 1;
```
mahboubii, line 1485:
```suggestion
        s.unreadCount = event.unread_count;
```
the same thing might need to be applied to `message.deleted`.

mahboubii: my guess is we return this value in some other channel related events too, we can check for the existence of this value at the beginning of the switch-case and keep it updated with all the events

mahboubii:
fixes #336

src/channel.ts
vishalnarkhede: `unread_count ` on event object is the total unread count (across all channels)!!

vishalnarkhede: @mahboubii ^^

vishalnarkhede, line 1485:
```suggestion
        if (event.user.id === this.getClient().user.id) {
          s.unreadCount = 0;
        } else {
          s.unreadCount = s.unreadCount + 1;
        }
```

mahboubii (approved):

PR #454, opened 2020-10-02
Feature: hard delete message event

mahboubii:


src/channel.ts
```diff
@@ -1496,9 +1497,14 @@ export class Channel<
           s.watchers = s.watchers.without(event.user.id);
         }
         break;
+      case 'message.deleted':
+        if (event.message) {
+          if (event.hard_delete) s.removeMessage(event.message as { id: string });
```
DanC5, line 1502: casting because id in MessageBase is possibly undefined?

src/channel.ts
```diff
@@ -1496,9 +1497,14 @@ export class Channel<
...
+          else s.addMessageSorted(event.message);
```
DanC5, line 1503: i wonder if we should update the `removeMessage` type parameter to match `addMessageSorted` so we don't have to cast, unless there's a reason it should stay as it is

src/channel.ts
mahboubii: yea, it's either checking the id for undefined(which somehow TS was failing to understand) or casting the message type

DanC5: i guess you could do `if (event.hard_delete && event.message.id)` if we don't want to change the type

src/channel.ts
mahboubii: actually, `removeMessage` param type is more correct than `addMessageSorted`, we should make the  `id` required in `MessageBase` type

DanC5: it definitely was required at one point, not sure why it turned optional, @nhannah do you remember?

mahboubii: swapped it to required

src/channel.ts
mahboubii: weirdly yea

src/channel.ts
DanC5: great! :vir_stache:, let's pretend that worked...

DanC5 (approved):

ferhatelmas (approved):
LGTM

PR #461, opened 2020-10-07
Handling rate limited calls

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -146,6 +146,9 @@ export class StreamChat<
   wsBaseURL?: string;
   wsConnection: StableWSConnection<ChannelType, CommandType, UserType> | null;
   wsPromise: ConnectAPIResponse<ChannelType, CommandType, UserType> | null;
+  rateLimitResetTime: {
+    [key: string]: number;
+  };
 
   /**
    * Initialize a client
@@ -281,6 +284,7 @@ export class StreamChat<
      * }
      */
     this.logger = isFunction(inputOptions.logger) ? inputOptions.logger : () => null;
+    this.rateLimitResetTime = {};
   }
 
   devToken(userID: string) {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -731,6 +735,17 @@ export class StreamChat<
     } = {},
   ): Promise<T> => {
     await this.tokenManager.tokenReady();
+
+    const resetTime = this.rateLimitResetTime[url]
```
ferhatelmas, line 739: how is old reset time cleaned ?

vishalnarkhede: Line 748

ferhatelmas: throw prevents it

vishalnarkhede: Yeah I mean ... we need it in case of error. If user calls the api again, it should again throw error. We should only cleanup if resetTime has expired

src/client.ts
```diff
@@ -765,13 +780,18 @@ export class StreamChat<
...
+        const code = e.response.data && e.response.data.code;
+        const headers = e.response.headers;
+
+        if (code === chatCodes.TOKEN_EXPIRED && !this.tokenManager.isStatic()) {
           this.tokenManager.loadToken();
           return await this.doAxiosRequest<T>(type, url, data, options);
         }
+
+        if (code === chatCodes.RATE_LIMITED && headers && headers['x-ratelimit-reset']) {
+          this.rateLimitResetTime[url] = e.response.headers['x-ratelimit-reset'];
```
mahboubii, line 792: is `url` a reliable key to cache the rate limits here? mostly because we have rate limits per endpoints and this is an enriched url.
I think either we go with a rate limit mapping to functions which could be lot of work or just cache based on function names, also a bit of work

vishalnarkhede: Yeah I realized that as well. Thinking of better way to approach it somehow:

1. Mapping based on handler names
2. Give access to some kind of callback function to user when rate limit is hit

vishalnarkhede:
Closing this for now. I needs more thought!!

PR #462, opened 2020-10-08
Flag options

### LINKED ISSUE #314: Missing server side support for Flag / Unflag
This endpoint requires the user that makes the action but JS client doesn't permit to set a user for server side auth.

/cc @vishalnarkhede

mahboubii:
closes #314

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -686,16 +686,6 @@ export type ChannelQueryOptions<
   watchers?: PaginationOptions;
 };
 
-export type FlagMessageOptions<UserType = UnknownType> = {
-  client_id?: string;
-  connection_id?: string;
-  created_by?: string;
-  target_message_id?: string;
-  target_user_id?: string;
-  user?: UserResponse<UserType>;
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -686,16 +686,6 @@ export type ChannelQueryOptions<
...
 
-export type FlagMessageOptions<UserType = UnknownType> = {
-  client_id?: string;
-  connection_id?: string;
-  created_by?: string;
-  target_message_id?: string;
-  target_user_id?: string;
-  user?: UserResponse<UserType>;
-  user_id?: string;
-};
```
mahboubii, line 697: looks like this specific type overcomplicates the functions for users, the only thing that is really needed here is `user_id?: string`

DanC5 (approved):
LGTM, and at some point we should probably add documentation about the different params used server-side. We currently have a tiny Server Side API docs section but it misses most of these use cases.

vishalnarkhede (approved):

PR #467, opened 2020-10-13
Adding support for event notification.invite_rejected

### LINKED ISSUE #416: Detecting when a user rejects an invite to a channel
When a user gets invited to a stream channel - 

```
const data = {
      members: [createdBy, ...invites],
      invites,
      created_by: { id: createdBy },
};
await this.client.channel(CustomChannelTypes.call, channelId, data).create();
```

I am watching for events - 

```
client.on((event) => {
   console.log(event)
})
```

an event is fired - `notification.invited` and similarly when the invitation has been accepted - `notification.invite_accepted`, however there is no specific event for when rejecting an invite, when I do - 

```
channel.rejectInvite()
```

instead, two events are fired - `channel.updated` and `member.updated`. Both of these event contain data about the fact that a user has an `invite_rejected_at` timestamp, however I believe so will any subsequent `channel.updated` and `member.updated` events, that were not caused by rejecting an invite. Therefore simply checking if an update event contains details of a user having an `invite_rejected_at` timestamp wouldn't be enough, as it may of happened in a previous `channel.updated` / `member.updated` event that was not when the user rejected.

Without a `notification.invite_rejected`, how can I detect when an invite has been rejected straight away?

gz#5775

vishalnarkhede:
# Submit a pull request

Depends on https://github.com/GetStream/chat/pull/1434/files

Closes #416

## CLA

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -20,6 +20,7 @@ export const EVENT_MAP = {
   'notification.channel_mutes_updated': true,
   'notification.channel_truncated': true,
   'notification.invite_accepted': true,
+  'notification.invite_rejected': true,
   'notification.invited': true,
   'notification.mark_read': true,
   'notification.message_new': true,
--- a/src/types.ts
+++ b/src/types.ts
@@ -886,6 +886,7 @@ export type EventTypes =
   | 'notification.channel_mutes_updated'
   | 'notification.channel_truncated'
   | 'notification.invite_accepted'
+  | 'notification.invite_rejected'
   | 'notification.invited'
   | 'notification.mark_read'
   | 'notification.message_new'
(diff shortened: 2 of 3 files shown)


test/invites.js
```diff
@@ -180,6 +187,7 @@ describe('Member style channel init', () => {
 		const response = await nickChannel.rejectInvite();
 		expect(response.members[1].user.id).to.equal('nick');
 		expect(response.members[1].invite_rejected_at).to.not.equal(null);
+		await notificationReceived;
```
ferhatelmas, line 190: `inviteRejectedReceived` would be better as a name

AnatolyRugalev:
Hi @vishalnarkhede 

The API support is almost there. I'm going to rebase this PR and test it with the recent changes

AnatolyRugalev:
Everything looks OK, now we need to wait for full API rollout

PR #470, opened 2020-10-14
perf: faster addMessagesSorted

### LINKED ISSUE #465: Improve performance of channelState.addMessageSorted
image

gz#6578

mahboubii:
fixes #465

image


src/channel_state.ts
```diff
@@ -495,35 +453,50 @@ export class ChannelState<
...
+    if (messages[messages.length - 1].created_at.getTime() < message.created_at.getTime())
+      return messages.concat(message);
+
+    // find the closest index to push the new message
+    let left = 0;
+    let middle = 0;
+    let right = messages.length - 1;
+    while (left <= right) {
+      middle = Math.floor((right + left) / 2);
+      if (messages[middle].created_at.getTime() <= message.created_at.getTime())
```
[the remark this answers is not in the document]
mahboubii: good call

src/channel_state.ts
```diff
@@ -495,35 +453,50 @@ export class ChannelState<
...
+    // if message is newer than last item in the list just concat and return
+    if (messages[messages.length - 1].created_at.getTime() < message.created_at.getTime())
+      return messages.concat(message);
+
+    // find the closest index to push the new message
+    let left = 0;
+    let middle = 0;
+    let right = messages.length - 1;
+    while (left <= right) {
+      middle = Math.floor((right + left) / 2);
```
nhannah, line 470: In most cases wouldn't a bottom up approach be faster than a binary search given the likelihood of a message being further up is less?

mahboubii: L462 takes care of that. cases like pagination could add some 100 more messages to the other end of the array which will be O(n), right now this part is O(log(n)) which is quite fast already

nhannah (approved):
Looks good lets test

vishalnarkhede (approved):

src/channel.ts
```diff
@@ -1658,7 +1658,7 @@ export class Channel<
     if (!this.state.messages) {
       this.state.messages = Immutable([]);
     }
-    this.state.addMessagesSorted(messages, true);
+    this.state.addMessagesSorted(messages, false, true);
```
ferhatelmas, line 1661: Having options an object with keys would help for readability.

ferhatelmas (approved):
:100: LGTM

src/channel.ts
mahboubii: 3 params are fine, in case it grows we can change it to object

ferhatelmas: I agree in general but when they are booleans, it's trickier.

PR #473, opened 2020-10-21
remove unnecessary _initializeState call in watch()

mahboubii:
`this._initializeState(state)` is already called in `channel.query` 
https://github.com/GetStream/stream-chat-js/blob/bug/duplicate-channel-initializeState/src/channel.ts#L1246


vishalnarkhede (approved):

ferhatelmas (approved):
LGTM

PR #475, opened 2020-10-23
[CHAT-1270] ignore shadowed messages in unread count

gumuz:
make Channel.countUnread() ignore shadowed messages

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1171,6 +1171,9 @@ export class Channel<
       if (this.getClient().userID === message.user?.id) {
         continue;
       }
+      if (m.shadowed) {
+        continue;
+      }
       if (m.silent) {
         continue;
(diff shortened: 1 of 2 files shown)


src/channel.ts
```diff
@@ -1141,6 +1141,9 @@ export class Channel<
       if (this.getClient().userID === message.user?.id) {
         continue;
       }
+      if (m.shadowed) {
```
mahboubii, line 1144: the same thing needs to be added to `countUnreadMentions` fn

mahboubii (changes requested):
shadowed should probably be discarded in this count too
https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L1551

mahboubii (approved):

PR #485, opened 2020-10-30
[CHAT-1172] Ban by IP feature

AnatolyRugalev:
This PR contains Ban by IP feature implementation.

I had to add some `X-Forwarded-For` trickery for IP manipulation, I would love to hear the feedback about this approach.

Also fixed a couple of warnings in `utils.js`

**Proposed Ban By IP API (Server-Side only):**

To ban user account and it's last known IP address you should provide `ip_ban: true` option to the `banUser` call:
```js
client.banUser(userToBan.id, {
  user_id: adminUser.id,
  ip_ban: true,
});
```

The unban procedure is unchanged:
```js
client.unbanUser(userToUnban.id);
```

When unbanning user who was banned by IP, IP will be unbanned even if there's another account who was using the same IP address.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -137,6 +137,7 @@ export class StreamChat<
   secret?: string;
   setUserPromise: ConnectAPIResponse<ChannelType, CommandType, UserType> | null;
   state: ClientState<UserType>;
+  testIPOverride?: string;
   tokenManager: TokenManager<UserType>;
   user?: UserResponse<UserType>;
   userAgent?: string;
@@ -1077,6 +1078,7 @@ export class StreamChat<
       wsBaseURL: client.wsBaseURL,
       clientID: client.clientID,
       userID: client.userID,
+      testIPOverride: client.testIPOverride,
       tokenManager: client.tokenManager,
       user: this._user,
       authType: this.getAuthType(),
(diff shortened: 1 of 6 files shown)


AnatolyRugalev:
Could use some help understanding why tests are failing

src/client.ts
```diff
@@ -137,6 +137,7 @@ export class StreamChat<
   secret?: string;
   setUserPromise: ConnectAPIResponse<ChannelType, CommandType, UserType> | null;
   state: ClientState<UserType>;
+  testIPOverride?: string;
```
AnatolyRugalev, line 140: Any name suggestions?

mahboubii: @AnatolyRugalev is this field only for testing purposes? 
if yes I'd suggest removing it from client code completely and just stub the `connection._buildUrl` function in the test env

AnatolyRugalev: Yes, this is only for testing. Good idea, will try that

AnatolyRugalev: I got it working with basic suffixing of `userAgent` string. I tried to make mock work but I cant override `_buildUrl` of the connection because the `wsConnection` property is null while client is not connected. And when client is already connected changing URL is useless.

mahboubii (approved):
🌮

PR #496, opened 2020-11-09
[CHAT-852] Add support for banned_by_id in ban methods

AnatolyRugalev:
This change deprecates user_id field in ban options to avoid misinterpretation of the field purpose. REST API will support both names, but the clients should migrate towards new name: `moderator_id`.

**This PR should be merged only after the backend release**

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1604,6 +1604,13 @@ export class StreamChat<
    * @returns {Promise<APIResponse>}
    */
   async banUser(targetUserID: string, options?: BanUserOptions<UserType>) {
+    if (options?.user_id !== undefined) {
+      options.banned_by_id = options.user_id;
+      delete options.user_id;
+      console.warn(
+        "banUser: 'user_id' is deprecated, please consider switching to 'banned_by_id'",
+      );
+    }
     return await this.post<APIResponse>(this.baseURL + '/moderation/ban', {
       target_user_id: targetUserID,
       ...options,
--- a/test/integration/moderation.js
+++ b/test/integration/moderation.js
@@ -389,6 +389,8 @@ describe('A user with a channel shadow ban', function () {
 
 describe('show ban status on member', function () {
 	const guyonID = `guyon-${uuidv4()}`;
+	const user1ID = `user1-${uuidv4()}`;
+	const user2ID = `user2-${uuidv4()}`;
 	const evilID = `evil-${uuidv4()}`;
 
 	const channelID = `channel-${uuidv4()}`;
@@ -397,7 +399,7 @@ describe('show ban status on member', function () {
 	const client = getTestClient(true);
 
 	before(async function () {
-		await createUsers([guyonID, evilID]);
+		await createUsers([guyonID, evilID, user1ID, user2ID]);
 		channel = client.channel('livestream', channelID, {
 			members: [guyonID, evilID],
 			created_by_id: guyonID,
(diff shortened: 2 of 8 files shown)


mahboubii:
@AnatolyRugalev I think this makes sense, the only downside is it would be inconsistent with the rest of the library. @ferhatelmas any idea?

AnatolyRugalev:
@mahboubii REST API will support both fields until we'll feel safe to remove user_id support

src/channel.ts
```diff
@@ -1263,6 +1263,9 @@ export class Channel<
    */
   async banUser(targetUserID: string, options: BanUserOptions<UserType>) {
     this._checkInitialized();
+    if (options?.user_id !== undefined) {
```
gumuz, line 1266: I don't think you need it here, since it gets dispatched to the client banUser()

test/integration/moderation.js
```diff
@@ -389,6 +389,8 @@ describe('A user with a channel shadow ban', function () {
 
 describe('show ban status on member', function () {
 	const guyonID = `guyon-${uuidv4()}`;
+	const user1ID = `user1-${uuidv4()}`;
```
gumuz, line 392: what's the purpose of these additional userids here, they don't seem to be used anywhere

gumuz (changes requested):
maybe it makes sense to leave some tests that still uses user_id, to ensure backwards compatibility

test/integration/moderation.js
AnatolyRugalev: This IDs are used in the tests for backwards compatibility [here](https://github.com/GetStream/stream-chat-js/pull/496/files#diff-d94fa608cd42452bb60618acc150397dd1e93ad02319fb4850c09cdcd32c2967R410)

gumuz: sorry, I missed that!

AnatolyRugalev:
> maybe it makes sense to leave some tests that still uses user_id, to ensure backwards compatibility

@gumuz this is exactly what is done [here](https://github.com/GetStream/stream-chat-js/pull/496/files#diff-d94fa608cd42452bb60618acc150397dd1e93ad02319fb4850c09cdcd32c2967R410)

gumuz:
> > maybe it makes sense to leave some tests that still uses user_id, to ensure backwards compatibility
> 
> @gumuz this is exactly what is done [here](https://github.com/GetStream/stream-chat-js/pull/496/files#diff-d94fa608cd42452bb60618acc150397dd1e93ad02319fb4850c09cdcd32c2967R410)

I don't know how I missed that 😆

gumuz (approved):

src/client.ts
```diff
@@ -1604,6 +1604,9 @@ export class StreamChat<
    * @returns {Promise<APIResponse>}
    */
   async banUser(targetUserID: string, options?: BanUserOptions<UserType>) {
+    if (options?.user_id !== undefined) {
+      options.moderator_id = options.user_id;
```
mahboubii, line 1608: should we delete the `user_id` from the options if we replace it with `moderator_id`?

mahboubii (approved):

src/client.ts
AnatolyRugalev: I marked field as deprecated to avoid backwards incompatible changes. I'm not very experienced in TypeScript, so I'm not sure if field removal will break SDK users' code. If it's not, then probably would be sane to do that

mahboubii: I mean deleting the `user_id` from the request  body before sending it to API, so you wouldn't have a request with `{user_id, moderator_id}`, however, this will make the backward compatibility test partially obsolete

AnatolyRugalev: Ah, yes, we probably should. Would be the best to narrow things as much as possible to be sure that everything works with new approach.

AnatolyRugalev:
This fails now because backend was not updated yet

AnatolyRugalev:
Update: renamed field moderator_id to banned_by_id to provide more clarity. Also added warning when developer uses old field

PR #497, opened 2020-11-09
Fix missing image_moderation_enabled field

AnatolyRugalev:
This PR fixes adds image_moderation_enabled field to AppSettings related types

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -81,6 +81,7 @@ export type AppSettingsAPIResponse<
     custom_action_handler_url?: string;
     disable_auth_checks?: boolean;
     disable_permissions_checks?: boolean;
+    image_moderation_enabled?: boolean;
     multi_tenant_enabled?: boolean;
     name?: string;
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -1173,6 +1174,7 @@ export type AppSettings = {
     notification_template?: string;
     server_key?: string;
   };
+  image_moderation_enabled?: boolean;
```
gumuz, line 1177: I'm not so sure about this, because it seems to be missing most app setting....

AnatolyRugalev: I followed the test code:

```js
                await client.updateAppSettings({
			image_moderation_enabled: true,
		});
```
And the method description is:

```js
async updateAppSettings(options: AppSettings)
```

I'm not sure how this tests are performed, but it seemed logical to put this option there too

AnatolyRugalev: Will try to run tests without this field

PR #498, opened 2020-11-10
Fix/unread count mutes

mahboubii:


src/channel.ts
```diff
@@ -1165,27 +1164,17 @@ export class Channel<
...
-      if (this.getClient().userID === message.user?.id) {
-        continue;
-      }
-      if (m.shadowed) {
-        continue;
-      }
-      if (m.silent) {
-        continue;
-      }
-      if (lastRead == null) {
```
mahboubii, line 1180: look like this check never passes but it is logically incorrect, removed.

src/client.ts
```diff
@@ -1684,6 +1692,21 @@ export class StreamChat<
     });
   }
 
+  /** userMuteStatus - check if a user is muted or not, can be used after setUser() is called
+   *
+   * @param {string} targetID
+   * @returns {boolean}
+   */
+  userMuteStatus(targetID: string) {
+    if (!this.user || !this.wsPromise)
```
mahboubii, line 1701: @vishalnarkhede does this make sense? this function is valid only if you set a user so client.mutes is initialized when the `health.check` is fired.
I think we need the same check for `client._muteStatus()` which handles channel mutes

src/client.ts
```diff
@@ -932,6 +936,10 @@ export class StreamChat<
     if (event.type === 'notification.channel_mutes_updated' && event.me?.channel_mutes) {
       this.mutedChannels = event.me.channel_mutes;
     }
+
+    if (event.type === 'notification.mutes_updated' && event.me?.mutes) {
+      this.mutedUsers = event.me.mutes;
```
mahboubii, line 941: @ferhatelmas does API always return the whole mutes list or could this be a partial list if the list is long?

nhannah (approved):
Overall looks good to me, not sure if you want to add that check on _muteStatus you mentioned

vishalnarkhede (approved):

src/channel.ts
```diff
@@ -1135,27 +1149,12 @@ export class Channel<
...
-        count++;
-        continue;
-      }
-      if (m.created_at > lastRead) {
+    for (let i = 0; i < this.state.messages.length; i += 1) {
```
ferhatelmas, line 1155: caching length might help

ferhatelmas: `i++` looks more idiomatic

src/channel.ts
```diff
@@ -1135,27 +1149,12 @@ export class Channel<
...
+      const message = this.state.messages[i];
+      if (this._countMessageAsUnread(message) && message.created_at > lastRead) {
```
ferhatelmas, line 1157: reversing conditions might be faster

src/client.ts
ferhatelmas: yes, it cuts by 100

src/channel.ts
```diff
@@ -1127,6 +1127,20 @@ export class Channel<
     }
   }
 
+  _countMessageAsUnread(message: {
```
ferhatelmas, line 1130: how is channel mute integrated into this calculation ?

src/channel.ts
mahboubii: the performance difference is negligible for caching the length
https://eslint.org/docs/rules/no-plusplus

src/channel.ts
mahboubii, line 1157:
```suggestion
      if (message.created_at > lastRead && this._countMessageAsUnread(message)) {
```

src/channel.ts
ferhatelmas: makes sense, we can assume this is resolved.

ferhatelmas (approved):
LGTM

PR #500, opened 2020-11-10
[CHAT-1176] Add tests to prove sort's deterministic nature

AnatolyRugalev:
This PR adds a test for sort functions which proves that field order in multi-sort situations is preserved

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1055,6 +1055,16 @@ export class StreamChat<
     }
   }
 
+  _buildSort(
+    sort: UserSort | ChannelSort,
+  ): Array<{ field: string; direction?: AscDesc }> {
+    const sortFields: Array<{ field: string; direction?: AscDesc }> = [];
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -1055,6 +1055,16 @@ export class StreamChat<
     }
   }
 
+  _buildSort(
+    sort: UserSort | ChannelSort,
+  ): Array<{ field: string; direction?: AscDesc }> {
+    const sortFields: Array<{ field: string; direction?: AscDesc }> = [];
+    for (const [field, direction] of Object.entries(sort)) {
```
ferhatelmas, line 1062: There is no order guarantee here in es5 and in es6, insertion order to object (numeric/symbol are irrelevant to us) which should be documented probably.

AnatolyRugalev: OK, I didn't know that. I will try to come up with a real solution then

PR #501, opened 2020-11-11
[CHAT-1314] Added ability to provide array of sort directions

AnatolyRugalev:
This PR adds ability to provide sort configuration using array of objects to guarantee order of the fields.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -355,7 +355,7 @@ export class Channel<
    * queryMembers - Query Members
    *
    * @param {UserFilters<UserType>}  filterConditions object MongoDB style filters
-   * @param {UserSort<UserType>} [sort] Sort options, for instance {created_at: -1}.
+   * @param {UserSort<UserType>} [sort] Sort options, for instance [{created_at: -1}].
    * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_active: -1}, {created_at: 1}]
    * @param {{ limit?: number; offset?: number }} [options] Option object, {limit: 10, offset:10}
    *
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -363,7 +363,7 @@ export class Channel<
    */
   async queryMembers(
     filterConditions: UserFilters<UserType>,
-    sort: UserSort<UserType> | UserSort<UserType>[] = {},
+    sort: UserSort<UserType> | UserSort<UserType>[] = [],
     options: { limit?: number; offset?: number } = {},
   ) {
     let id: string | undefined;
@@ -382,7 +382,7 @@ export class Channel<
           type,
           id,
           members,
-          sort: this.getClient()._buildSort(sort),
+          sort: this.getClient()._normalizeQuerySort(sort),
           filter_conditions: filterConditions,
           ...options,
         },
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1,7 +1,7 @@
 import Immutable from 'seamless-immutable';
 import { ChannelState } from './channel_state';
 import { isValidEventType } from './events';
-import { logChatPromiseExecution } from './utils';
+import { logChatPromiseExecution, normalizeQuerySort } from './utils';
 import { StreamChat } from './client';
 import {
   APIResponse,
@@ -355,7 +355,7 @@ export class Channel<
    * queryMembers - Query Members
    *
    * @param {UserFilters<UserType>}  filterConditions object MongoDB style filters
-   * @param {UserSort<UserType> | UserSort<UserType>[]} [sort] Sort options, for instance {created_at: -1}.
+   * @param {UserSort<UserType>} [sort] Sort options, for instance {created_at: -1}.
    * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_active: -1}, {created_at: 1}]
    * @param {{ limit?: number; offset?: number }} [options] Option object, {limit: 10, offset:10}
    *
@@ -363,7 +363,7 @@ export class Channel<
    */
   async queryMembers(
     filterConditions: UserFilters<UserType>,
-    sort: UserSort<UserType> | UserSort<UserType>[] = [],
+    sort: UserSort<UserType> = [],
     options: { limit?: number; offset?: number } = {},
   ) {
     let id: string | undefined;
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -1055,6 +1055,26 @@ export class StreamChat<
...
+    sort: UserSort | ChannelSort | ChannelSort[] | UserSort[],
+  ): Array<{ field: string; direction?: AscDesc }> {
+    const sortFields: Array<{ field: string; direction?: AscDesc }> = [];
+    const sortArr = Array.isArray(sort) ? sort : [sort];
+    for (const item of sortArr) {
+      const entries = Object.entries(item);
+      if (entries.length > 1) {
+        console.warn(
+          "client._buildSort() - multiple fields in a single sort object detected. Object's field order is not guaranteed",
+        );
```
AnatolyRugalev, line 1069: Not sure about this one

src/client.ts
```diff
@@ -1055,6 +1055,26 @@ export class StreamChat<
     }
   }
 
+  _buildSort(
+    // todo: generalize type
+    sort: UserSort | ChannelSort | ChannelSort[] | UserSort[],
+  ): Array<{ field: string; direction?: AscDesc }> {
```
AnatolyRugalev, line 1061: Couldn't come up with generalized type that works for `sort`. Would appreciate the ideas

AnatolyRugalev:
A little clarification for the reviewers to speed up the process.

The original issue comes from the fact that JavaScript's language specification doesn't enforce Object's field order. While EcmaScript 2015 introduces some ruleset to make the behaivor more predictable, still, the field order may not follow the order of the insertion on different platforms and browsers.

While the Node's implementation does return string fields in the desired order, we can't be sure that this behavior will be preserved.

I've tested following code in modern versions of Firefox, Chrome and Node and I see that this behavior is consistent:

```js
> Object.entries({"created_at": 1, "updated_at": -1, "zoo": 2, "xxx":"xxx", "6":"6"})
< 0: Array [ "6", "6" ]
< 1: Array [ "created_at", 1 ]
< 2: Array [ "updated_at", -1 ]
< 3: Array [ "zoo", 2 ]
< 4: Array [ "xxx", "xxx" ]
< 5: Array [ "6nine", 69 ]
> Object.entries(JSON.parse('{"created_at": 1, "updated_at": -1, "zoo": 2, "xxx":"xxx", "6":"6", "6nine": 69}'))
< 0: Array [ "6", "6" ]
< 1: Array [ "created_at", 1 ]
< 2: Array [ "updated_at", -1 ]
< 3: Array [ "zoo", 2 ]
< 4: Array [ "xxx", "xxx" ]
< 5: Array [ "6nine", 69 ]
```

As you can see, numeric fields are moved to the "beginning" of the object, but the string fields appearance order is preserved. Frankly, we don't have and probably won't have numeric fields names ever, so the change (and especially warning) may not worth it.

I suppose the merge decision should be made by @ferhatelmas as the issue reporter.

One more thing worth mentioning: the most of our SDKs use this "array with single-field objects" approach to provide sort direction, so may be this PR would bring some consistency between them

src/client.ts
```diff
@@ -1055,6 +1055,26 @@ export class StreamChat<
     }
   }
 
+  _buildSort(
```
mahboubii, line 1058: maybe rename this to `_normalizeQuerySort` or sth more related?

test/unit/sort.js
```diff
@@ -0,0 +1,55 @@
+import chai from 'chai';
+import { StreamChat } from '../../src';
+
+const expect = chai.expect;
+
+describe('test if sort is deterministic', () => {
```
mahboubii, line 6: can we move this test to the client test file?

src/channel.ts
```diff
@@ -352,23 +352,20 @@ export class Channel<
...
+   * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_active: -1}, {created_at: 1}]
    * @param {{ limit?: number; offset?: number }} [options] Option object, {limit: 10, offset:10}
    *
-   * @return {Promise<ChannelMemberAPIResponse<UserType>>} search members response
+   * @return {Promise<ChannelMemberAPIResponse<UserType>>} Query Members response
    */
   async queryMembers(
     filterConditions: UserFilters<UserType>,
-    sort: UserSort<UserType> = {},
+    sort: UserSort<UserType> | UserSort<UserType>[] = {},
```
mahboubii, line 366: we should probably make the `UserSort<UserType>[]` a mutually exclusive type so that each array entry object has only one attribute inside. @nhannah is it possible?

src/client.ts
```diff
@@ -1109,21 +1129,17 @@ export class StreamChat<
...
+   * @param {UserSort<UserType> | UserSort<UserType>[]} sort Sort options, for instance {last_active: -1}.
+   * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_active: -1}, {created_at: 1}]
    * @param {UserOptions} options Option object, {presence: true}
    *
    * @return {Promise<APIResponse & { users: Array<UserResponse<UserType>> }>} User Query Response
    */
   async queryUsers(
     filterConditions: UserFilters<UserType>,
-    sort: UserSort<UserType> = {},
+    sort: UserSort<UserType> | UserSort<UserType>[] = {},
```
mahboubii, line 1140: default `[]` if we are moving towards deprecating the `UserSort<UserType>`?

src/channel.ts
AnatolyRugalev: I tried this with union type like this:

```typescript
export type ChannelSort = {created_at: AscDesc} | {updated_at: AscDesc}
```

but it seemed ugly to me. Also, type parameters of `Sort` made it so bad so I couldn't come up with proper `buildSort` input type which will work for everyone. As I said, would appreciate any help in this area

test/unit/sort.js
AnatolyRugalev: You mean, rename it to `client.js`? I don't see any client-related js files right now. There's `channel_state.js` and a bunch of integration tests

mahboubii: you're right, `client.js` test file is not yet merged to master. let's leave it as it is and I will move the tests there later.

AnatolyRugalev:
Thanks a lot to @mahboubii and @nhannah for helping with this PR! :1st_place_medal: 

I've rebased the branch and it's ready for the final review now

mahboubii (approved):

PR #502, opened 2020-11-13
[CHAT-852] Add support for banned_by_id in ban methods

AnatolyRugalev:
Current versions of SDKs have very unclear way of executing a ban from server-side:

```
client.banUser('user1', {
  user_id: 'user2'
})
```

Our customers struggle to understand the meaning of user_id field. This often leads to banning the wrong user account.

The new behavior will look like this:

```
client.banUser('user1', {
  banned_by_id: 'user2'
})
```

This change is fully backwards compatible and will throw a warning if developer is still using "user_id" field.

This PR should only be merged after API update deployment

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -590,10 +590,18 @@ export type UserResponse<UserType = UnknownType> = User<UserType> & {
  */
 
 export type BanUserOptions<UserType = UnknownType> = UnBanUserOptions & {
+  banned_by?: UserResponse<UserType>;
+  banned_by_id?: string;
   ip_ban?: boolean;
   reason?: string;
   timeout?: number;
+  /**
+   * @deprecated please use banned_by
+   */
   user?: UserResponse<UserType>;
+  /**
+   * @deprecated please use banned_by_id
+   */
   user_id?: string;
 };
 
--- a/test/integration/moderation.js
+++ b/test/integration/moderation.js
@@ -413,6 +413,8 @@ describe('show ban status on member', function () {
 		await channel.create();
 	});
 
+	// this tests backwards-compatibility of the server
+	// so you can safely drop it after user and user_id fields removal
 	it('check user_id->banned_by_id backwards compatibility', async function () {
 		await channel.banUser(evilID2, { banned_by_id: guyonID });
 		await channel.banUser(evilID3, { user_id: guyonID });
--- a/test/integration/moderation.js
+++ b/test/integration/moderation.js
@@ -390,14 +390,22 @@ describe('A user with a channel shadow ban', function () {
 describe('show ban status on member', function () {
 	const guyonID = `guyon-${uuidv4()}`;
 	const evilID = `evil-${uuidv4()}`;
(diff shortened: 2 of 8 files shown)


src/types.ts
```diff
@@ -590,10 +590,14 @@ export type UserResponse<UserType = UnknownType> = User<UserType> & {
  */
 
 export type BanUserOptions<UserType = UnknownType> = UnBanUserOptions & {
+  banned_by_id?: string;
```
gumuz, line 593: didn't the backend changes also include `BannedBy`?

test/integration/serverside.js
```diff
@@ -604,7 +604,7 @@ describe('Managing users', function () {
 
 	it('ban user', async function () {
 		await client.banUser(evilUser, {
-			user_id: user.id,
+			banned_by_id: user.id,
```
mahboubii, line 607: it might be a good idea to add an extra test to make sure the old field is still supported

```js
await client.post(client.baseURL + '/moderation/ban', {
      target_user_id: evilUser,
      user_id: user.id
 });

```

AnatolyRugalev: It is done already, please see https://github.com/GetStream/stream-chat-js/pull/502/files#diff-d94fa608cd42452bb60618acc150397dd1e93ad02319fb4850c09cdcd32c2967R410

mahboubii: these two sends a similar request to the API, code snippet above ensure API continues supporting deprecated field

src/types.ts
AnatolyRugalev: Yes. I'm not exactly sure how it could be used from the SDK as we don't had it previously. I was following Tommaso's suggestion to support everything we had previously, including `User` field. So old `User` became `BannedBy`

gumuz: there's a `user` param here on line 597

test/integration/serverside.js
AnatolyRugalev: Ah, yes, got it.

src/types.ts
AnatolyRugalev: I'm going to add some tests cases for this one, will check if it even works

test/integration/moderation.js
```diff
@@ -390,23 +390,46 @@ describe('A user with a channel shadow ban', function () {
...
 		});
 		await channel.create();
 	});
 
+	it('check user_id->banned_by_id backwards compatibility', async function () {
+		await channel.banUser(evilID2, { banned_by_id: guyonID });
+		await channel.banUser(evilID3, { user_id: guyonID });
+		await channel.banUser(evilID, { banned_by: guyon });
+		await channel.banUser(evilID, { user: guyon });
+		await client.post(client.baseURL + '/moderation/ban', {
```
ferhatelmas, line 421: good to add a note to drop this part while removing deprecated part

ferhatelmas (approved):
LGTM

mahboubii (approved):
🌮

PR #504, opened 2020-11-16
Add channel export API

tbarbugli:


src/client.ts
```diff
@@ -2150,4 +2153,31 @@ export class StreamChat<
   deleteBlockList(name: string) {
     return this.delete<APIResponse>(`${this.baseURL}/blocklists/${name}`);
   }
+
+  exportChannels(request: Array<ExportChannelRequest>) {
```
mahboubii, line 2157: this is a server only feature? should we throw error when it's called using client tokens?

src/client.ts
```diff
@@ -2150,4 +2153,31 @@ export class StreamChat<
...
+  }
+
+  getExportChannelStatus(id: string) {
+    return this.get<APIResponse & ExportChannelStatusResponse>(
+      `${this.baseURL}/export_channels/${id}`,
+    );
+  }
+}
+
+function makeExportPayload(v: ExportChannelRequest) {
```
mahboubii, line 2176: i'd merge this into `client.exportChannels` since it's simple and not used anywhere else

src/types.ts
```diff
@@ -1203,6 +1214,13 @@ export type BlockList = {
   words: string[];
 };
 
+export type ExportChannelRequest = {
+  id: string;
+  type: string;
+  messages_since?: Date;
+  messages_until?: Date;
```
mahboubii, line 1221: we might want to support string too for ease of use

src/types.ts
```diff
@@ -103,6 +103,17 @@ export type BlockListResponse = BlockList & {
   updated_at?: string;
 };
 
+export type ExportChannelResponse = {
+  task_id: string;
+};
+
+export type ExportChannelStatusResponse = {
```
mahboubii, line 110: missing ` status` field?

src/client.ts
tbarbugli: yes

mahboubii (approved):

PR #508, opened 2020-11-17
Update sort types

nhannah:
The tests need to be fixed as I didn't have time to look at them, I believe these should be the correct typings and I don't think the helper function needs to reside on the client. @mahboubii we need to discuss the allowed syntax though. Should it be that the sort takes a sort object or an array of sort object? or should it be 1:1 with the backend, where it takes a sort object or and array of {field, direction} objects?


AnatolyRugalev (approved):
LGTM. And the tests are OK. I'm going to merge into the main PR and address review comments there

PR #509, opened 2020-11-20
fix: duplicated new messages by current user

mahboubii:
There are some edgy cases with react components when message preview is added and gets updated after it's actually is sent. Few users are experiencing race condition when the actual event update the list before react is able to do so.


vishalnarkhede (approved):

DanC5 (approved):

ferhatelmas (approved):
LGTM

PR #516, opened 2020-11-21
custom channel event tests

tbarbugli:


mahboubii (approved):

gumuz (commented):
What happens when you don't include an event type or keep it empty?

What happens if you pretend to send builtin events?

PR #518, opened 2020-11-24
replace uuid with nanoid

mahboubii:
we don't really need cryptographically strong random values here, simpler implementation to support ReactNative/Expo apps.

clientID will look like this `firh1Xuozx1Q0UHUgUENp` instead of UUID format in API logs. 

[`nanoid/non-secure`](https://github.com/ai/nanoid/blob/master/non-secure/index.js) implemention : 
```js
// This alphabet uses `A-Za-z0-9_-` symbols. The genetic algorithm helped optimize the gzip compression for this alphabet.
urlAlphabet = 'ModuleSymbhasOwnPr-0123456789ABCDEFGHNRVfgctiUvz_KqYTJkLxpZXIjQW';

 nanoid = (size = 21) => {
  let id = ''
  // A compact alternative for `for (var i = 0; i < step; i++)`.
  let i = size
  while (i--) {
    // `| 0` is more compact and faster than `Math.floor()`.
    id += urlAlphabet[(Math.random() * 64) | 0]
  }
  return id
}

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "form-data": "^3.0.0",
     "isomorphic-ws": "^4.0.1",
     "jsonwebtoken": "^8.5.1",
-    "nanoid": "^3.1.18",
     "seamless-immutable": "^7.1.4",
     "ws": "^7.3.1"
   },
--- a/package.json
+++ b/package.json
@@ -38,15 +38,14 @@
     "@babel/runtime": "^7.12.1",
     "@types/jsonwebtoken": "^8.5.0",
     "@types/seamless-immutable": "7.1.13",
-    "@types/uuid": "^8.3.0",
     "@types/ws": "^7.2.7",
     "axios": "^0.21.0",
     "base64-js": "^1.3.1",
     "form-data": "^3.0.0",
     "isomorphic-ws": "^4.0.1",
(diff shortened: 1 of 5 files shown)


virdesai (approved):
hopefully this works even tho the nanoid site says the extra lib is still needed: https://github.com/ai/nanoid#react-native

mahboubii:
> hopefully this works even tho the nanoid site says the extra lib is still needed: https://github.com/ai/nanoid#react-native

only required for the secure version

package.json
```diff
@@ -109,7 +108,8 @@
     "rollup-plugin-replace": "^2.1.0",
     "rollup-plugin-terser": "^7.0.2",
     "sinon": "^9.2.0",
-    "typescript": "^4.0.3"
+    "typescript": "^4.0.3",
+    "uuid": "^8.3.1"
```
ferhatelmas, line 112: any plan to use it in dev too ?

ferhatelmas (approved):
LGTM

package.json
mahboubii: too many file changes which clash with other PRs, another time.

tbarbugli (approved):

vishalnarkhede:
Thanks @mahboubii \m/

PR #529, opened 2020-12-04
Add connectUser function

ruudniew:
* Add connectUser function as an alias for setUser
* Warn when we use connectUser or setUser server side. This warning can be skipped by passing `allowServerSideConnect: true` to the client when initializing
* This should not break QA


src/client.ts
```diff
@@ -315,22 +315,29 @@ export class StreamChat<
...
     if (this.userID) {
       throw new Error(
-        'Use client.disconnect() before trying to connect as a different user. setUser was called twice.',
+        'Use client.disconnect() before trying to connect as a different user. connectUser or setUser was called twice.',
+      );
+    }
+
+    if (this._isUsingServerAuth() && !this.options.serverConnectUser) {
+      throw new Error(
+        'Do not use connectUser or setUser server side. If you have a valid use-case, add "serverConnectUser: true" to the client options to ignore this warning.',
```
ferhatelmas, line 337: > ignore this warning

this is an error

src/client.ts
```diff
@@ -354,6 +361,20 @@ export class StreamChat<
     return this.setUserPromise;
   };
 
+  /**
+   * setUser - Set the current user, this triggers a connection to the API
```
ferhatelmas, line 365: sounds like this is deprecated so better says so

src/types.ts
```diff
@@ -795,6 +795,7 @@ export type SearchOptions = {
 export type StreamChatOptions = AxiosRequestConfig & {
   browser?: boolean;
   logger?: Logger;
+  serverConnectUser?: boolean;
```
ferhatelmas, line 798: Not sure but this name sounds confusing and could make understanding its purpose harder.

src/client.ts
ruudniew: It can lead to confusion, because it's not clear how urgent it is to change to `connectUser`. We had this in the past

src/types.ts
ruudniew: Any suggestions?

ferhatelmas: naming is hard but a verbose guess to make it harder to misinterpret, `enableBecomingOnlineFromServerSide`

src/client.ts
```diff
@@ -315,22 +315,29 @@ export class StreamChat<
...
     if (this.userID) {
       throw new Error(
-        'Use client.disconnect() before trying to connect as a different user. setUser was called twice.',
+        'Use client.disconnect() before trying to connect as a different user. connectUser or setUser was called twice.',
+      );
+    }
+
+    if (this._isUsingServerAuth() && !this.options.serverConnectUser) {
+      throw new Error(
+        'Do not use connectUser or setUser server side. If you have a valid use-case, add "serverConnectUser: true" to the client options to disable this error.',
```
gumuz, line 337: What's the usecase for calling `connectUser` with server auth?

AnatolyRugalev: We use it in QA, so I guess we can't fully restrict the things we use by ourselves :)

The only use case I can come up with is desktop/CLI client

gumuz: Don't we use `getTestClient(serverSide)` for this?

src/client.ts
AnatolyRugalev: I think we can clarify this in the deprecation notice. Otherwise it would not be clear which method to call because they do the same thing. The original issue is caused by the fact that developer did not think that `setUser` has this huge side-effect. We can just say in the notice that this method is renamed to `connectUser` to better reflect its purpose

src/client.ts
AnatolyRugalev: Yeah, you're right, that doesn't make any sense

src/client.ts
```diff
@@ -315,22 +315,29 @@ export class StreamChat<
...
    */
-  setUser = (
+  connectUser = (
     user: UserResponse<UserType>,
     userTokenOrProvider: TokenOrProvider,
   ): ConnectAPIResponse<ChannelType, CommandType, UserType> => {
     if (this.userID) {
       throw new Error(
-        'Use client.disconnect() before trying to connect as a different user. setUser was called twice.',
+        'Use client.disconnect() before trying to connect as a different user. connectUser or setUser was called twice.',
```
AnatolyRugalev, line 331: I think we can drop `setUser` from the message

src/types.ts
AnatolyRugalev: What if we drop the option and force developer to stop using server key? If there's a use case for this, you can always use two different clients :)

vishalnarkhede:
Whats the reason for switching `setUser` --> `connectUser`?

AnatolyRugalev:
@vishalnarkhede 

Recently we discovered that this method naming leads some developers to believe that it could/should be used on a server-side before executing an operation on this user. This creates unnecessary websocket connections which eventually cause rate-limiting of the real clients.

Current method name does not reflect the things that it does. Also we have `disconnect` which probably suggests that we have to have `connect` somewhere in the SDK as well

AnatolyRugalev:
Hey @ruudniew! I hope you're doing well :slightly_smiling_face: 

I've got some feedback on this change and this is what we need to do:
1. Replace error with a warning - we don't want to break existing implementations
2. Allow to ignore the warning with some kind of flag
3. Use `node` environment detection to see if we're running on backend instead of checking the secret

I think the flag part is the trickiest one: we have to choose clear naming. A couple of options from the top of my head:
1. `allowServerSideConnect`
2. `disableServerSideConnectWarning`

src/client.ts
```diff
@@ -2040,6 +2085,16 @@ export class StreamChat<
    */
   _isUsingServerAuth = () => !!this.secret;
 
+  _isRunningNode = () => {
```
AnatolyRugalev, line 2088: Probably missed preposition here

src/client.ts
```diff
@@ -2040,6 +2085,16 @@ export class StreamChat<
...
+    let isNode;
+    try {
+      isNode = Object.prototype.toString.call(global.process) === '[object process]';
```
AnatolyRugalev, line 2091: Looks too complicated IMO. Cant we just check if `global.process` is defined instead of catching an exception?

src/types.ts
```diff
@@ -793,6 +793,7 @@ export type SearchOptions = {
 };
 
 export type StreamChatOptions = AxiosRequestConfig & {
+  allowServerSideConnect?: boolean;
```
AnatolyRugalev, line 796: The naming is great now I think

test/integration/utils.js
```diff
@@ -7,7 +7,10 @@ const apiKey = process.env.STREAM_API_KEY;
 const apiSecret = process.env.STREAM_API_SECRET;
 
 export function getTestClient(serverSide) {
-	return new StreamChat(apiKey, serverSide ? apiSecret : null, { timeout: 15000 });
+	return new StreamChat(apiKey, serverSide ? apiSecret : null, {
+		timeout: 15000,
+		serverConnectUser: serverSide,
```
AnatolyRugalev, line 12: Old flag name here

test/typescript/utils.js
```diff
@@ -38,7 +38,10 @@ module.exports = {
 		return this.getTestClient(true);
 	},
 	getTestClient: function getTestClient(serverSide) {
-		return new StreamChat(apiKey, serverSide ? apiSecret : null, { timeout: 8000 });
+		return new StreamChat(apiKey, serverSide ? apiSecret : null, {
+			timeout: 8000,
+			serverConnectUser: serverSide,
```
AnatolyRugalev, line 43: Same

src/client.ts
ruudniew: This is supposed to be a very reliable way to do it. Source: https://github.com/iliakan/detect-node.

test/unit/client.js
```diff
@@ -62,3 +62,48 @@ describe('Client userMuteStatus', function () {
...
 	});
 });
+
+describe('Detect node environment', () => {
+	const client = new StreamChat('', '');
+	it('_isRunningNode should be true', () => {
+		expect(client._isRunningNode()).to.be.true;
+	});
+
+	it('should warn when using connectUser on a node environment', async () => {
```
AnatolyRugalev, line 72: Nice test :+1:

AnatolyRugalev (approved):
LGTM

src/client.ts
```diff
@@ -2044,6 +2089,16 @@ export class StreamChat<
    */
   _isUsingServerAuth = () => !!this.secret;
 
+  _isRunningInNode = () => {
```
mahboubii, line 2092: this is not something that changes in runtime, better to initialize it in the constructor

mahboubii: actually, you can reuse `this.node` value

src/types.ts
mahboubii, line 796: could you add a comment clarifying this attribute?

src/client.ts
```diff
@@ -332,22 +332,32 @@ export class StreamChat<
   _hasConnectionID = () => Boolean(this.connectionID);
 
   /**
-   * setUser - Set the current user, this triggers a connection to the API
+   * connectUser - Set the current user, this triggers a connection to the API
```
mahboubii, line 335:
```suggestion
   * connectUser - Set the current user and open a WebSocket connection
```

same thing for `connectAnonymousUser` to clarify a bit further?

vishalnarkhede (approved):
Overall looks good to me. The only concern is, we have used this function on all sdks. So if we are changing the name, better to do it everywhere :)

src/client.ts
```diff
@@ -1422,7 +1470,7 @@ export class StreamChat<
     custom: ChannelData<ChannelType> = {} as ChannelData<ChannelType>,
   ) {
     if (!this.userID && !this._isUsingServerAuth()) {
```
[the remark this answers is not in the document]
AnatolyRugalev: The server side is allowed to call this method because it should explicitly set the user. I think we should leave this one as is

AnatolyRugalev:
> we have used this function on all sdks. So if we are changing the name, better to do it everywhere :)

@vishalnarkhede other SDKs don't have the same issue because this is the only one which supports both client and server side calls. Server side SDKs do not support WS anyway, so they should not have this method at all. And the client side SDKs don't have this problem :)

AnatolyRugalev:
Let's carefully merge it whenever you are ready @ruudniew

test/unit/client.js
```diff
@@ -62,3 +62,48 @@ describe('Client userMuteStatus', function () {
...
+			warning = msg;
+		};
+
+		try {
+			await client.connectUser({ id: 'user' }, 'fake token');
+		} catch (e) {}
+
+		await client.disconnect();
+		expect(warning).to.equal(
+			'Please do not use connectUser server side. If you have a valid use-case, add "allowServerSideConnect: true" to the client options to disable this warning.',
```
lambardi, line 85: Perhaps we should mention the billing implications here?

'Please do not use connectUser server side. connectUser impacts MAU and concurrent connection usage and thus your bill. If you have a valid use-case, add "allowServerSideConnect: true" to the client options to disable this warning.'

PR #523, opened 2020-11-30
[CHAT-1146] Add Before Message Send webhook tests

AnatolyRugalev:
This PR adds before-message-send webhook tests which check for webhook_failed and webhook_id field.

Also I decided to refactor all webhook-related tests simplifying the setup code and grouping them together.

PR #533, opened 2020-12-07
[CHAT-1422] Rolled back webhook_id and webhook_failed message fields

AnatolyRugalev:
The original change was made in #523. The corresponding API is going to be rolled back


ferhatelmas (approved):
LGTM

PR #534, opened 2020-12-07
recoverStateOnReconnect option and type UpdatedMessage

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -80,6 +80,7 @@ import {
   UpdateChannelResponse,
   UpdateCommandOptions,
   UpdateCommandResponse,
+  UpdatedMessage,
   UpdateMessageAPIResponse,
   UserFilters,
   UserOptions,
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -1039,6 +1051,10 @@ export class StreamChat<
         tags: ['connection', 'client'],
       });
 
+      this.dispatchEvent({
```
vishalnarkhede, line 1054: Also, if the queryChannels on line 1044 fails, then we don't dispatch connection.recovered event, which our UI sdks rely upon. So we run into situations where network has been recovered, but we never fired connection.recovered event. By letting UI sdk handle state recovery themselves, components can keep track of api failures and display relevant error indicator.

mahboubii (approved):

nhannah (approved):
Looks good to me as long as the types and integration tests are both working.

PR #540, opened 2020-12-09
isUpToDate flag on channel state

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stream-chat",
-  "version": "2.9.1-dev.0",
+  "version": "2.9.0",
   "description": "JS SDK for the Stream Chat API",
   "author": "GetStream",
   "homepage": "https://getstream.io/chat/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stream-chat",
-  "version": "2.9.0",
+  "version": "2.10.0",
   "description": "JS SDK for the Stream Chat API",
   "author": "GetStream",
   "homepage": "https://getstream.io/chat/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stream-chat",
-  "version": "2.10.0",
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "stream-chat",
-  "version": "2.9.0",
+  "version": "2.9.1-dev.0",
```
mahboubii, line 3:
```suggestion
  "version": "2.9.0",
```

src/channel_state.ts
```diff
@@ -424,6 +437,10 @@ export class ChannelState<
     }
   }
 
+  setIsUptoDate = (isUpToDate: boolean) => {
```
mahboubii, line 440: does it make sense to move this to the channel level?

mahboubii, line 440: some jsdoc would also be good explaining this similar to this.isUpToDate

src/channel.ts
```diff
@@ -1547,13 +1547,19 @@ export class Channel<
         if (event.message) {
           /* if message belongs to current user, always assume timestamp is changed to filter it out and add again to avoid duplication */
           const ownMessage = event.user?.id === this.getClient().user?.id;
-          s.addMessageSorted(event.message, ownMessage);
+
+          if (this.state.isUpToDate || event.message.parent_id) {
```
mahboubii, line 1551: thread message with `show_in_channel` flag? maybe disable this completely for threaded messages

mahboubii (commented):
could you also add a few tests for it

src/channel_state.ts
vishalnarkhede: Its only about messages, so I thought state would be a better place!!

src/channel.ts
vishalnarkhede: Good catch

mahboubii (approved):
@vishalnarkhede looks good 👍, could you also add some tests for it?

PR #545, opened 2020-12-14
Removed unused field from `query_channels`

AnatolyRugalev:
Field `user_details` is not used in `query_channels` API in any way. In cases when `this._user` field gets too large, application could experience performance issues.


thesyncim (approved):
LGTM

vishalnarkhede (approved):

PR #546, opened 2020-12-15
[CHAT-1428] Switched queryChannels to POST

AnatolyRugalev:
We added support for POST in the `queryChannels` API to address performance issues


ferhatelmas (approved):
LGTM

PR #549, opened 2020-12-17
Fix presence test

ruudniew:
Sometimes a notification is delivered, which fails the test. This should fix it.

### DIFF
--- a/test/integration/presence.js
+++ b/test/integration/presence.js
@@ -135,17 +135,19 @@ describe('Presence', function () {
 			const results = [];
 			const eventPromise = new Promise((resolve) => {
 				b.on('all', (e) => {
-					results.push([e.watcher_count, e.user.id]);
-					expect(e.watcher_count).to.equal(b.state.watcher_count);
-					// expect to see thierry join, james join and james leave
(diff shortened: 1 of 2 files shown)


test/integration/presence.js
```diff
@@ -135,17 +135,19 @@ describe('Presence', function () {
...
-					// expect to see thierry join, james join and james leave
-					if (results.length === 3) {
-						const expected = [
-							[1, 'user1'],
-							[2, 'james'],
-							[1, 'james'],
-						];
-						expect(results).to.deep.equal(expected);
-						resolve();
+					if (e.type.split('.')[0] !== 'notification') {
```
gumuz, line 138: `!(e.type.startsWith('notification.'))` might be a bit more readable

gumuz: but wouldn't it be nicer to instead subscribe or check for the specific event we're waiting for?

AnatolyRugalev: I agree. Seems strange to filter out events this way. At least we could reverse the condition and make it return early to make it simpler to read and understand

gumuz (approved):
LGTM!

PR #550, opened 2020-12-17
[CHAT-1477] partial update channels 

thesyncim:
Partial Update
A partial update can be used to set and unset specific fields when it is necessary to retain additional custom data fields on the object. AKA a patch style update.

```javascript
// Here's a channel with some custom field data that might be useful
channel = client.channel(type, id , {
   source: "user",
   source_detail:{ user_id: 123 },
   channel_detail:{ topic: "Plants and Animals", rating: "pg" }
})

//  let's change the source of this channel
await channel.updatePartial({ set:{ source: "system" } });

//  since it's system generated we no lonfer need source_detail
await channel.updatePartial({ unset: ["source_detail"] });

// set nested field
await channel.updatePartial({ set: { "channel_detail.topic": "Nature" } });

// remove the topic property from channel_detail object 
await channel.updatePartial({ unset: ["channel_detail.topic"] });
```

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -535,11 +535,9 @@ export class Channel<
    * @return {Promise<PartialUpdateChannelAPIResponse<ChannelType,CommandType, UserType>>}
    */
   async updatePartial(update: PartialChannelUpdate<ChannelType>) {
-    return await this.getClient().patch<
+    return this.getClient().patch<
       PartialUpdateChannelAPIResponse<ChannelType, CommandType, UserType>
-    >(this._channelURL(), {
-      ...update,
-    });
+    >(this._channelURL(), update);
   }
 
   /**
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -27,6 +27,8 @@ import {
   MessageFilters,
   MuteChannelAPIResponse,
   PaginationOptions,
+  PartialChannelUpdate,
+  PartialUpdateChannelAPIResponse,
   Reaction,
   ReactionAPIResponse,
   SearchAPIResponse,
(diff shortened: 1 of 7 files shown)


vishalnarkhede:
Build will be good after this - https://github.com/GetStream/stream-chat-js/pull/551/files

src/channel.ts
```diff
@@ -525,6 +527,21 @@ export class Channel<
...
 
+  /**
+   * updatePartial - partial update channel properties
+   *
+   * @param {PartialChannelUpdate<ChannelType>} partial update request
+   *
+   * @return {Promise<PartialUpdateChannelAPIResponse<ChannelType,CommandType, UserType>>}
+   */
+  async updatePartial(update: PartialChannelUpdate<ChannelType>) {
+    return await this.getClient().patch<
```
mahboubii, line 538:
```suggestion
    return this.getClient().patch<
```

src/channel.ts
```diff
@@ -525,6 +527,21 @@ export class Channel<
...
+      PartialUpdateChannelAPIResponse<ChannelType, CommandType, UserType>
+    >(this._channelURL(), {
+      ...update,
+    });
```
mahboubii, line 542:
```suggestion
    >(this._channelURL(), update);
```

src/channel.ts
```diff
@@ -27,6 +27,8 @@ import {
   MessageFilters,
   MuteChannelAPIResponse,
   PaginationOptions,
+  PartialChannelUpdate,
```
mahboubii, line 30: maybe rename this to `PartialUpdateChannel`  to be more consistent

AnatolyRugalev (approved):
LGTM

src/channel.ts
ferhatelmas: @thesyncim why did you resolve ?

PR #569, opened 2020-12-30
add isOwnUser type guard and add OwnUser to Client User type

nhannah:
Adjusting types and adding type guard to reflect OwnUserResponse as a possible type on the client.user variable.


virdesai (approved):
LGTM

vishalnarkhede (approved):

PR #572, opened 2021-01-04
Removed .only from the tests

AnatolyRugalev:
All tests are being ignored right now :)


AnatolyRugalev:
Going to merge ASAP to avoid false positives in CI

PR #578, opened 2021-01-07
[CHAT-1516] Support pinned messages in channel state

AnatolyRugalev:
This PR adds support for `pinned_messages` property in channel state.

More about this feature in the docs: https://getstream.io/chat/docs/pinned_messages/?language=js

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1728,11 +1728,10 @@ export class Channel<
       this.state.messages = Immutable([]);
     }
     this.state.addMessagesSorted(messages, false, true);
-    const pinnedMessages = state.pinned_messages || [];
-    if (!this.state.pinned_messages) {
-      this.state.pinned_messages = Immutable([]);
+    if (!this.state.pinnedMessages) {
+      this.state.pinnedMessages = Immutable([]);
     }
-    this.state.setPinnedMessages(pinnedMessages);
+    this.state.setPinnedMessages(state.pinned_messages || []);
     this.state.watcher_count = state.watcher_count ? state.watcher_count : 0;
     // convert the arrays into objects for easier syncing...
     if (state.watchers) {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1556,6 +1556,9 @@ export class Channel<
         if (event.message) {
           if (event.hard_delete) s.removeMessage(event.message);
           else s.addMessageSorted(event.message);
+          if (event.message.pinned) {
+            s.removePinnedMessage(event.message);
+          }
         }
         break;
       case 'message.new':
@@ -1569,7 +1572,7 @@ export class Channel<
             s.addMessageSorted(event.message, ownMessage);
           }
           if (event.message.pinned) {
-            s.setPinnedMessage(event.message);
+            s.addPinnedMessage(event.message);
           }
 
           if (ownMessage && event.user?.id) {
@@ -1589,7 +1592,11 @@ export class Channel<
       case 'message.updated':
         if (event.message) {
           s.addMessageSorted(event.message);
-          s.setPinnedMessage(event.message);
+          if (event.message.pinned) {
+            s.addPinnedMessage(event.message);
+          } else {
+            s.removePinnedMessage(event.message);
+          }
         }
         break;
       case 'channel.truncated':
@@ -1731,7 +1738,7 @@ export class Channel<
     if (!this.state.pinnedMessages) {
       this.state.pinnedMessages = Immutable([]);
     }
-    this.state.setPinnedMessages(state.pinned_messages || []);
+    this.state.addPinnedMessages(state.pinned_messages || []);
     this.state.watcher_count = state.watcher_count ? state.watcher_count : 0;
     // convert the arrays into objects for easier syncing...
     if (state.watchers) {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1568,6 +1568,9 @@ export class Channel<
           if (this.state.isUpToDate || isThreadMessage) {
             s.addMessageSorted(event.message, ownMessage);
           }
+          if (event.message.pinned) {
+            s.setPinnedMessage(event.message);
+          }
 
           if (ownMessage && event.user?.id) {
             s.unreadCount = 0;
@@ -1586,6 +1589,7 @@ export class Channel<
       case 'message.updated':
         if (event.message) {
           s.addMessageSorted(event.message);
+          s.setPinnedMessage(event.message);
         }
         break;
       case 'channel.truncated':
@@ -1724,6 +1728,11 @@ export class Channel<
       this.state.messages = Immutable([]);
     }
     this.state.addMessagesSorted(messages, false, true);
+    const pinnedMessages = state.pinned_messages || [];
+    if (!this.state.pinned_messages) {
+      this.state.pinned_messages = Immutable([]);
+    }
+    this.state.setPinnedMessages(pinnedMessages);
     this.state.watcher_count = state.watcher_count ? state.watcher_count : 0;
     // convert the arrays into objects for easier syncing...
     if (state.watchers) {
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -1,4 +1,4 @@
-import Immutable from 'seamless-immutable';
+import Immutable, { ImmutableDate } from 'seamless-immutable';
 import { Channel } from './channel';
 import {
   ChannelMemberResponse,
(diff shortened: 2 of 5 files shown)


src/channel.ts
```diff
@@ -1724,6 +1728,11 @@ export class Channel<
       this.state.messages = Immutable([]);
     }
     this.state.addMessagesSorted(messages, false, true);
+    const pinnedMessages = state.pinned_messages || [];
+    if (!this.state.pinned_messages) {
+      this.state.pinned_messages = Immutable([]);
+    }
+    this.state.setPinnedMessages(pinnedMessages);
```
mahboubii, line 1735:
```suggestion
    if (!this.state.pinned_messages) {
      this.state.pinned_messages = Immutable([]);
    }
    this.state.setPinnedMessages(state.pinned_messages || []);
```

src/channel_state.ts
```diff
@@ -207,6 +221,7 @@ export class ChannelState<
       ...message,
       __html: message.html,
       // parse the date..
+      pinned_at: message.pinned_at ? new Date(message.pinned_at) : new Date(),
```
mahboubii, line 224: shouldn't this be null if the message is not pinned?

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
     }
   }
 
+  /**
+   * setPinnedMessages - updates messages in pinned_messages property. See setPinnedMessages for details
```
mahboubii, line 288:
```suggestion
   * setPinnedMessages - updates messages in pinned_messages property.
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+   *
+   * @param {Array<MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} newMessages A list of messages
```
mahboubii, line 290:
```suggestion
   * @param {Array<MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} pinnedMessages A list of messages
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+   *
+   */
+  setPinnedMessages(
+    newMessages: MessageResponse<
```
mahboubii, line 294:
```suggestion
    pinnedMessages: MessageResponse<
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+      AttachmentType,
+      ChannelType,
+      CommandType,
+      MessageType,
+      ReactionType,
+      UserType
+    >[],
+  ) {
+    for (let i = 0; i < newMessages.length; i += 1) {
+      this.setPinnedMessage(newMessages[i]);
```
mahboubii, line 304:
```suggestion
    for (let i = 0; i < pinnedMessages.length; i += 1) {
      this.setPinnedMessage(pinnedMessages[i]);
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+    }
+  }
+
+  /**
+   * setPinnedMessage - update message in pinned_messages property. The messages will be added to pinned_messages if it
+   * is pinned and removed otherwise
+   *
+   * @param {MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>} newMessage message to update
```
mahboubii, line 312:
```suggestion
   * @param {MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>} pinnedMessage message to update
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+   *
+   */
+  setPinnedMessage(
+    newMessage: MessageResponse<
```
mahboubii, line 316:
```suggestion
    pinnedMessage: MessageResponse<
```

src/channel_state.ts
```diff
@@ -269,6 +284,58 @@ export class ChannelState<
...
+      AttachmentType,
+      ChannelType,
+      CommandType,
+      MessageType,
+      ReactionType,
+      UserType
+    >,
+  ) {
+    const message = this.messageToImmutable(newMessage);
```
mahboubii, line 325:
```suggestion
    const message = this.messageToImmutable(pinnedMessage);
```
better to move this inside the if case

src/channel.ts
```diff
@@ -1586,6 +1589,7 @@ export class Channel<
       case 'message.updated':
         if (event.message) {
           s.addMessageSorted(event.message);
+          s.setPinnedMessage(event.message);
```
mahboubii, line 1592: `message.deleted` case is not handled

src/channel_state.ts
```diff
@@ -62,6 +62,19 @@ export class ChannelState<
       >['messageToImmutable']
     >
   >;
+  pinned_messages: Immutable.ImmutableArray<
```
mahboubii, line 65: this should be camelCased to stay consistent with the rest of the state

src/channel_state.ts
```diff
@@ -507,8 +574,34 @@ export class ChannelState<
...
+          AttachmentType,
+          ChannelType,
+          CommandType,
+          EventType,
+          MessageType,
+          ReactionType,
+          UserType
+        >['messageToImmutable']
+      >,
+    ) => ImmutableDate = (msg) => msg.created_at,
```
mahboubii, line 589: while this is a more powerful approach, it can be simplified to a simple string for the field name: `sortBy  = 'created_at'`

src/channel_state.ts
```diff
@@ -546,19 +639,6 @@ export class ChannelState<
...
-          ChannelType,
-          CommandType,
-          EventType,
-          MessageType,
-          ReactionType,
-          UserType
-        >['messageToImmutable']
-      >
-    >;
     mutable.splice(left, 0, message);
```
mahboubii, line 642: looks like an accidental delete

mahboubii (changes requested):

src/channel_state.ts
AnatolyRugalev: TS was complaining when I was trying to do that

src/channel_state.ts
AnatolyRugalev: I moved it to the top of the method because I need this type for `sortBy`

src/channel_state.ts
AnatolyRugalev: Any help with that is appreciated of course

src/channel_state.ts
```diff
@@ -507,8 +574,34 @@ export class ChannelState<
...
+          AttachmentType,
+          ChannelType,
+          CommandType,
+          EventType,
+          MessageType,
+          ReactionType,
+          UserType
+        >['messageToImmutable']
+      >
+    >;
```
mahboubii, line 604:
```suggestion
```
this is at the end of the function to avoid running it for early return cases

src/channel_state.ts
mahboubii: will look into the error when you updated it, worst case we just silence the TS compiler

AnatolyRugalev: It works fine now. I guess I had other issue which I fixed in a meanwhile

src/channel_state.ts
AnatolyRugalev: It will become `Date | null` which is impossible to reuse in `_addToMessageList`

src/channel_state.ts
gumuz: @mahboubii should probably confirm this, but this might work:
```typescript
// changed typedef to string
sortBy = 'created_at'
(...)
const attr = message[sortBy]
if(attr instanceof Date) {
  const messageTime = attr.getTime();
}
```

src/channel_state.ts
AnatolyRugalev: Nevermind, it is broken. Fixing...

src/channel_state.ts
```diff
@@ -519,10 +600,10 @@ export class ChannelState<
     // for empty list just concat and return
     if (messageArr.length === 0) return messageArr.concat(message);
 
-    const messageTime = message.created_at.getTime();
+    const messageTime = message[sortBy]?.getTime() || 0;
```
mahboubii, line 603: did you try casting the type? `(message[sortBy] as ImmutableDate).getTime()` or `(message[sortBy] as Date).getTime()`

AnatolyRugalev: As I said, the types works fine now. It was an issue in `pinned_at` definition when I was trying to do this originally

mahboubii: great, it's generally good practice to avoid adding unnecessary js code to satisfy TS compiler, type casting is cleaner and faster here, especially since this function is a performance bottleneck

AnatolyRugalev: Is this better? https://github.com/GetStream/stream-chat-js/pull/578/commits/2d2bea43f7cb2959b4141fc752446d1e2d3be2a4

mahboubii: Looks great!

mahboubii:
@AnatolyRugalev is this feature still pending prod deploy? 
Type errors: `Property 'pinned_messages' is missing in type ...`

mahboubii (approved):
🌮

AnatolyRugalev:
@mahboubii yes, performing some final checks.

Thank you for the feedback!

PR #579, opened 2021-01-08
Cleaned up deprecated methods usages

AnatolyRugalev:
We often copy-paste test code to implement new tests and with this we also increase the amount of incorrect/deprecated method calls.

This is what had been done in this PR:

1. Updated all `setUser` calls to `connectUser`
2. Updated all `updateUser(s)` calls to `upsertUser(s)`
3. Fixed extra `await` for `getTestClient` non-async calls
4. Fixed `connectUser` warning when running integration tests in CI
5. Removed deprecated and unused `webhook_id` and `webhook_failed` fields from `MessageResponse`

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,13 +138,16 @@ Custom types are carried into all creation functions as well.
 
 ```typescript
 // Valid
-client.setUser({ id: 'testId', nickname: 'testUser', age: 3 }, 'TestToken');
-client.setUser({ id: 'testId', nickname: 'testUser', avatar: 'testAvatar' }, 'TestToken');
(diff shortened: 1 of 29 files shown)


AnatolyRugalev:
Integration tests passed :heavy_check_mark:

ruudniew (approved):
Noice. LGTM.

PS: We might want to deprecate `setGuestUser` and add a `connectGuestUser` function later. When we do, we should make sure we update the integration tests with it.

vishalnarkhede (approved):

PR #580, opened 2021-01-11
Expose channel cid on messages

ruudniew:


src/types.ts
```diff
@@ -1483,6 +1484,7 @@ export type MessageBase<
   id: string;
   pinned: boolean;
   attachments?: Attachment<AttachmentType>[];
+  cid?: string;
```
mahboubii, line 1487: isn't this part of the MessageResponse type? we don't expect the user to set this

PR #582, opened 2021-01-12
[CHAT-1471] Allowed/blocked file extensions and content types

miagilepner:
# Submit a pull request

## CLA


src/types.ts
```diff
@@ -1251,12 +1253,24 @@ export type AppSettings = {
   disable_auth_checks?: boolean;
   disable_permissions_checks?: boolean;
   enforce_unique_usernames?: 'no' | 'app' | 'team';
+  file_upload_config?: {
+    allowed_file_extensions?: string[];
+    allowed_mime_types?: string[];
```
vishalnarkhede, line 1258: Just to confirm, do we have some exhaustive list of all these types/extensions? @miagilepner

helloworld.md
```diff
@@ -0,0 +1 @@
+# Hello World!
```
vishalnarkhede, line 1:
```suggestion
```

src/types.ts
miagilepner: We don't. We could make an exhaustive list (e.g. for the media types the full list is here: https://www.iana.org/assignments/media-types/media-types.xhtml) and do validation, if you think this should be prioritized.

vishalnarkhede: Not necessary tbh. If backend is prepared to handle any wierd type ... then its totally fine!!

vishalnarkhede (approved):
Lets wait for @nhannah 's  review as well :)

src/types.ts
mahboubii: let's put a comment in here with the link to the accepted types

ferhatelmas (approved):
LGTM (for tests)

src/types.ts
```diff
@@ -1460,6 +1475,13 @@ export type Field = {
   value?: string;
 };
 
+export type FileUploadConfig = {
```
ferhatelmas, line 1478: why don't we use this type above ?

miagilepner: was following the way that it's done for other objects, which repeat fields rather than using the type.

ferhatelmas: any example for a reference ?

miagilepner: Ah, you're right. I thought firebase_config and apn_config did this, but I see now they have slightly different fields in AppSettings. I'll fix that.

ferhatelmas: Yes, they are different slightly but here is exactly same. Thanks for updating.

ferhatelmas (approved):
LGTM

nhannah (approved):

PR #586, opened 2021-01-15
Fix for member based channel

vishalnarkhede:
# Submit a pull request

## CLA

**Existing logic:**
We maintain a list of active channels on client - `client.activeChannel`. 

So every-time, user requests a channel using `client.channel('messaging', 'random_channel_id')`, we first [check](https://github.com/GetStream/stream-chat-js/blob/5f1669925f3ff22e4948f28bd0e81aa218aafe6a/src/client.ts#L1550) if the `cid` already exist in `client.activeChannels`
  |--> if yes, then return the channel available on client.activeChannels
  |--> if not, then create new instance of Channel class and return that.

This way we ensure that user doesn't end up creating multiple copies of same channel. And this is important, because when client receives ws event, it delegates it to all channels in `activeChannel` map. So if your channel isn't part of `activeChannels`, it won't receive any of the events.

**Issue**

This logic only worked for id based channels. If user requests a member based channel (e.g., `client.channel('messaging', { members: ['vishal', 'amin'] })`), then we skipped that check, because the logic for generating id from members list is on backend side, and its not a good idea to expose it (checked with @thesyncim ) on frontend. So we couldn't really check if the channel already exists in `client.activeChannels`

**Fix**

We are adding this missing check for member based channels.

For member based channel, we don't get cid until channel is watched/queried. So until we get the key from backend, we are going to store it in `activeChannels` with the key `${channelType}:${members.sort().join(',')}`. E.g. `messaging:amin,vishal` or `messaging:amin,jaap,tom`. So that we can ensure uniqueness even if channel is not queries/watched yet.

We assert that a particular channel in `activeChannels` is same as requested channel by user:

- key === `${channelType}:${members.sort().join(',')}`

OR

- If key starts with `${channelType}:!members-` AND
- channel has same members as the channel which is requested.


src/client.ts
```diff
@@ -1513,6 +1513,21 @@ export class StreamChat<
     }
     // support channel("messaging", {options})
     if (typeof channelIDOrCustom === 'object') {
```
vishalnarkhede, line 1515: **Existing logic:**
We maintain a list of active channels on client - `client.activeChannel`. 

So every-time, user requests a channel using `client.channel('messaging', 'random_channel_id')`, we first [check](https://github.com/GetStream/stream-chat-js/blob/5f1669925f3ff22e4948f28bd0e81aa218aafe6a/src/client.ts#L1550) if the `cid` already exist in `client.activeChannels`
--> if yes, then return the channel available on client.activeChannels
--> if not, then create new instance of Channel class and return that.

This way we ensure that user doesn't end up creating multiple copies of same channel. And this is important, because when client receives ws event, it delegates it to all channels in `activeChannel` map. So if your channel isn't part of `activeChannels`, it won't receive any of the events.

**Issue**

This logic only worked for id based channels. If user requests a member based channel (e.g., `client.channel('messaging', { members: ['vishal', 'amin'] })`), then we skipped that check, because the logic for generating id from members list is on backend side, and its not a good idea to expose it (checked with @thesyncim ) on frontend. So we couldn't really check if the channel already exists in `client.activeChannels`

**Fix**

We are adding this check for member based channels. We assert that a particular channel in `activeChannels` is same as requested channel by user:

1. If `cid` of channel starts with `${channelType}:!members-`
2. channel has same members as the channel which is requested.

test/integration/channels.js
```diff
@@ -2975,3 +2975,35 @@ describe('Channel - isUpToDate', async () => {
...
+		const userIdAmin = 'amin';
+		await createUsers([userIdVish, userIdAmin]);
+
+		const clientVish = await getTestClientForUser(userIdVish);
+		const channelVish_copy1 = clientVish.channel('messaging', {
+			members: ['amin', 'vishal'],
+		});
+		await channelVish_copy1.watch();
+		const channelVish_copy2 = clientVish.channel('messaging', {
+			members: ['amin', 'vishal'],
```
ferhatelmas, line 3005: better to change the order

vishalnarkhede: Done :)

test/integration/channels.js
```diff
@@ -2975,3 +2975,35 @@ describe('Channel - isUpToDate', async () => {
 		).to.be.equal(-1);
 	});
 });
+
+describe('Ensure single channel per cid on client', async () => {
```
mahboubii, line 2979:
```suggestion
describe('Ensure single channel per cid on client activeChannels state', async () => {
```

test/integration/channels.js
```diff
@@ -2975,3 +2975,35 @@ describe('Channel - isUpToDate', async () => {
...
+		await createUsers([userIdVish, userIdAmin]);
+
+		const clientVish = await getTestClientForUser(userIdVish);
+		const channelVish_copy1 = clientVish.channel('messaging', {
+			members: ['amin', 'vishal'],
+		});
+		await channelVish_copy1.watch();
+		const channelVish_copy2 = clientVish.channel('messaging', {
+			members: ['vishal', 'amin'],
+		});
```
mahboubii, line 3006: kinda an edge case to this fix which itself is an edge case but if you call `channel.watch` or `channel.create` after instantiating the second channel you still end up with a duplicate channel in the state. (this can actually happen if different parts of the app instantiate channels at the same time)

```js
const channelVish_copy1 = clientVish.channel('messaging', { members: ['amin', 'vishal'] });
const channelVish_copy2 = clientVish.channel('messaging', { members: ['amin', 'vishal'] });
// channelVish_copy1 !== channelVish_copy2

await channelVish_copy1.watch();
await channelVish_copy2.watch();

// channelVish_copy1 !== channelVish_copy2

```

vishalnarkhede: hmm probably we should early return from `watch()` function if its already being watched
```
watch() {
  if (this.initialized) {
    return;
  }

  ... // rest of the logic
}
```

mahboubii: yea it's a potential fix but there could be changes in the option param which cause other issues. To cover this we need to make some changes to the new code you implemented to also check for channels that don't have their `cid` yet but have the same member list

vishalnarkhede: Makes sense!! Have updated the PR.

So when channel doesn't have `cid` yet, I am storing it in `activeChannels` as following:

```js
{
  "messaging:amin,vishal": uniqueChannelAminVishal,
  "messaging:k4j2n3k4j2n3kn42k3": channelWithId
}
```

When cid gets set on channel, I replace `messaging:amin,vishal` with actual cid generated from backend.

```js
{
  "messaging:!members:kajshdkjsakjsdkjas": uniqueChannelAminVishal,
  "messaging:k4j2n3k4j2n3kn42k3": channelWithId
}
```

Also updated tests to reflect this case :)

mahboubii (approved):

ferhatelmas (approved):
LGTM

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
-        `Invalid channel id ${channelIDOrCustom}, can't contain the : character`,
-      );
+    return this.getChannelById(channelType, channelIDOrCustom, custom);
+  }
+
+  /**
+   * Its a submethod for `client.channel()` method, used to create unique conversation or
+   * channel based on member list instead of id.
+   *
+   * If the channel already exist in `activeChannels` list, then we simply return it, since that
```
ferhatelmas, line 1527: exists

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
+   * means the same channel was already requested or created.
+   *
+   * Otherwise we create new instance of Channel class and return it.
```
ferhatelmas, line 1530: a new

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
 
-    if (typeof channelIDOrCustom === 'string' && ~channelIDOrCustom.indexOf(':')) {
-      throw Error(
-        `Invalid channel id ${channelIDOrCustom}, can't contain the : character`,
-      );
+    return this.getChannelById(channelType, channelIDOrCustom, custom);
+  }
+
+  /**
+   * Its a submethod for `client.channel()` method, used to create unique conversation or
```
ferhatelmas, line 1524: It's

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
+    // Only allow 1 channel object per cid
+    const membersStr = custom.members?.sort().join(',');
+    if (!membersStr) {
+      throw Error('Please specify atleast one member when creating unique conversation');
+    }
+
+    // channel could exist in `activeChannels` list with either one of the following two keys:
+    // 1. cid - Which gets set on channel only after calling channel.query or channel.watch or channel.create
+    // 2. Sorted membersStr - E.g., "messaging:amin,vishal" OR "messaging:amin,jaap,tom"
+    //                        This gets set when you create a channel, but haven't queried yet. After query,
```
ferhatelmas, line 1548: s/gets/is/

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
+    >(this, channelType, undefined, custom);
+
+    // For the time being set the key as membersStr, since we don't know the cid yet.
+    // In channel.query, we will replace it with 'cid'.
+    this.activeChannels[`${channelType}:${membersStr}`] = channel;
+    return channel;
+  };
+
+  /**
+   * Its a submethod for `client.channel()` method, used to channel given the id of channel.
```
ferhatelmas, line 1583: It's

src/client.ts
```diff
@@ -1511,27 +1511,99 @@ export class StreamChat<
...
+   *
+   * If the channel already exist in `activeChannels` list, then we simply return it, since that
```
ferhatelmas, line 1585: exists

src/client.ts
ferhatelmas: `submethod` doesn't intuitive, `a helper` maybe

mahboubii: also, let's add a `@private` to these two functions

vishalnarkhede: Done :)

vishalnarkhede:
@mahboubii added `!members-` to membersStr !!

src/channel.ts
```diff
@@ -1249,6 +1249,12 @@ export class Channel<
       this.id = state.channel.id;
       this.cid = state.channel.cid;
       // set the channel as active...
+      const membersStr = Object.keys(this.state.members)?.sort().join(',');
+
+      if (`${this.type}:!members-${membersStr}` in this.getClient().activeChannels) {
+        delete this.getClient().activeChannels[membersStr];
```
mahboubii, line 1255: should we delete the first instance from the activeChannels?

vishalnarkhede: What do you mean? Its a map, so there will be only one instance.

But just noticed that we are not deleting the correct key from activeChannels. It should be something like:
```js
delete this.getClient().activeChannels[`${this.type}:!members-${membersStr}`];
```

Gonna add to test as well :)

vishalnarkhede:
Moved tests to unit-test :)

src/channel.ts
vishalnarkhede: Done !!

PR #591, opened 2021-01-21
Fix member ordering for creating member based channels

ferhatelmas:
since it's sorted now.

### DIFF
--- a/test/integration/channels.js
+++ b/test/integration/channels.js
@@ -14,7 +14,6 @@ import {
 } from './utils';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-import { get } from 'https';
 
 const expect = chai.expect;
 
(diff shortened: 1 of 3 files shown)


ferhatelmas:
Btw, it is a unit test than an integration test, we could remove it too since covered.

test/integration/channels.js
```diff
@@ -123,7 +123,7 @@ describe('Channels - Constructor', function () {
 
 	it('short version with options', function (done) {
 		const channel = client.channel('messaging', { members: ['tommaso', 'thierry'] });
-		expect(channel.data).to.eql({ members: ['tommaso', 'thierry'] });
+		expect(channel.data).to.eql({ members: ['thierry', 'tommaso'] });
```
mahboubii, line 126: this test seems to be a better fit for unit tests. is this due to recent changes? the member order should remain the same

ferhatelmas: yes, it comes from member based channel creation fix.

I agree with unit.

vishalnarkhede (approved):

test/integration/channels.js
mahboubii: oh I see, the sort is applied in place, I think it's better to revert that change

ferhatelmas: I will make a fix to sort the copy and move this test to unit tests.

mahboubii: @ferhatelmas https://github.com/GetStream/stream-chat-js/pull/592

ferhatelmas: ahh, updated this one

mahboubii (approved):
🌮

PR #594, opened 2021-01-21
update token provider doc

mahboubii:
# Submit a pull request

## CLA

### DIFF
--- a/docs/userToken.md
+++ b/docs/userToken.md
@@ -20,6 +20,13 @@ Make sure the TokenProvider function always resolves, you can wrap your async fu
 
 ```js
 const retry = require('async-retry');
+
+const fetchTokenFromApi = async () => {
+  const response = await fetch('https://my-api.com/token');
+  const data = await response.json();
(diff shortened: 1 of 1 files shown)


docs/userToken.md
```diff
@@ -1,14 +1,17 @@
...
 ```js
 const client = new StreamChat('api_key');
 client.connectUser({ id: 'vishal' }, 'user_token_string');
 ```
 
 ## Token Provider
 
+You can use this feature to use short-lived token within your app. When the token expires, the WS connection will get a new token from the Token Provider and reconnect.
+
+Make sure the TokenProvider function always resolves, you can wrap your async function with an async-retryable to not reject the promise immediately in case of a network issue.
```
ferhatelmas, line 13: do you think providing an example for this scenario would be better ?

mahboubii: yea adding

vishalnarkhede (approved):
LGTM

PR #595, opened 2021-01-21
Add custom Safari header

ferhatelmas:


Royalkorr (approved):
Web

src/client.ts
```diff
@@ -235,6 +236,11 @@ export class StreamChat<
     }
 
     this.axiosInstance = axios.create(this.options);
+    if (this.options.safari || isSafari()) {
```
vishalnarkhede, line 239: Whats the purpose of `isSafari()`? I can see that it always returns false

ferhatelmas: We released this for a beta so that we can test with customer. If works, we will implement detection. Right now, it bypasses with an option for testing.

ferhatelmas:
done without client cooperation, only server side detection was enough.

PR #599, opened 2021-01-27
[CJS3] client single instance

mahboubii:
Unfortunately Lots of TS boilerplates due to generics.

Potential Future improvements: `connectUser` can return the resolved promise for the second call instead of throwing an error if the user id in the token remains the same

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -104,6 +104,8 @@ export class StreamChat<
   ReactionType extends UnknownType = UnknownType,
   UserType extends UnknownType = UnknownType
 > {
+  private static _instance: StreamChat;
+
   _user?: OwnUserResponse<ChannelType, CommandType, UserType> | UserResponse<UserType>;
   activeChannels: {
     [key: string]: Channel<
@@ -171,6 +173,8 @@ export class StreamChat<
 
   /**
    * Initialize a client
+   *
+   * **Only use constructor for advanced usages. It is strongly advised to use `StreamChat.getInstance()` instead of `new StreamChat()` to reduce integration issues due to multiple WebSocket connections**
    * @param {string} key - the api key
    * @param {string} [secret] - the api secret
    * @param {StreamChatOptions} [options] - additional options, here you can pass custom options to axios instance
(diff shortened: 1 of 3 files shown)


test/unit/client.js
```diff
@@ -3,6 +3,35 @@ import { StreamChat } from '../../src/client';
...
+		expect(client2).to.equal(client3);
+	});
+
+	it('changin params has no effect', function () {
+		const client1 = StreamChat.getInstance('key2');
+		const client2 = StreamChat.getInstance('key3');
+
+		expect(client1).to.equal(client2);
+		expect(client2.key).to.eql('key2');
+	});
```
vishalnarkhede, line 32: Can we also add following scenario to unit-test?

```js
const client1 = StreamChat.getInstance('key2');
await client1.setUser({ id: 'vishal });
const client2 = StreamChat.getInstance('key3'); // or key3
await client2.setUser({ id: 'vishal });

```

src/client.ts
```diff
@@ -312,6 +316,125 @@ export class StreamChat<
...
+        StreamChat._instance = new StreamChat<
+          AttachmentType,
+          ChannelType,
+          CommandType,
+          EventType,
+          MessageType,
+          ReactionType,
+          UserType
+        >(key, secretOrOptions);
+      }
```
vishalnarkhede, line 424: Do we need this if-else? StreamChat constructor handles this logic as well

test/unit/client.js
vishalnarkhede: Currently if you call setUser twice, it throws following error:

```
'Use client.disconnect() before trying to connect as a different user. connectUser was called twice.'
```

But I think we should allow it if current user is same as id provided to `setUser`

mahboubii: pushed a test https://github.com/GetStream/stream-chat-js/pull/599/commits/d2fcfc3394cdc0501db4dc7e3aa5f7e4363324e9

src/client.ts
mahboubii: from JS perspective we don't need this, but due to fn overrides this seems the way to have a happy TS compiler

test/unit/client.js
mahboubii: @vishalnarkhede yea I actually put a comment in the PR description about this which is out of the scope of this PR. This is still an integration mistake but we can make it easier for users by not throwing error

vishalnarkhede: Ahaa ... sorry missed it!!

Can we add it though ... seems like a simple check

nhannah (approved):

vishalnarkhede (approved):

test/unit/client.js
mahboubii: yea totally will open another PR on top of this one

PR #602, opened 2021-01-28
CJS-5: Remove immutable

virdesai:
# Submit a pull request

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices
- [ ] Code changes are tested (add some information if not applicable)

PR #606, opened 2021-02-01
[CHAT-1565] Sync own_reactions in events

AnatolyRugalev:
This PR fixes the issue of synchronizing `own_reactions` in incoming events.

The original problem comes from the fact that `own_reactions` field is not populated for each individual client when broadcasting the event. But fortunately, we already sync this data in the channel state, so replacing related fields in the event object seems like a reasonable solution.


src/channel_state.ts
```diff
@@ -536,11 +539,41 @@ export class ChannelState<
...
+  _syncReactions(
+    eventMessage: MessageResponse<
+      AttachmentType,
+      ChannelType,
+      CommandType,
+      MessageType,
+      ReactionType,
+      UserType
+    >,
+    stateMessage: Immutable.Immutable<
```
virdesai, line 557: https://github.com/GetStream/stream-chat-js/pull/602 getting rid of this Immutable stuff here

vishalnarkhede (commented):
@AnatolyRugalev So the change looks fine to me. I have a suggestion:

With current implementation we mutate the event object inside channelState class. And thats a little confusing. If we override the event object, I think it should be done in `channel.ts`, where we handle the event. So something like following:

```js
_handleChannelEvent(event) {
  ...
  case 'reaction.new':
    if (event.reaction) {
      /**
       * richReactionsData = {
       *  latest_reactions: {},
       *  own_reactions: {},
       *  reactions_count: {}
       * }
       */
      const richReactionsData = s.addReaction(event.reaction, event.message);
      event.message = { ...event.message, ...richReactionsData};
    }
}
```

This is assuming we are shipping it with https://github.com/GetStream/stream-chat-js/pull/602

@mahboubii what do you think?

AnatolyRugalev:
@vishalnarkhede I was trying to do this be returning entire message from `addReaction` and `removeReaction` methods, but it tuned out challenging with all this immutable stuff. Let's wait for #602 to merge and I will get back to addressing this point as you suggested :+1:

virdesai:
@AnatolyRugalev heads up that #602 was just merged into master

AnatolyRugalev:
I am still working on this. More improvements are coming:
* correct reaction_counts sync
* on enforce_unique counters could go out of sync if we have more than 1 reaction from this user
* some reactions sync code improvements

AnatolyRugalev:
After some experimenting I came to the conclusion that we don't need to sync `latest_reactions`, `reaction_counters` and `reaction_scores` manually. It was tempting to do, but in some edge cases this just wouldn't work properly (e.g. reaction.updated event for a reaction that is not loaded to `latest_reactions`).

Incoming events contain this fields, so we simply update the message with them in the state. `own_reactions` field though needs to be synchronized manually because of the reasons given above.

vishalnarkhede:
I will review this tomorrow :)

src/channel.ts
```diff
@@ -1616,19 +1616,19 @@ export class Channel<
...
         if (event.reaction) {
-          channelState.removeReaction(event.reaction, event.message);
+          event.message = channelState.removeReaction(event.reaction, event.message);
         }
         break;
       case 'reaction.updated':
         if (event.reaction) {
           // assuming reaction.updated is only called if enforce_unique is true
-          channelState.addReaction(event.reaction, event.message, true);
+          event.message = channelState.addReaction(event.reaction, event.message, true);
```
ferhatelmas, line 1631: why is this mutation needed ?

AnatolyRugalev: We want subscribers of reaction events to receive correct `own_reactions` field. We synchronise contents of this fields on the client side, so in order to propagate updated field to subscribers, we need to update it. Updating message object  implicitly in channels state didn't seem right on the previous code review (see @vishalnarkhede comment above)

ferhatelmas: Got it, LGTM, thanks

ferhatelmas (approved):
LGTM

A bit harder to follow the call but dropped a lot of repetition, I think it's a great improvement in code quality.

ferhatelmas:
@AnatolyRugalev might want to update PR title since only `own_reactions` are synced.

PR #608, opened 2021-02-02
Add skip push flag into send message/reaction

ferhatelmas:
If flag is true, push notification is not sent.


src/channel.ts
```diff
@@ -414,9 +414,12 @@ export class Channel<
     if (!reaction || Object.keys(reaction).length === 0) {
       throw Error(`Reaction object is missing`);
     }
+
+    const { skip_push, ...rest } = reaction;
```
mahboubii, line 418: a bit weird to put this into the reaction object, maybe we should change the function signature to `sendReaction(id, reaction, options)`? we have a major release coming up so we can unify this across all similar function making them future proof

ferhatelmas: nice, if we can change signature, I would prefer it too.

mahboubii: what do you think @vishalnarkhede?

ferhatelmas: will do this change and will also introduce options to `sendMessage` for the same reason (not to mix control keys with custom fields). Sounds good ?

mahboubii: sounds good

ferhatelmas: updated

mahboubii (approved):
🌮

PR #611, opened 2021-02-05
Introduce options for sendReaction

ferhatelmas:
Since immutable is breaking, it's right moment to add more breaking changes such that we switch `sendReaction` similar to `sendMessage` and can have different/new options.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -395,29 +395,20 @@ export class Channel<
    *
    * @param {string} messageID the message id
    * @param {Reaction<ReactionType, UserType>} reaction the reaction object for instance {type: 'love'}
-   * @param {string} [user_id] the id of the user (used only for server side request) default null
-   * @param {boolean} [enforce_unique] set true to overwrite existing reactions if any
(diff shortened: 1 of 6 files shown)


AnatolyRugalev (approved):
LGTM

src/channel.ts
```diff
@@ -395,30 +395,21 @@ export class Channel<
...
+   * @param {{ enforce_unique?: boolean }} [options] Option object, {enforce_unique: true} to override any existing reaction
    *
    * @return {Promise<ReactionAPIResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} The Server Response
    */
   async sendReaction(
     messageID: string,
     reaction: Reaction<ReactionType, UserType>,
-    user_id?: string,
-    enforce_unique?: boolean,
+    options?: { enforce_unique?: boolean },
```
mahboubii, line 405: missing user_id

AnatolyRugalev: user_id is moved to reaction object

ferhatelmas: I did intentionally similar to `sendMessage`.

mahboubii (approved):

PR #629, opened 2021-03-08
Support reactions sync for pinned messages in channel state

AnatolyRugalev:
This PR adds reaction sync for pinned messages in channel state


ferhatelmas (approved):
LGTM

PR #631, opened 2021-03-08
Add support for GetRateLimits

ruggi:
## CLA

This PR introduces support for the GetRateLimits endpoint.
Note: related integration tests to be added on the QA repo.

## Changelog

- Added the `StreamClient.getRateLimits` function to retrieve rate limit quotas and usage, with the option to filter per platform and endpoint.


ferhatelmas (approved):
LGTM

src/types.ts
```diff
@@ -1776,3 +1776,20 @@ export type TranslationLanguages =
...
+
+export type RateLimitsInfo = {
+  limit: number;
+  remaining: number;
+  reset: number;
+};
+
+export type RateLimitsMap = {
+  [key: string]: RateLimitsInfo;
+};
```
vishalnarkhede, line 1788: @ruggi Do we know exactly what key could be? If yes, we can add more strong types:

```ts
export type RateLimitsType = "key1" | "key2" | "key3";
export type RateLimitsMap = Record<RateLimitsKey, RateLimitsInfo>

```

ruggi: it's one key per endpoint, so the problem here is keeping that in sync

vishalnarkhede: If it's not going to change frequently (weekly/monthly basis), we can just build types it for once. If you can point me to code in chat repo, I can add it tomorrow :)

vishalnarkhede: @nhannah what do you think? Worth it?

src/client.ts
```diff
@@ -1591,6 +1592,33 @@ export class StreamChat<
...
 
+  /**
+   * getRateLimits - Returns the rate limits quota and usage for the current app, possibly filter for a specific platform and/or endpoints.
+   * Only available server-side.
+   *
+   * @param {object} params The params for the call. If none of the params are set, all limits for all platforms are returned.
+   */
+  async getRateLimits(params?: {
+    android?: boolean;
+    endpoints?: string[];
```
mahboubii, line 1603: we need to type all the different endpoint names that we support and keep them in sync here, it can also be reused on the response type

src/types.ts
```diff
@@ -1776,3 +1776,20 @@ export type TranslationLanguages =
...
+
+export type GetRateLimitsResponse = {
+  android?: RateLimitsMap;
+  ios?: RateLimitsMap;
+  server_side?: RateLimitsMap;
+  web?: RateLimitsMap;
```
mahboubii, line 1794: so we have all these different rate limits per each endpoint? are some of them omitted from response or will be just set to default?

ruggi: By default the endpoint returns all limits for all endpoints and all platforms (web, android, ios, server-side); it can be narrowed to just some of the platforms (with the booleans) and/or specific endpoints (with the `string[]` endpoints param)

src/types.ts
```diff
@@ -1777,15 +1777,98 @@ export type TranslationLanguages =
   | 'zh'
   | 'zh-TW';
 
+export type EndpointNames =
```
ruggi, line 1780: maybe `EndpointName` would be better (singular <> plural)

mahboubii: good point, fixed

src/client.ts
```diff
@@ -1591,6 +1593,30 @@ export class StreamChat<
...
+   */
+  async getRateLimits(params?: {
+    android?: boolean;
+    endpoints?: EndpointName[];
+    ios?: boolean;
+    serverSide?: boolean;
+    web?: boolean;
+  }) {
+    const { serverSide, web, android, ios, endpoints } = params || {};
+    return this.get<APIResponse & GetRateLimitsResponse>(this.baseURL + '/rate_limits', {
```
vishalnarkhede, line 1611: Can we create a separate type for this?

`export type GetRateLimitsApiResponse = APIResponse & GetRateLimitsResponse`

mahboubii: https://github.com/GetStream/stream-chat-js/pull/631/commits/8b02e222816d9ffbaddccb5ab7c1a0a2920bbece

vishalnarkhede (approved):

ferhatelmas (approved):
LGTM

PR #633, opened 2021-03-09
Update PULL_REQUEST_TEMPLATE.md

mahboubii:
we should start consolidating these templates across the projects


vishalnarkhede (dismissed):
Would be even more amazing, if one could not open a PR without writing something in Description.

ferhatelmas:
Idea is good but I have couple of suggestions:

* let's setup cla bot and require a status check: https://cla-assistant.io/
* let's keep github related content in `.github` folder, not in the root

mahboubii:
@ferhatelmas CLA assistant tool requires some access and needs to be confirmed before we can add it. We can create a custom action to check the CLA though, it can also reject the PRs without any description.

ferhatelmas (approved):

vishalnarkhede (approved):

PR #637, opened 2021-03-09
Pass endpoints as comma-separated values

ruggi:
This PR follows from https://github.com/GetStream/stream-chat-js/pull/631 updating the way endpoints are passed via query params, as a list of comma-separated values on the same key.


ferhatelmas (approved):
LGTM

mahboubii (approved):

PR #639, opened 2021-03-11
fix: moving channel._disconnect call from closeConnection to disconnectUser

vishalnarkhede:
## CLA

- [ ] Code changes are tested

`channel._disconnect` should only happen in `disconnectUser`, and not `closeConnection`. Because once you call channel._disconnect, it will be completely blocked of all the functionality. This issue was introduced in latest minor version. We need to think a bit more about this disconnect functionality, but for now to unblock react/react-native SDKs. This should fix it.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -556,10 +556,6 @@ export class StreamChat<
       this.cleaningIntervalRef = undefined;
     }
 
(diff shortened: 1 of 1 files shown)


ferhatelmas (approved):
LGTM

PR #644, opened 2021-03-15
Handle given base url from options

ferhatelmas:

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -245,11 +245,7 @@ export class StreamChat<
 
     this.axiosInstance = axios.create(this.options);
 
-    if (this.options.baseURL) {
-      this.setBaseURL(this.options.baseURL);
-    } else {
-      this.setBaseURL('https://chat-us-east-1.stream-io-api.com');
-    }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -245,7 +245,11 @@ export class StreamChat<
 
     this.axiosInstance = axios.create(this.options);
 
-    this.setBaseURL('https://chat-us-east-1.stream-io-api.com');
+    if (this.options.baseURL) {
```
vishalnarkhede, line 248: Let's also explicitly add it to `StreamChatOptions` type!!

vishalnarkhede (approved):

src/client.ts
```diff
@@ -245,7 +245,11 @@ export class StreamChat<
...
+      this.setBaseURL(this.options.baseURL);
+    } else {
+      this.setBaseURL('https://chat-us-east-1.stream-io-api.com');
+    }
```
mahboubii, line 252:
```suggestion
    this.setBaseURL(this.options.baseURL || 'https://chat-us-east-1.stream-io-api.com');
```

vishalnarkhede (approved):

PR #647, opened 2021-03-17
Ensure mentioned_users is array of user_ids. in updateMessage

vishalnarkhede:
## CLA

- [ ] Code changes are tested

Slack discussion - https://getstream.slack.com/archives/C01FT5767AS/p1615968119001100

Updating pinned message fails, if there is mentioned_users in message. This is something missed in react sdk, (to ensure conversion of array of user objects to user ids). I think its better we handle it on js client as well.

## Changelog

- Extra check to see if `message.mentioned_users` is array of string or objects
- Convert array of user objects to array of user ids.
- Unit tests for the change

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2422,24 +2422,15 @@ export class StreamChat<
 
     /**
      * Server always expects mentioned_users to be array of string. We are adding extra check, just in case
-     * SDK missed this conversation.
+     * SDK missed this conversion.
      */
     if (
       Array.isArray(clonedMessage.mentioned_users) &&
       !isString(clonedMessage.mentioned_users[0])
     ) {
-      /**
-       * We need to typecast mentioned_users first to unknown, to let typescript know that further typecasting is intentional.
-       *
-       * Error raised by ts:
-       * ```
-       * Conversion of type 'string[]' to type 'UserResponse<UserType>[]' may be a mistake because neither
-       * type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first
-       * ```
-       */
       clonedMessage.mentioned_users = ((clonedMessage.mentioned_users as unknown) as Array<
         UserResponse<UserType>
-      >)?.map((mu) => mu.id);
+      >)?.map((mu) => ((mu as unknown) as UserResponse).id);
     }
 
     return await this.post<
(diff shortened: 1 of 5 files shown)


mahboubii (dismissed):

AnatolyRugalev (dismissed):

src/client.ts
```diff
@@ -2419,6 +2419,29 @@ export class StreamChat<
...
       }
     }
+
+    /**
+     * Server always expects mentioned_users to be array of string. We are adding extra check, just in case
+     * SDK missed this conversation.
+     */
+    if (
+      Array.isArray(clonedMessage.mentioned_users) &&
+      !isString(clonedMessage.mentioned_users[0])
```
ferhatelmas, line 2429: Can't we improve type detection here so that we don't need to keep following comments?

vishalnarkhede: Ideally according to our types, `clonedMessage` shouldn't be array of strings at all. But still we want to make sure that if its missed, we handle that ... so as to not break the API.

vishalnarkhede: So I think following type casting is necessary here. Not sure if we can avoid it without allowing array of user objects on `UpdatedMessage` type.

src/client.ts
```diff
@@ -2419,6 +2419,29 @@ export class StreamChat<
...
+       *
+       * Error raised by ts:
+       * ```
+       * Conversion of type 'string[]' to type 'UserResponse<UserType>[]' may be a mistake because neither
+       * type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first
+       * ```
+       */
+      clonedMessage.mentioned_users = ((clonedMessage.mentioned_users as unknown) as Array<
+        UserResponse<UserType>
+      >)?.map((mu) => mu.id);
```
mahboubii, line 2442:
```suggestion
      clonedMessage.mentioned_users = clonedMessage.mentioned_users.map(
        (mu) => ((mu as unknown) as UserResponse).id,
      );
```

src/client.ts
ferhatelmas: Sounds good, let's apply suggestion from @mahboubii then I guess we can roll with it

vishalnarkhede: Ferhat, updated!!

ferhatelmas: Thanks, do we need array type case in the final version ?

mahboubii (dismissed):

ferhatelmas (approved):
LGTM

ferhatelmas (approved):
LGTM

was forgetting `only` :smile:

vishalnarkhede:
> LGTM
> 
> was forgetting `only` 😄

lol yeah, thats typical "Vish" :D

PR #651, opened 2021-03-19
Return unsubscribe handler from `channel.on`

vishalnarkhede:
## CLA

- [ ] Code changes are tested

- Slack discussion: https://getstream.slack.com/archives/C01FT5767AS/p1616153988038900

## Changelog

- Added access to unsubscribe handler from `channel.on` function.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1380,7 +1380,7 @@ export class Channel<
       ReactionType,
       UserType
     >,
-  ): void;
+  ): { unsubscribe: () => void };
   on(
     callback: EventHandler<
       AttachmentType,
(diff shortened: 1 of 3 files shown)


DanC5 (dismissed):

src/channel.ts
```diff
@@ -1433,6 +1433,12 @@ export class Channel<
     );
 
     this.listeners[key].push(callback);
+
+    return {
+      unsubscribe: () => {
+        this.listeners[key] = this.listeners[key].filter((el) => el !== callback);
+      },
```
mahboubii, line 1440: should we call `this.off` or put the logger message to stay consistent?

vishalnarkhede: Let's go with logger message. We have `unsubscribe` on `client.on`, so lets be consistent in naming!!

mahboubii (approved):

PR #654, opened 2021-03-27
Fix ChannelState#clean() to handle received_at as Date type

kainosnoema:
## CLA

- [x] Code changes are tested

Because the client sets `received_at` as a Date type (src/client.ts#L1168) and `ChannelState#clean()` was expecting a string, it hasn't been working at all. A better fix might be to restrict the Event typing to Date since it is never a string after being handled by the client, but in lieu of that we'll handle both types here and add a test.

## Changelog

- Fix ChannelState#clean() to properly clear stale typing events

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -620,10 +620,11 @@ export class ChannelState<
     const now = new Date();
     // prevent old users from showing up as typing
     for (const [userID, lastEvent] of Object.entries(this.typing)) {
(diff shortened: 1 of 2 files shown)


mahboubii (approved):
Thanks for the fix @kainosnoema

virdesai (approved):

vishalnarkhede (approved):

PR #662, opened 2021-04-16
fixed response test for frozen update channel

gumuz:
This "fixes" a test that started failing recently, due to a backend change. However, I'm not sure what the intent of the test was and if it's still valid like this.

Tagging you @nhannah since you were the author of the test


vishalnarkhede:
Screenshot 2021-04-16 at 15 08 17

vishalnarkhede (approved):

PR #670, opened 2021-04-28
CRNS-321: Update message references for deleted or updated users

vishalnarkhede:
## CLA

- [ ] Code changes are tested

## Changelog

- Delete user messages when you receive `user.deleted` event
- Update `message.user` object, when `user.updated` event is received

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -621,8 +621,8 @@ export class ChannelState<
    * Updates the `message.user` property on messages from given user,
    * and returns the updates messages array
    *
-   * @param messages
-   * @param user
+   * @param {Array<ReturnType<ChannelState<AttachmentType, ChannelType, CommandType, EventType, MessageType, ReactionType, UserType>['formatMessage']>>} messages
+   * @param {UserResponse<UserType>} user
    */
   _updateUserMessages = (
     messages: Array<
@@ -654,7 +654,7 @@ export class ChannelState<
   /**
    * Updates the message.user property with updated user object, for messages.
    *
-   * @param user
+   * @param {UserResponse<UserType>} user
    */
   updateUserMessages = (user: UserResponse<UserType>) => {
     this.messages = this._updateUserMessages(this.messages, user);
@@ -671,9 +671,9 @@ export class ChannelState<
    *
    * Marks messages from given user as deleted, and returns updated message list
    *
-   * @param messages
-   * @param user
-   * @param hardDelete
+   * @param {Array<ReturnType<ChannelState<AttachmentType, ChannelType, CommandType, EventType, MessageType, ReactionType, UserType>['formatMessage']>>} messages
+   * @param {UserResponse<UserType>} user
+   * @param {boolean} hardDelete
    */
   _deleteUserMessages = (
     messages: Array<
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -1169,6 +1169,131 @@ export class StreamChat<
...
+    const refMap = this.state.userChannelReferences[user.id] || {};
+
+    for (const channelID in refMap) {
+      const channel = this.activeChannels[channelID];
+      const state = channel.state;
+
+      /** update the messages from this user. */
+      if (state) {
+        state?.updateUserMessages(user);
+      }
```
mahboubii, line 1217:
```suggestion
      /** update the messages from this user. */
      channel.state?.updateUserMessages(user);
```

src/channel_state.ts
```diff
@@ -603,6 +614,153 @@ export class ChannelState<
...
+          MessageType,
+          ReactionType,
+          UserType
+        >['formatMessage']
+      >
+    >,
+    user: UserResponse<UserType>,
+    hardDelete = false,
+  ) =>
+    messages.map((m) => {
```
mahboubii, line 695: for loop is way faster in this case

src/channel_state.ts
```diff
@@ -603,6 +614,153 @@ export class ChannelState<
...
+          latest_reactions: [],
+          mentioned_users: [],
+          own_reactions: [],
+          parent_id: m.parent_id,
+          reply_count: m.reply_count,
+          status: m.status,
+          thread_participants: m.thread_participants,
+          type: 'deleted',
+          updated_at: m.updated_at,
+          user: m.user,
```
mahboubii, line 719: does it make sense to extract this into a default empty msg or sth?

src/channel_state.ts
```diff
@@ -603,6 +614,153 @@ export class ChannelState<
...
+            CommandType,
+            EventType,
+            MessageType,
+            ReactionType,
+            UserType
+          >['formatMessage']
+        >;
+      } else {
+        return {
+          ...m,
```
mahboubii, line 733: what about the `user` attribute?

src/channel_state.ts
```diff
@@ -603,6 +614,153 @@ export class ChannelState<
...
+  /**
+   * @private
+   *
+   * Updates the `message.user` property on messages from given user,
+   * and returns the updates messages array
+   *
+   * @param {Array<ReturnType<ChannelState<AttachmentType, ChannelType, CommandType, EventType, MessageType, ReactionType, UserType>['formatMessage']>>} messages
+   * @param {UserResponse<UserType>} user
+   */
+  _updateUserMessages = (
```
mahboubii, line 627: this can be an inline function within the `updateUserMessages` method

mahboubii (commented):
Looks good
so much stuff is going on with this one, would be better to have some unit tests as well

src/channel_state.ts
vishalnarkhede: It will be part of `...m`

src/channel_state.ts
vishalnarkhede: I did think about it as well, but it's not exactly empty message. There are certain properties that we retain and some we remove. So didn't make sense to me to extract it out.

src/channel_state.ts
mahboubii: will it retain the old user data?

vishalnarkhede: Yeah, its supposed to!! Sounds wierd, but thats how backend behaves right now!

mahboubii (approved):

PR #672, opened 2021-05-04
[CRNS-324] Add specific event types and narrowing functionality

tsirlucas:
**WARNING: This PR includes breaking changes on the types**

I moved the events into different files (adding all the types for events would make it too big), if you only want to get the diff related to the code changes, please compare after first commit.

- Moves types to types folder, splitting them into files
- Adds types for all events declared in the internal spreadsheet (Jira task has the link)
- Adds some specific properties based on the Android SDK implementation
- Adds new EventUnion type which is a union from all event types
- Removes old exported `Event` type in order to use new `StreamEvent` to avoid problems with users using the default `Event` type from the system

Types names are based on the event name, which means the `foo.bar` event would have a `FooBarEvent` type. I ended up changing the existent `ConnectionChangeEvent` to `ConnectionChangedEvent`. This will make things easier for users when trying to figure out which event type they should import.

Also created the `Event base types`: Base types that are extended inside the regular event types e.g. `EventWithReaction` is a type object with the `reaction` property. All events that have the reaction entity included should extend this base type. They all follow the same naming pattern `EventWith[entity_here]` e.g. `EventWithMember` contains the `member` property.


vishalnarkhede:
@tsirlucas can we retain the old `Event` type as well?

vishalnarkhede:
Its never a good idea to introduce breaking change in this package. I know that generally in semantic versioning you can introduce breaking change in major release, and we tried that previous but got a lot of complains and issues because of it.

PR #677, opened 2021-05-10
[CHAT-1814] Updates to message search

miagilepner:
## CLA

- [x] Code changes are tested

Updates the `channel.search()` and `client.search()` endpoints to support pagination using a next value and sorting parameters. For example: 
```js
// get a page of sorted results
const page1 = await client.search(
{ cid: my_cid }, 
'search text', 
{ limit: 100, sort: [{ relevance: -1 }, { updated_at: 1 }, { my_custom_field: -1 }]  });

// Query the second page of results. The same sort order will be used as in your first query, so there's no need to include it
const page2 = await client.search(
{ cid: my_cid }, 
'search text', 
{ limit: 100, next: page1.next });

// Pagination backwards by using the returned `previous` value
const page1Again = await client.search(
{ cid: my_cid }, 
'search text', 
{ limit: 100, next: page2.previous });
```
## Changelog

-

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -332,7 +332,7 @@ export class Channel<
     } = {},
   ) {
     // Return a list of channels
-    const { sort: sort_value, ...options_without_sort } = { ...options };
+    const { sort, ...optionsWithoutSort } = { ...options };
     const payload: SearchPayload<
       AttachmentType,
       ChannelType,
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -331,7 +331,7 @@ export class Channel<
     } = {},
   ) {
     // Return a list of channels
-    const { sort, ...options_without_sort } = { ...options };
+    const { sort: sort_value, ...options_without_sort } = { ...options };
     const payload: SearchPayload<
       AttachmentType,
       ChannelType,
@@ -355,8 +355,8 @@ export class Channel<
       throw Error(`Invalid type ${typeof query} for query parameter`);
     }
 
-    if (sort) {
-      payload.sort = normalizeQuerySort(sort);
+    if (sort_value) {
+      payload.sort = normalizeQuerySort(sort_value);
     }
     // Make sure we wait for the connect promise if there is a pending one
     await this.getClient().wsPromise;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -44,6 +44,7 @@ import {
   UserFilters,
   UserResponse,
   UserSort,
+  SearchMessageSort,
 } from './types';
 import { Role } from './permissions';
 
@@ -317,7 +318,7 @@ export class Channel<
           UserType
         >
       | string,
-    options: SearchOptions & {
+    options: SearchOptions<MessageType> & {
       client_id?: string;
       connection_id?: string;
       message_filter_conditions?: MessageFilters<
@@ -331,8 +332,11 @@ export class Channel<
       query?: string;
     } = {},
   ) {
+    if ('offset' in options && ('sort' in options || 'next' in options)) {
+      throw Error(`Cannot specify offset with sort or next parameters`);
+    }
     // Return a list of channels
-    const { sort, ...optionsWithoutSort } = { ...options };
+    const { sort: sortValue, ...optionsWithoutSort } = { ...options };
     const payload: SearchPayload<
       AttachmentType,
       ChannelType,
@@ -355,9 +359,8 @@ export class Channel<
     } else {
       throw Error(`Invalid type ${typeof query} for query parameter`);
     }
-
-    if (sort) {
-      payload.sort = normalizeQuerySort(sort);
+    if (sortValue) {
+      payload.sort = normalizeQuerySort<SearchMessageSort<MessageType>>(sortValue);
     }
     // Make sure we wait for the connect promise if there is a pending one
     await this.getClient().wsPromise;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -44,7 +44,7 @@ import {
   UserFilters,
   UserResponse,
   UserSort,
-  SearchMessageSort,
+  SearchMessageSortBase,
 } from './types';
 import { Role } from './permissions';
 
@@ -350,7 +350,9 @@ export class Channel<
         UserType
       >,
       ...options,
-      sort: options.sort ? normalizeQuerySort<SearchMessageSort<MessageType>>(options.sort) : undefined,
+      sort: options.sort
+        ? normalizeQuerySort<SearchMessageSortBase<MessageType>>(options.sort)
+        : undefined,
     };
     if (typeof query === 'string') {
       payload.query = query;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -346,7 +346,7 @@ export class Channel<
         CommandType,
         UserType
       >,
-      ...options_without_sort,
+      ...optionsWithoutSort,
     };
     if (typeof query === 'string') {
       payload.query = query;
@@ -356,8 +356,8 @@ export class Channel<
       throw Error(`Invalid type ${typeof query} for query parameter`);
     }
 
-    if (sort_value) {
-      payload.sort = normalizeQuerySort(sort_value);
+    if (sort) {
+      payload.sort = normalizeQuerySort(sort);
     }
     // Make sure we wait for the connect promise if there is a pending one
     await this.getClient().wsPromise;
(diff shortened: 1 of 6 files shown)


src/channel.ts
```diff
@@ -326,14 +328,25 @@ export class Channel<
...
       >;
-      offset?: number;
       query?: string;
     } = {},
   ) {
     // Return a list of channels
-    const payload = {
-      filter_conditions: { cid: this.cid },
-      ...options,
+    const { sort: sort_value, ...options_without_sort } = { ...options };
```
mahboubii, line 335:
```suggestion
    const { sort, ...optionsWithoutSort } = { ...options };
```

mahboubii (commented):
LGTM 🌮 
@nhannah could you also take a look?

src/types.ts
```diff
@@ -935,11 +944,19 @@ export type QueryMembersOptions = {
...
+export type LimitOffsetSearchOptions = {
   limit?: number;
   offset?: number;
 };
 
+export type NextSearchOptions = {
+  limit?: number;
+  next?: string;
+  sort?: SearchMessageSort;
+};
```
vishalnarkhede, line 958:
```suggestion
export type LimitOffsetSearchOptions = {
  limit?: number;
  offset?: number;
};

export type NextSearchOptions = {
  limit?: number;
  next?: string;
  sort?: SearchMessageSort;
};

export type SearchOptions = LimitOffsetSearchOptions | NextSearchOptions;
```

src/types.ts
```diff
@@ -1407,10 +1424,20 @@ export type UserSort<UserType = UnknownType> =
...
+  | { [field: string]: AscDesc };
+
+export type SearchMessageSort = SearchMessageSortBase | Array<SearchMessageSortBase>;
+
 export type QuerySort<ChannelType = UnknownType, UserType = UnknownType> =
   | BannedUsersSort
   | ChannelSort<ChannelType>
-  | UserSort<UserType>;
+  | UserSort<UserType>
+  | SearchMessageSort;
```
vishalnarkhede, line 1440:
```suggestion
  | SearchMessageSort;
  | UserSort<UserType>
```

vishalnarkhede: Just a small abc ordering!!

src/types.ts
```diff
@@ -1407,10 +1424,20 @@ export type UserSort<UserType = UnknownType> =
   | Sort<UserResponse<UserType>>
   | Array<Sort<UserResponse<UserType>>>;
 
+export type SearchRelevanceSort = { relevance?: AscDesc };
+
+export type SearchMessageSortBase =
+  | SearchRelevanceSort
+  | Sort<Message>
```
vishalnarkhede, line 1431: Shouldn't this generic be an input to `SearchMessageSortBase`?

src/types.ts
vishalnarkhede: Defining before using ... makes it a little easier to read!!

vishalnarkhede: Also, do we really need separate `LimitOffsetSearchOptions` and  `NextSearchOptions`? Because anyways user can't switch between v1 or v2 search, right?

```
export type SearchOptions = {
  limit?: number;
  next?: string;
  sort?: SearchMessageSort;
};
```

vishalnarkhede (changes requested):

src/types.ts
miagilepner: I thought it was better to keep them separate - A user should not be allowed to use offset with next or with sort.

vishalnarkhede: Ah my bad. I thought `sort` was the only difference there :) Thanks for clarification

vishalnarkhede (dismissed):

src/types.ts
```diff
@@ -1407,9 +1424,25 @@ export type UserSort<UserType = UnknownType> =
   | Sort<UserResponse<UserType>>
   | Array<Sort<UserResponse<UserType>>>;
 
-export type QuerySort<ChannelType = UnknownType, UserType = UnknownType> =
+export type SearchRelevanceSort = { relevance?: AscDesc };
+
+export type SearchMessageSortBase<MessageType = UnknownType> =
+  | SearchRelevanceSort
+  | Sort<MessageType>
+  | { [field: string]: AscDesc };
```
nhannah, line 1432: Is there not a tighter set of fields here? `Sort<MessageType>` covers all user created ones but would `Message` or `MessageBase` with `Sort` be usable for the rest instead of [field: string] ?

src/types.ts
```diff
@@ -1911,6 +1944,10 @@ export type SearchPayload<
     UserType
   >;
   query?: string;
+  sort?: Array<{
+    direction: AscDesc;
+    field: string | number;
```
nhannah, line 1949: Can this field not be narrowed either?

src/channel.ts
```diff
@@ -326,14 +328,25 @@ export class Channel<
...
       >;
-      offset?: number;
       query?: string;
     } = {},
   ) {
     // Return a list of channels
-    const payload = {
-      filter_conditions: { cid: this.cid },
-      ...options,
+    const { sort, ...optionsWithoutSort } = { ...options };
```
nhannah, line 335: So sort here is throwing a TS error, this is because only one of the 2 search options types has sort on it.

src/channel.ts
```diff
@@ -343,6 +356,9 @@ export class Channel<
       throw Error(`Invalid type ${typeof query} for query parameter`);
     }
 
+    if (sort) {
+      payload.sort = normalizeQuerySort(sort);
```
nhannah, line 360: There are also some type issues here because of sort being any...

I think bringing these all into one type sorts it all if that's acceptable for the usecase.
```
export type SearchOptions = {
  limit?: number;
  next?: string;
  offset?: number;
  sort?: SearchMessageSort;
};
```

src/channel.ts
nhannah: Don't think you need to spread options here.

src/client.ts
```diff
@@ -1779,7 +1779,7 @@ export class StreamChat<
         >,
     options: SearchOptions = {},
   ) {
-    // Return a list of channels
+    const { sort: sort_value, ...options_without_sort } = { ...options };
```
nhannah, line 1782: Same as in channel but also we can probably avoid the rest spread as well, and pop everything to one line via 

```
      sort: sort &&= normalizeQuerySort(sort_value)
```

nhannah (changes requested):
Hey let's look at this together, if you have to hop in the TS world more I am happy to help, and we can add a test to give some piece of mind on these as well.

src/client.ts
miagilepner: I'm not quite sure what you want to do here. Which lines will be replaced with your suggestion?

src/utils.ts
```diff
@@ -102,7 +102,7 @@ export function normalizeQuerySort<T extends QuerySort>(sort: T) {
   const sortArr = Array.isArray(sort) ? sort : [sort];
   for (const item of sortArr) {
     const entries = (Object.entries(item) as unknown) as [
-      T extends (infer K)[] ? keyof K : keyof T,
+      T extends (infer K)[] ? Extract<keyof K, string> : Extract<keyof T, string>,
```
nhannah, line 105: I think the typing of this function can be improved to 
```
export function normalizeQuerySort<T extends Record<string, AscDesc | undefined>>(
  sort: T | T[],
) {
  const sortFields: Array<{ direction: AscDesc; field: keyof T }> = [];
  const sortArr = Array.isArray(sort) ? sort : [sort];
  for (const item of sortArr) {
    const entries = Object.entries(item) as [keyof T, AscDesc][];
    if (entries.length > 1) {
      console.warn(
        "client._buildSort() - multiple fields in a single sort object detected. Object's field order is not guaranteed",
      );
    }
    for (const [field, direction] of entries) {
      sortFields.push({ field, direction });
    }
  }
  return sortFields;
}```

**Review by nhannah — CHANGES_REQUESTED** [review]
I put in some suggestions; looks like that old util func is messy as well so I put in a suggestion there that I think will correct it's off typing.

File: src/types.ts [hunk 11]
```diff
@@ -1911,6 +1947,10 @@ export type SearchPayload<
     UserType
   >;
   query?: string;
+  sort?: Array<{
+    direction: AscDesc;
+    field: Extract<keyof SearchMessageSortBase<MessageType>, string>;
```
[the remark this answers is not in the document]
miagilepner: `keyof SearchMessageSortBase<MessageType>` is `string | number`

nhannah (approved):
Thanks for the changes, looks good!

vishalnarkhede (approved):

PR #678, opened 2021-05-13
CRNS-352: Unread counts and muted channels

vishalnarkhede:
## CLA

- [ ] Code changes are tested

https://getstream.slack.com/archives/C0185U214TF/p1620718706190100

We didn't consider muted channels while incrementing unread count, when new message comes in. We shouldn't increase unread count if channel is muted.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1335,6 +1335,23 @@ export class StreamChat<
     }
 
     if (event.type === 'notification.channel_mutes_updated' && event.me?.channel_mutes) {
+      const currentMutedChannelIds: string[] = [];
+      const nextMutedChannelIds: string[] = [];
+
+      this.mutedChannels.forEach(
+        (mute) => mute.channel && currentMutedChannelIds.push(mute.channel.cid),
+      );
+      event.me.channel_mutes.forEach(
+        (mute) => mute.channel && nextMutedChannelIds.push(mute.channel.cid),
+      );
+
+      /** Set the unread count of un-muted channels to 0, which is the behaviour of backend */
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -2242,6 +2215,40 @@ export class StreamChat<
     return false;
   }
 
+  /**
+   * channelMuteStatus - check if the channel is muted or not, can be used after connectUser() is called
+   *
+   * @param cid
+   */
+  channelMuteStatus = (cid: string) => this._muteStatus(cid).muted;
```
vishalnarkhede, line 2223: We have a similar function for `userMuteStatus`!! So makes sense to have this function as well

src/client.ts
```diff
@@ -2242,6 +2215,40 @@ export class StreamChat<
...
+      expiresAt: null,
+    };
+  }
+
+  /**
+   * channelMuteStatus - check if the channel is muted or not, can be used after connectUser() is called
+   *
+   * @param cid
+   */
+  channelMuteStatus = (cid: string) => this._muteStatus(cid).muted;
```
vishalnarkhede, line 2250: we have a similar function for `userMuteStatus`, so makes sense to have one for channel as well!!

src/client.ts
```diff
@@ -2242,6 +2232,40 @@ export class StreamChat<
     return false;
   }
 
+  _muteStatus(cid: string) {
```
DanC5, line 2235: since this function only takes one cid as param, shouldn't we break out of the loop if the looped cid matches the param cid?

mahboubii (approved):

PR #683, opened 2021-05-24
make passing revoke date in application level revoke token mandatory

SiddhantAgarwal:
## CLA

- [ ] Code changes are tested

## Changelog

- Make passing revoke date for application level revoke mandatory.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -638,11 +638,7 @@ export class StreamChat<
     return await this.patch<APIResponse>(this.baseURL + '/app', options);
   }
 
-  _normalizeDate = (before: Date | string | null | undefined): string | null => {
-    if (before === undefined) {
-      before = new Date().toISOString();
-    }
-
+  _normalizeDate = (before: Date | string | null): string | null => {
     if (before instanceof Date) {
       before = before.toISOString();
     }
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -660,6 +656,12 @@ export class StreamChat<
    * Revokes all tokens on application level issued before given time
    */
   async revokeTokens(before?: Date | string | null) {
+    if (before === undefined) {
+      throw new Error(
+        'Passing before parameter is mandatory for application level token revocation',
+      );
+    }
```
AnatolyRugalev, line 663: Let's just remove `?` from function arguments

SiddhantAgarwal: we can remove `?` but we use the call as plain javascript call for ex - in chat-qa which causes the before param to be undefined.
I think we should have the check for safety.

AnatolyRugalev: Well, even in JS linter will provide you with error message. And there's nothing bad in sending undefined anyway as the backend will respond with an error. Having `?` in parameters is misleading as one might think that it is possible to call it without an argument

AnatolyRugalev (approved):

ferhatelmas (approved):
LGTM

PR #692, opened 2021-05-31
[CHAT-1851] Add `assignRoles` method to Channel

AnatolyRugalev:
This PR adds support for new `setRoles` operation in `UpdateChannel` API.

Also, did some refactoring around repetitive UpdateChannel SDK calls

UPDATE: renamed `setRoles` to `assignRoles` after discussion with the Product Team
UPDATE 2: switched format of `assignRoles` to object-based

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -646,12 +646,12 @@ export class Channel<
   /**
    * addMembers - add members to the channel
    *
-   * @param {string[]} members An array of member identifiers
+   * @param {{user_id: string, channel_role?: Role}[]} members An array of members to add to the channel
    * @param {Message<AttachmentType, MessageType, UserType>} [message] Optional message object for channel members notification
    * @return {Promise<UpdateChannelAPIResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} The server response
    */
   async addMembers(
-    members: string[],
+    members: string[] | { user_id: string; channel_role?: Role }[],
     message?: Message<AttachmentType, MessageType, UserType>,
   ) {
     return await this._update({
@@ -680,12 +680,12 @@ export class Channel<
   /**
    * assignRoles - sets member roles in a channel
    *
-   * @param {[userID: string]: string} roles An object with role assignments
+   * @param {{channel_role: Role, user_id: string}[]} roles List of role assignments
    * @param {Message<AttachmentType, MessageType, UserType>} [message] Optional message object for channel members notification
    * @return {Promise<UpdateChannelAPIResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} The server response
    */
   async assignRoles(
-    roles: { [userID: string]: Role },
+    roles: { channel_role: Role; user_id: string }[],
     message?: Message<AttachmentType, MessageType, UserType>,
   ) {
     return await this._update({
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -714,21 +670,27 @@ export class Channel<
...
+
+  /**
+   * setRoles - set member roles in a channel
+   *
+   * @param {[userID: string]: string} roles An object with role assignments
+   * @param {Message<AttachmentType, MessageType, UserType>} [message] Optional message object for channel members notification
+   * @return {Promise<UpdateChannelAPIResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} The server response
+   */
+  async setRoles(
+    roles: { [userID: string]: string },
```
AnatolyRugalev, line 687: Let me know if we need type for this

src/channel.ts
```diff
@@ -798,6 +738,18 @@ export class Channel<
...
+      message,
+    });
+  }
+
+  /**
+   * _update - executes channel update request
+   * @param payload Object Update Channel payload
+   * @return {Promise<UpdateChannelAPIResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} The server response
+   */
+  async _update(payload: Object) {
```
AnatolyRugalev, line 752: We don't have type for this. Do we need it?

ferhatelmas (dismissed):
LGTM

Nice refactor. New addition is good, and types should be fine but I delegate types to type experts for final check 🙂

src/types.ts
```diff
@@ -1531,6 +1532,7 @@ export type ChannelData<ChannelType = UnknownType> = ChannelType & {
 
 export type ChannelMembership<UserType = UnknownType> = {
   banned?: boolean;
+  channel_role?: string;
```
mahboubii, line 1535: should we extend this to the default role we have? sth like:
```suggestion
  channel_role?: string | 'admin' | 'moderator' | ......;
```

src/channel.ts
mahboubii: generally yes, since the entire lib is in TS, private functions need to be also typed to make everyone lifes easier

src/types.ts
AnatolyRugalev: We are prepearing infrastructure for new permissions implementation which includes custom roles. These you cannot describe with a enum I'm afraid. We will still have builtin roles though. Can we  express something like this as type?

src/channel.ts
AnatolyRugalev: I tried that and there are some problems that I'm not sure if we want to address now.

1. We already have `InviteOptions` type which should be called `ChannelUpdateOptions` or `UpdateChannelOptions`. We can create an alias to keep BC, but...
2. We also have `UpdateChannelOptions` which should be called `UpdateChannelTypeOptions`

I would suggest we resolve this naming conflict in the next major release and only then introduce strict type here. Introducing new type now will only add confusion

mahboubii: In that case, it's not an urgent fix for now, could you leave a TODO there?
let's pull @nhannah to be in the loop as well

src/types.ts
mahboubii: yea you can have a union of builtin roles with a string type

mahboubii (approved):

ferhatelmas (approved):
LGTM

mahboubii (dismissed):

ferhatelmas (approved):
LGTM

PR #699, opened 2021-06-04
CHAT-1828: clean pinned messages after channel truncation

doesnotcommit:
This is the client-side fix to retained pinned messages after a channel is truncated


AnatolyRugalev (approved):
Looks good :)

mahboubii (approved):

PR #705, opened 2021-06-23
Fix #701: AppSettings type multi_tenant_enabled

### LINKED ISSUE #701: `multi_tenant_enabled`  attribute missing from AppSettings type definition
Quite a small detail, but the [AppSettings type](https://github.com/GetStream/stream-chat-js/blob/master/src/types.ts#L1434) does not declare the boolean attribute `multi_tenant_enabled` which is used to enable/disable teams. This causes an error when using Typescript.

Obviously the workaround is to simply cast `(options as any) as AppSettings` so not a big problem!

gz#12814

mahboubii:
## CLA

- [ ] Code changes are tested

### Fix
- AppSettings type support `multi_tenant_enabled` attribute

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1462,6 +1462,7 @@ export type AppSettings = {
   };
   image_moderation_enabled?: boolean;
   image_upload_config?: FileUploadConfig;
+  multi_tenant_enabled?: boolean;
   push_config?: {
     version?: string;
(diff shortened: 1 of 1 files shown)


ferhatelmas (approved):
LGTM 

but there are still missing ones.

PR #707, opened 2021-06-24
[CHAT-1809] Updated Permissions API

AnatolyRugalev:
This PR update permissions API to sync SDK with backend changes


mahboubii:
@AnatolyRugalev could you add a changelog and how to update their current code for the customers upgrading?

AnatolyRugalev:
@mahboubii we don't have any customers using these endpoints

nhannah:
@AnatolyRugalev So no one was using V1 permissions? Tests are broken now.,

PR #713, opened 2021-07-08
Fix/thread loading bug and message update bug

nhannah:
Because the SDKs rely on the current message state to determine when and what they should reload there were multiple edge cases currently present that could cause issues with loading. Updates, and deletions to both messages and thread messages could add to the message list messages which were out of sync with the current desired state. These caused the loading logic to function incorrectly as it is dependent on the currently loaded state for pagination.

This situation also arose with messages in that are shown both in a thread and in the main channel. They would be loaded for the main channel but also added to the thread, these randomly added thread messages could cause the loading logic for threads to also get out of sync.

These fixes address these edge cases to keep a state more inline with that desired by the SDKs and users.

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -143,6 +143,7 @@ export class ChannelState<
    *
    * @param {MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>} newMessage A new message
    * @param {boolean} timestampChanged Whether updating a message with changed created_at value.
+   * @param {boolean} addIfDoesNotExist Add message if it is not in the list, used to prevent out of order updated messages from being added.
    *
    */
   addMessageSorted(
@@ -209,6 +210,7 @@ export class ChannelState<
    * @param {Array<MessageResponse<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>>} newMessages A list of messages
    * @param {boolean} timestampChanged Whether updating messages with changed created_at value.
    * @param {boolean} initializing Whether channel is being initialized.
+   * @param {boolean} addIfDoesNotExist Add message if it is not in the list, used to prevent out of order updated messages from being added.
    *
    */
   addMessagesSorted(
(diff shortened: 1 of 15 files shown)


AnatolyRugalev:
@nhannah we have upcoming changes regarding this exact problem from the backend, so let's not rush merging this as this may fix issue without SDK changes

src/channel_state.ts
```diff
@@ -252,13 +259,33 @@ export class ChannelState<
...
-      // add to the thread if applicable..
-      if (parentID) {
+      /**
+       * Add message to thread if applicable and the message
+       * was added when querying for replies, or the thread already exits.
+       * This is to prevent the thread state from getting out of sync if
+       * a thread message is shown in channel but older than the newest thread
+       * message. This situation can result in a thread state where a random
+       * message is "oldest" message, and newer messages are therefore not loaded.
+       * This can also occur is an old thread message is updated.
```
vishalnarkhede, line 278:
```suggestion
       * This can also occur if an old thread message is updated.
```

src/channel_state.ts
```diff
@@ -526,14 +555,26 @@ export class ChannelState<
...
+      return [...messageArr];
+    }
 
     const messageTime = (message[sortBy] as Date).getTime();
 
-    // if message is newer than last item in the list concat and return
-    if ((messageArr[messageArr.length - 1][sortBy] as Date).getTime() < messageTime)
+    // if message is newer than last item in the list concat and return unless it's an update or deletion
+    if (
+      (messageArr[messageArr.length - 1][sortBy] as Date).getTime() < messageTime &&
```
vishalnarkhede, line 569: maybe a separate const for `(messageArr[messageArr.length - 1][sortBy] as Date).getTime()` as well?

src/channel_state.ts
```diff
@@ -517,7 +544,9 @@ export class ChannelState<
     >,
     timestampChanged = false,
     sortBy: 'pinned_at' | 'created_at' = 'created_at',
+    addIfDoesNotExist = true,
```
vishalnarkhede, line 547: needs to go in jsdoc annotation as well!!

vishalnarkhede (dismissed):

vishalnarkhede (approved):

PR #724, opened 2021-07-26
Campaign crud 

ferhatelmas:

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2009,3 +2009,62 @@ export type User<UserType = UnknownType> = UserType & {
   teams?: string[];
   username?: string;
 };
+
+export type SegmentData = {
+  description: string;
+  filter: {
+    channel?: object;
+    user?: object;
+  };
+  name: string;
+};
+
+export type Segment = {
+  app_pk: number;
+  created_at: string;
+  id: string;
+  updated_at: string;
+  recipients?: number;
+} & SegmentData;
+
+export type CampaignData = {
+  attachments: Attachment[];
+  defaults: { [key: string]: string };
+  segment_id: string;
+  sender_id: string | null;
+  text: string;
+  push_notifications?: boolean;
+  track_opened?: boolean;
+};
+
+export type CampaignStatus = {
+  errors: string[];
+  status:
+    | 'draft'
+    | 'stopped'
+    | 'scheduled'
+    | 'completed'
+    | 'failed'
+    | 'canceled'
+    | 'in_progress';
+  completed_at?: string;
+  failed_at?: string;
+  progress?: {
+    delivered: number;
+    errored: number;
+    sent: number;
+  };
+  resumed_at?: string;
+  scheduled_at?: string;
+  stopped_at?: string;
+};
+
+export type Campaign = {
+  app_pk: string;
+  created_at: string;
+  id: string;
+  updated_at: string;
+  test?: boolean;
+  test_original_id?: string;
+} & CampaignData &
+  CampaignStatus;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -2063,3 +2063,62 @@ export type User<UserType = UnknownType> = UserType & {
   teams?: string[];
   username?: string;
 };
+
+export type SegmentData = {
+  description: string;
+  filter: {
+    channel?: object;
+    user?: object;
```
mahboubii, line 2071: are these enriched objects from API?

src/types.ts
```diff
@@ -2063,3 +2063,62 @@ export type User<UserType = UnknownType> = UserType & {
...
+  id: string;
+  updated_at: string;
+  recipients?: number;
+} & SegmentData;
+
+export type CampaignData = {
+  attachments: Attachment[];
+  defaults: { [key: string]: string };
+  segment_id: string;
+  sender_id: string | null;
```
mahboubii, line 2088: should this be optional to be omitted from the request?

src/types.ts
```diff
@@ -2063,3 +2063,62 @@ export type User<UserType = UnknownType> = UserType & {
...
+  app_pk: number;
+  created_at: string;
+  id: string;
+  updated_at: string;
+  recipients?: number;
+} & SegmentData;
+
+export type CampaignData = {
+  attachments: Attachment[];
+  defaults: { [key: string]: string };
```
mahboubii, line 2086:
```suggestion
  defaults: Record<string, string>;
```

mahboubii (commented):
Should we also add some jsdoc for these new endpoints?

src/types.ts
gumuz: no, they're either a User or Channel filter (like queryusers/querychannels), example:

`{"id": 1223, "username": {$eq: "amin"}}`

src/types.ts
gumuz: Yeah, so I guess it should be `string?` instead. I sthat what you're suggesting?

src/types.ts
mahboubii: in that case let's pull @nhannah to help us with this type

src/types.ts
mahboubii: exactly

mahboubii (approved):
🌮

gumuz (approved):

src/types.ts
```diff
@@ -2063,3 +2063,63 @@ export type User<UserType = UnknownType> = UserType & {
...
+} & SegmentData;
+
+export type CampaignData = {
+  attachments: Attachment[];
+  defaults: Record<string, string>;
+  segment_id: string;
+  text: string;
+  push_notifications?: boolean;
+  sender_id?: string;
+  track_opened?: boolean;
```
ferhatelmas, line 2092: we don't have this

src/types.ts
```diff
@@ -2063,3 +2063,63 @@ export type User<UserType = UnknownType> = UserType & {
...
+    | 'draft'
+    | 'stopped'
+    | 'scheduled'
+    | 'completed'
+    | 'failed'
+    | 'canceled'
+    | 'in_progress';
+  completed_at?: string;
+  failed_at?: string;
+  progress?: {
```
ferhatelmas, line 2107: this isn't aligned with api either, we might want to have this way, more user friendly

src/types.ts
```diff
@@ -2063,3 +2063,63 @@ export type User<UserType = UnknownType> = UserType & {
...
+  stopped_at?: string;
+};
+
+export type Campaign = {
+  app_pk: string;
+  created_at: string;
+  id: string;
+  updated_at: string;
+  test?: boolean;
+  test_original_id?: string;
```
ferhatelmas, line 2123: these flags aren't there either

PR #726, opened 2021-07-28
Add remove quoted messages logics

tsirlucas:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -653,7 +653,6 @@ export class ChannelState<
     if (addMessageToList) {
       messageArr.splice(left, 0, message);
     }
-    console.log('finished state update');
     return [...messageArr];
   }
 
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -437,6 +437,51 @@ export class ChannelState<
     return messageWithReaction;
   }
 
+  removeQuotedMessages(
+    message: MessageResponse<
+      AttachmentType,
+      ChannelType,
+      CommandType,
+      MessageType,
+      ReactionType,
+      UserType
+    >,
+  ) {
+    const parseMessage = (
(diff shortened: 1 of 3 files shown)


src/channel_state.ts
```diff
@@ -437,6 +437,51 @@ export class ChannelState<
...
+    message: MessageResponse<
+      AttachmentType,
+      ChannelType,
+      CommandType,
+      MessageType,
+      ReactionType,
+      UserType
+    >,
+  ) {
+    const parseMessage = (
```
tsirlucas, line 450: @vishalnarkhede second time this function is needed (the other one was in the sdk) maybe we can put it on utils or something?

src/channel_state.ts
```diff
@@ -437,6 +437,51 @@ export class ChannelState<
     return messageWithReaction;
   }
 
+  removeQuotedMessages(
```
vishalnarkhede, line 440: I think should be more like `updateQuotedMessages`, and it should also be called on `message.updated` event. Right?

vishalnarkhede: Or maybe we need another function called `updateQuotedMessages`, since there some functionality specific to removing/cleaning quoted message in this function.

vishalnarkhede: Also can you add unit test for this function?

tsirlucas: yeah tbh this function could be used for updating as well but the main problem is that it wasnt erasing the attachments so we had to add that empty array on attachments.

lets add the update in another pr once i had some trouble with this remove logics and dont really want to have the same kind of problems in the update before moving the sprint card lol

vishalnarkhede (approved):

vishalnarkhede (approved):

PR #727, opened 2021-07-28
CRNS-317: handling deleted properties of user

vishalnarkhede:
## CLA

- [ ] Code changes are tested

When user gets updated on backend, the event handler for event `user.updated` doesn't take care of properties which were removed. Purpose of this PR is to remove such deleted properties from `client.user` and `client._user` object.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -13,6 +13,7 @@ import { JWTUserToken, DevToken, CheckSignature } from './signing';
 import { TokenManager } from './token_manager';
 import {
   isFunction,
+  isOwnUserProperty,
   addFileToFormData,
   chatCodes,
   normalizeQuerySort,
@@ -1346,20 +1347,9 @@ export class StreamChat<
       (event.type === 'user.presence.changed' || event.type === 'user.updated') &&
       event.user.id === this.userID
     ) {
-      const ownUserProperties = [
-        'channel_mutes',
-        'devices',
-        'mutes',
-        'total_unread_count',
-        'unread_channels',
-        'unread_count',
-        'invisible',
-        'roles',
-      ];
-
       // Remove deleted properties from user objects.
       for (const key in this.user) {
-        if (key in event.user || ownUserProperties.includes(key)) {
+        if (key in event.user || isOwnUserProperty(key)) {
           continue;
         }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -13,7 +13,7 @@ import { JWTUserToken, DevToken, CheckSignature } from './signing';
 import { TokenManager } from './token_manager';
 import {
   isFunction,
-  isOwnUserProperty,
+  isOwnUserBaseProperty,
   addFileToFormData,
   chatCodes,
   normalizeQuerySort,
@@ -1349,7 +1349,7 @@ export class StreamChat<
     ) {
       // Remove deleted properties from user objects.
       for (const key in this.user) {
-        if (key in event.user || isOwnUserProperty(key)) {
+        if (key in event.user || isOwnUserBaseProperty(key)) {
           continue;
         }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -103,6 +103,10 @@ import {
   UserResponse,
   UserSort,
   SearchMessageSortBase,
+  SegmentData,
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -1339,17 +1342,46 @@ export class StreamChat<
...
+        'total_unread_count',
+        'unread_channels',
+        'unread_count',
+        'invisible',
+        'roles',
+      ];
+
+      // Remove deleted properties from user objects.
+      for (const key in this.user) {
+        if (key in event.user || ownUserProperties.includes(key)) {
```
mahboubii, line 1362: why do we care about the ownUserProperties here?

vishalnarkhede: We don't want to remove those ownUserProperties ... they are kinda reserved fields!!

mahboubii: oh i see and they are not in the event object, in that case maybe we should extract these fields and put them somewhere as a source of truth, in future, if we add more reserve fields and forgot to add them here it will cause hard to find issues

src/utils.ts
```diff
@@ -74,6 +75,40 @@ export function isOwnUser<
...
+/**
+ * @todo Fix the typescript to ensure all the properties of `OwnUserBase` exist in return statement.
+ *
+ * @returns {array} "Own user" specific properties
+ */
+function getOwnUserBaseProperties<
+  ChannelType extends UnknownType = UnknownType,
+  CommandType extends string = LiteralStringForUnion,
+  UserType extends UnknownType = UnknownType
+>(): (keyof OwnUserBase<ChannelType, CommandType, UserType>)[] {
```
mahboubii, line 87: let's drop all the generics here since they dont have any value

src/utils.ts
```diff
@@ -74,6 +75,40 @@ export function isOwnUser<
...
+    'unread_count',
+    'invisible',
+    'roles',
+  ];
+}
+
+export function isOwnUserProperty<
+  ChannelType extends UnknownType = UnknownType,
+  CommandType extends string = LiteralStringForUnion,
+  UserType extends UnknownType = UnknownType
```
mahboubii, line 103: same drop here

src/utils.ts
```diff
@@ -74,6 +75,40 @@ export function isOwnUser<
...
+>(property: string) {
+  return (getOwnUserBaseProperties<
+    ChannelType,
+    CommandType,
+    UserType
+  >() as string[]).includes(property);
```
mahboubii, line 109: is ts forcing you to cast here?

mahboubii (dismissed):

src/client.ts
```diff
@@ -1339,17 +1343,35 @@ export class StreamChat<
...
+      event.user.id === this.userID
+    ) {
+      // Remove deleted properties from user objects.
+      for (const key in this.user) {
+        if (key in event.user || isOwnUserBaseProperty(key)) {
+          continue;
+        }
+
+        if (this.user?.[key]) {
+          delete this.user[key];
```
tsirlucas, line 1357: hey maybe create a shallow copy of user and delete properties on it instead of deleting this.user[key] directly? this way we can only set this.user in the end of the function so we will only mutate it once instead of manually deleting the keys during the function execution.
this is useful to avoid half-updated states in case this function fails in the middle of its execution.

src/client.ts
```diff
@@ -1339,17 +1343,35 @@ export class StreamChat<
...
+        }
+
+        if (this._user?.[key]) {
```
tsirlucas, line 1360: whats the difference between user and _user here? just a question, not really a review feedback

tsirlucas (changes requested):
requested one change and had a question.

also, this is taking care of the problem where the event doesnt return deleted properties, right? asking cause remember hearing you guys talking about a similar problem the other day

src/client.ts
vishalnarkhede: makes sense :)

src/client.ts
vishalnarkhede: Screenshot 2021-07-28 at 18 14 34

This comment might answer this one!!

tsirlucas (approved):

PR #732, opened 2021-07-29
Fixing delete channel logic

vishalnarkhede:
## CLA

- [ ] Code changes are tested

## Changelog

- Once client receive `channel.deleted` or `notification.channel_deleted`, we remove it from `client.activeChannels`. But this prevents channel listeners to be called for that event. So we should only remove it from activeChannels after listeners have been fired.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1216,7 +1216,7 @@ export class StreamChat<
     >,
   ) => {
     // client event handlers
-    this._handleClientEvent(event);
+    const postListenerCallbacks = this._handleClientEvent(event);
 
     // channel event handlers
     const cid = event.cid;
@@ -1230,6 +1230,8 @@ export class StreamChat<
     if (channel) {
       channel._callChannelListeners(event);
     }
+
+    postListenerCallbacks.forEach((c) => c());
   };
 
   handleEvent = (messageEvent: WebSocket.MessageEvent) => {
@@ -1400,6 +1402,7 @@ export class StreamChat<
     >,
   ) {
     const client = this;
+    const postListenerCallbacks = [];
     this.logger(
       'info',
       `client:_handleClientEvent - Received event of type { ${event.type} }`,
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -1230,6 +1230,8 @@ export class StreamChat<
     if (channel) {
       channel._callChannelListeners(event);
     }
+
+    postListenerCallbacks.forEach((c) => c());
```
mahboubii, line 1234:
```suggestion
    postListenerCallbacks.forEach(c);
```

src/client.ts
```diff
@@ -2063,6 +2073,10 @@ export class StreamChat<
     //                        we will replace it with `cid`
     for (const key in this.activeChannels) {
       const channel = this.activeChannels[key];
+      if (channel.disconnected) {
+        continue;
+      }
```
mahboubii, line 2078: should we delete the channel in this case or let it be replaced later?

src/client.ts
```diff
@@ -1458,10 +1461,17 @@ export class StreamChat<
...
-      this.activeChannels[event.cid]?._disconnect();
-      delete this.activeChannels[event.cid];
       client.state.deleteAllChannelReference(event.cid);
+      this.activeChannels[event.cid]?._disconnect();
+
+      postListenerCallbacks.push(() => {
+        if (!event.cid) return;
+
+        delete this.activeChannels[event.cid];
+      });
```
mahboubii, line 1471:
```suggestion
      if (event.cid) 
            postListenerCallbacks.push(() => {
              delete this.activeChannels[event.cid];
            });
```

mahboubii (approved):

src/client.ts
vishalnarkhede: As discussed, resolved

src/client.ts
vishalnarkhede: We already have the check, this one is just to satisfy typescript!!

PR #742, opened 2021-08-06
fix: fixing connectUser error handling

vishalnarkhede:
## CLA

- [ ] Code changes are tested

## Changelog

- With our existing logic, in case of failure in call to `/connect` api (case of request never reached our server for example), we never threw an error. So `await client.connectUser()` would falsely get resolved. If also affected calls to `client.queryChannels`, `client.queryUsers` where we rely on result of promise returned by `connectUser` function. This issue has been fixed by throwing appropriate error.
- Disallow reconnect requests if first call to `connectUser` was never succesful. Basically when `client.connectUser` fails, the failures should be handled on app level instead of js client. This brings consistency with other LLCs (flutter, android) of Stream. Also if the call to connectUser was failed due to invalid config (apiKey, token etc), there is no point in making reconnect calls.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -346,6 +346,17 @@ export class StableWSConnection<
     this.logger('info', 'connection:_reconnect() - Initiating the reconnect', {
       tags: ['connection'],
     });
+    if (!this.isFirstConnectSuccesful) {
+      this.logger(
+        'info',
+        'connection:_reconnect() - Abort (0) since first call to "client.connectUser" never succeded',
+        {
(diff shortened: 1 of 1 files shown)


mahboubii:
what part of the changes address this?
>With our existing logic, in case of failure in call to /connect api (case of request never reached our server for example), we never threw an error. So await client.connectUser() would falsely get resolved. If also affected calls to client.queryChannels, client.queryUsers where we rely on result of promise returned by connectUser function. This issue has been fixed by throwing appropriate error.

PR #748, opened 2021-08-11
Offline support preparation

tsirlucas:
This PR adds some necessary functions in order to allow the RN SDK to have offline support.


tsirlucas:
@nhannah do you think it makes sense to move the types to another file instead of keeping then on top of the current ones?

src/client.ts
```diff
@@ -506,6 +524,14 @@ export class StreamChat<
       console.warn(
         'Consecutive calls to connectUser is detected, ideally you should only call this function once in your app.',
       );
+
+      // If last connection attempt failed, we reset user connection with new data. Otherwise we just return the old connection
```
tsirlucas, line 528: @mahboubii I added this change thinking about keeping the behavior you would want and adapting it a little bit in order to keep the support for offline mode.

Users would still need to await this call and it would still fail in case the connection fail. So they can try to reconnect and things should work fine. The difference is that, on offline mode, I can ignore that error and would still have the user, etc in the client state.

mahboubii: still without the disconnect users can't really retry the connectUser()

src/client.ts
```diff
@@ -472,6 +485,24 @@ export class StreamChat<
 
   _hasConnectionID = () => Boolean(this.wsConnection?.connectionID);
 
+  _setUserConnection = (
```
vishalnarkhede, line 488: Is this used anywhere?

src/types.ts
```diff
@@ -1475,6 +1475,179 @@ export type QuerySort<
...
+ */
+
+export type ClientStateData<UserType = UnknownType> = {
+  userChannelReferences: { [key: string]: { [key: string]: boolean } };
+  users: {
+    [key: string]: UserResponse<UserType>;
+  };
+};
+
+export type ClientStateAndData<
```
DanC5, line 1489: this type name is slightly confusing when looked at closely with the above, maybe `ClientStateAndUserData` would make more sense?

DanC5 (approved):

PR #755, opened 2021-08-19
[CHAT-2048] deactivate & delete User helper method

gumuz:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2280,6 +2280,20 @@ export class StreamChat<
     >(this.baseURL + `/users/${userID}`, params);
   }
 
+  async deleteUsers(
+    userIDs: string[],
+    params?: {
+      delete_conversation_channels?: boolean;
+      hard_delete?: boolean;
+      mark_messages_deleted?: boolean;
+    },
+  ) {
+    return await this.post<APIResponse>(this.baseURL + `/users-delete`, {
+      user_ids: userIDs,
+      ...params,
+    });
+  }
+
   async reactivateUser(
     userID: string,
     options?: { created_by_id?: string; name?: string; restore_messages?: boolean },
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -2305,6 +2305,28 @@ export class StreamChat<
...
 
+  /** deactivateAndDeleteUser - convenience method that both deactivates and deletes a user
+   *
+   * @param {string} userID target user ID
+   * @param {object} [options] options for both deactivateUser and deleteUser
+   * @returns {Promise<APIResponse>}
+   */
+  async deactivateAndDeleteUser(
+    userID: string,
+    options?: {
```
ferhatelmas, line 2316: It is better to extract these options to types and combine here. Otherwise easy to go outdated.

src/client.ts
```diff
@@ -2305,6 +2305,28 @@ export class StreamChat<
...
+      delete_conversation_channels?: boolean;
+      hard_delete?: boolean;
+      mark_messages_deleted?: boolean;
+    },
+  ) {
+    await this.deactivateUser(userID, {
+      created_by_id: options?.created_by_id,
+      mark_messages_deleted: options?.mark_messages_deleted,
+    });
+    return await this.deleteUser(userID, options);
```
AnatolyRugalev, line 2327: What happens when we do `mark_messages_deleted` twice?

ferhatelmas: Should be noop but good use-case test for QA.

AnatolyRugalev (approved):

src/client.ts
```diff
@@ -2305,6 +2305,28 @@ export class StreamChat<
...
+  async deactivateAndDeleteUser(
+    userID: string,
+    options?: {
+      created_by_id?: string;
+      delete_conversation_channels?: boolean;
+      hard_delete?: boolean;
+      mark_messages_deleted?: boolean;
+    },
+  ) {
+    await this.deactivateUser(userID, {
```
ferhatelmas, line 2323: Also, should we fail here if user is already deactivated? I think not.

mahboubii: yea it's a noop

PR #762, opened 2021-08-25
[CHAT-2191] async batch delete users

gumuz:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3276,4 +3276,37 @@ export class StreamChat<
     );
     return campaign;
   }
+
+  /**
+   * deleteUsers - Batch delete Users
+   *
+   * @param {string[]} userIDs User IDs
+   * @param {{delete_conversation_channels?: boolean, hard_delete?: boolean, mark_messages_deleted?: boolean}} params User delete parameters
+   *
+   * @return {APIResponse} A task ID
+   */
+  async deleteUsers(
+    userIDs: string[],
+    params?: {
+      delete_conversation_channels?: boolean;
+      hard_delete?: boolean;
+      mark_messages_deleted?: boolean;
+    },
+  ) {
+    return await this.post<APIResponse>(this.baseURL + `/users-delete`, {
+      user_ids: userIDs,
+      ...params,
+    });
+  }
+
+  /**
+   * getTask - Gets a task
+   *
+   * @param {string} id Task ID
+   *
+   * @return {APIResponse} The task status
+   */
+  async getTask(id: string) {
+    return this.get<APIResponse>(`${this.baseURL}/tasks/${id}`);
+  }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -3293,7 +3293,7 @@ export class StreamChat<
       mark_messages_deleted?: boolean;
     },
   ) {
-    return await this.post<APIResponse>(this.baseURL + `/users-delete`, {
+    return await this.post<APIResponse>(this.baseURL + `/async/delete/users`, {
       user_ids: userIDs,
       ...params,
     });


src/client.ts
```diff
@@ -3276,4 +3276,37 @@ export class StreamChat<
...
+    },
+  ) {
+    return await this.post<APIResponse>(this.baseURL + `/async/delete/users`, {
+      user_ids: userIDs,
+      ...params,
+    });
+  }
+
+  /**
+   * getTask - Gets a task
```
yaziine, line 3303:
```suggestion
   * getTask - Gets the status of a task
```

src/client.ts
```diff
@@ -3305,4 +3307,17 @@ export class StreamChat<
...
+   * deleteUsers - Batch Delete Users
+   *
+   * @param {DeleteUserOptions} options Configuration to delete users; which users and how
+   *
+   * @return {APIResponse} A task ID
+   */
+  async deleteUsers(options: DeleteUserOptions) {
+    return await this.post<APIResponse & TaskResponse>(this.baseURL + `/users/delete`, {
+      ...options,
+    });
```
mahboubii, line 3321:
```suggestion
    return await this.post<APIResponse & TaskResponse>(this.baseURL + `/users/delete`, options);
```

src/types.ts
```diff
@@ -2122,6 +2122,19 @@ export type User<UserType = UnknownType> = UserType & {
   username?: string;
 };
 
+export type DeleteType = 'soft' | 'hard';
+
+export type DeleteUserOptions = {
+  user_ids: string[];
+  conversations?: DeleteType;
+  messages?: DeleteType;
+  user?: DeleteType;
```
mahboubii, line 2131: do we skip deleting these if no option is passed?

gumuz: actually, I think `user` should not be optional, it will be set to default on the backend anyway

gumuz:
```suggestion
  user?: DeleteType;
  conversations?: DeleteType;
  messages?: DeleteType;
```

mahboubii, line 2131: you mean it should be required?
```suggestion
  user_ids: string[];
  user: DeleteType;
  conversations?: DeleteType;
  messages?: DeleteType;
```

gumuz: isn't that what I did?

mahboubii: it's a bit confusing how the behavior is on this one, if no option is passed is it soft delete or no delete for `messages` and `conversations`? also what's the difference between these two?

mahboubii: > isn't that what I did?

it was still showing it as optional!

gumuz: Seems confusing to me to make "deleting a user" optional, on a Deleteusers endpoint. However, messages and conversations are optional.

mahboubii: what exactly is `conversations`?

gumuz: 1on1 channels

gumuz: any with 2 members, so after deleting the user we'd be left with a channel with only 1 member, in which case we delete it (with this option)

mahboubii: makes sense, should we add some docs to clarify this in the type?

src/client.ts
```diff
@@ -3305,4 +3307,15 @@ export class StreamChat<
...
+
+  /**
+   * deleteUsers - Batch Delete Users
+   *
+   * @param {DeleteUserOptions} options Configuration to delete users; which users and how
+   *
+   * @return {APIResponse} A task ID
+   */
+  async deleteUsers(options: DeleteUserOptions) {
+    return await this.post<APIResponse & TaskResponse>(this.baseURL + `/users/delete`, options);
```
mahboubii, line 3319: lint
```suggestion
    return await this.post<APIResponse & TaskResponse>(
      this.baseURL + `/users/delete`,
      options,
    );```

PR #764, opened 2021-09-03
Clear text in export

ferhatelmas:
optional flags for export channels: this one requests delete message text to be overwritten (by default).


mahboubii (approved):
🌮

PR #770, opened 2021-09-17
[CHAT-2294] GetTask endpoint to retrieve async task results

gumuz:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3279,4 +3279,15 @@ export class StreamChat<
     );
     return campaign;
   }
+
+  /**
+   * getTask - Gets status of a long running task
+   *
+   * @param {string} id Task ID
+   *
+   * @return {APIResponse} The task status
+   */
+  async getTask(id: string) {
+    return this.get<APIResponse>(`${this.baseURL}/tasks/${id}`);
+  }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -3289,6 +3289,6 @@ export class StreamChat<
    * @return {TaskStatus} The task status
    */
   async getTask(id: string) {
-    return this.get<TaskStatus>(`${this.baseURL}/tasks/${id}`);
+    return this.get<APIResponse & TaskStatus>(`${this.baseURL}/tasks/${id}`);
   }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -111,6 +111,7 @@ import {
   Segment,
   Campaign,
   CampaignData,
+  TaskStatus,
 } from './types';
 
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -3279,4 +3280,15 @@ export class StreamChat<
...
+
+  /**
+   * getTask - Gets status of a long running task
+   *
+   * @param {string} id Task ID
+   *
+   * @return {TaskStatus} The task status
+   */
+  async getTask(id: string) {
+    return this.get<TaskStatus>(`${this.baseURL}/tasks/${id}`);
```
ferhatelmas, line 3292: add `APIResponse` too

gumuz:
```suggestion
    return this.get<APIResponse & TaskStatus>(`${this.baseURL}/tasks/${id}`);
```

src/types.ts
```diff
@@ -2146,3 +2146,11 @@ export type Campaign = {
   updated_at: string;
 } & CampaignData &
   CampaignStatus;
+
+export type TaskStatus = {
+  created_at: string;
+  status: string;
+  task_id: string;
+  updated_at: string;
+  result?: object;
```
ferhatelmas, line 2155: `error` is missing

src/types.ts
```diff
@@ -2146,3 +2146,15 @@ export type Campaign = {
...
   CampaignStatus;
+
+export type TaskStatus = {
+  created_at: string;
+  status: string;
+  task_id: string;
+  updated_at: string;
+  error?: {
+    description: string;
+    type: string;
```
ferhatelmas, line 2157: stacktrace and version?

gumuz: there are omitted to to prevent leaking internal info

ferhatelmas: good

ferhatelmas (dismissed):
LGTM

src/types.ts
```diff
@@ -2146,3 +2146,15 @@ export type Campaign = {
...
+  };
+  result?: object;
```
mahboubii, line 2159: could this get expanded a bit more?

gumuz: so this result in of an undefined type/shape, what do you suggest the typedef should be in that case?

mahboubii: it's either `undefined` or an unknown object? in this case:
```suggestion
  result?: UnknownType;
```

mahboubii (approved):

ferhatelmas (approved):

PR #782, opened 2021-09-29
Fix browser bundle

ferhatelmas:


ferhatelmas:
we're reverting the change in #768 via #783 so this is not needed.

PR #789, opened 2021-10-14
[CHAT-2417] Add Sync option to include inaccessible_cids

ffenix113:
Add new /sync option to add `inaccessible_cids`, which will include CIDs user don't have access to instead of having simulated `NotificationRemovedFromChannel` events.

Test are currently not provided because it requires new sync endpoint enabled for test client. Will be added when endpoint will be updated.


AnatolyRugalev:
Let's hold on for a bit, we are not 100% sure if we want to merge this today

PR #792, opened 2021-10-15
[CHAT-1467] Expose webhook_events field

nmerkulov:
We're extending `updateApp` API  with new property, called "webhook_events". It is used to filter events sent via webhooks/sqs. 
* Sending `null` will reset this filter to default state;
* If no filters were set `GetApp` response will contain default event types set


mahboubii:
> It would also be nice to squash the commits here if possible, even though I see we haven't been super strict with the commit history so far.

@vicnicius I've limited the merge options, would be good to set up a workflow for changelog as well

PR #804, opened 2021-11-03
[CHAT-2520] Remove flag command

yaziine:


ferhatelmas:
From Other SDKs, only .NET had it, dropped it too. https://github.com/GetStream/stream-chat-net/pull/61

PR #814, opened 2021-11-22
LongPoll

mahboubii:
- [x] Retry
- [x] Recover state
- [ ] Test


src/connection_fallback.ts
```diff
@@ -0,0 +1,120 @@
+import axios, { AxiosRequestConfig, CancelTokenSource } from 'axios';
+import { chatCodes, retryInterval, sleep } from './utils';
+import { StreamChat } from './client';
+import { ConnectionOpen, Event, UnknownType, UR } from './types';
+
+enum ConnectionState {
+  Closed = 'CLOSED',
+  Connected = 'CONNECTED',
+  Connecting = 'CONNECTING',
+  Disconnectted = 'DISCONNECTTED',
```
thesyncim, line 10: Disconnected

src/client.ts
```diff
@@ -736,7 +751,7 @@ export class StreamChat<
     // reset client state
     this.state = new ClientState();
     // reset token manager
-    this.tokenManager.reset();
+    setTimeout(this.tokenManager.reset); // delay reseting to use token for disconnect calls
```
vishalnarkhede, line 754: whats the reason for this `setTimeout`? Just curious what was the usecase for it

src/connection.ts
```diff
@@ -256,14 +266,14 @@ export class StableWSConnection<
    * @return {ConnectAPIResponse<ChannelType, CommandType, UserType>} Promise that completes once the first health check message is received
    */
   async _connect() {
-    if (this.isConnecting) return; // simply ignore _connect if it's currently trying to connect
+    if (this.isConnecting || this.isDisconnected) return; // simply ignore _connect if it's currently trying to connect
```
vishalnarkhede, line 269: This will affect background/foreground logic on mobile, no? Once disconnected, user  won't be able to connect again when app comes back to foreground

src/client.ts
```diff
@@ -1436,7 +1449,33 @@ export class StreamChat<
...
-    return await this.wsConnection.connect();
+    try {
+      // if WSFallback is enabled, ws connect should timeout faster so fallback can try
+      return await this.wsConnection.connect(
+        this.options.enableWSFallback ? this.defaultWSTimeoutWithFallback : this.defaultWSTimeout,
+      );
+    } catch (err) {
+      // run fallback only if it's WS/Network error and not a normal API error
+      // make sure browser is online before even trying the longpoll
+      if (this.options.enableWSFallback && isWSFailure(err) && isOnline()) {
```
ferhatelmas, line 1460: might want to revert logic to early return

src/client.ts
```diff
@@ -2571,6 +2610,22 @@ export class StreamChat<
...
+   * encode ws url payload
+   * @private
+   * @returns json string
+   */
+  _buildWSPayload = (client_request_id?: string) => {
+    return JSON.stringify({
+      user_id: this.userID,
+      user_details: this._user,
+      user_token: this.tokenManager.getToken(),
+      server_determines_connection_id: true,
```
ferhatelmas, line 2623: this isn't used anymore

src/client.ts
mahboubii: basically it's there to change the order of execution without breaking change which is after closeConnection sync part is done(reading the token one last time) and this function returns

src/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+type APIError = Error & { code?: number; isWSFailure?: boolean };
+
+export function isAPIError(error: Error): error is APIError {
+  return (error as APIError).code !== undefined;
+}
+
+export function isErrorRetryable(error: APIError) {
+  if (!error.code) return false;
+  const err = APIErrorCodes[`${error.code}`];
+  if (!err) return false;
```
ferhatelmas, line 40: 👍🏻

src/connection.ts
ferhatelmas: needs maintenance of it in callback to me

mahboubii: what do you call to connect again when the app comes to foreground?

src/client.ts
mahboubii: what about user_token as part of the json payload?

ferhatelmas: same, not needed

ferhatelmas: authorization header is passed

src/connection.ts
mahboubii: if you call disconnect(), and then call connectUser() again, it should work since connect will set isDisconnected to false

src/client.ts
mahboubii: yea, will get rid of on another PR

src/connection_fallback.ts
```diff
@@ -0,0 +1,205 @@
...
+    }
+  };
+
+  /** @private */
+  _poll = async () => {
+    while (this.state === ConnectionState.Connected) {
+      try {
+        const data = await this._req<{
+          events: Event<AttachmentType, ChannelType, CommandType, EventType, MessageType, ReactionType, UserType>[];
+        }>({}, { timeout: 30000 }, true); // 30s => API responds in 20s if there is no event
```
ferhatelmas, line 117: 27s to be exact

src/utils.ts
```diff
@@ -203,3 +203,36 @@ export function convertErrorToJson(err: Error) {
...
+ */
+export function isOnline() {
+  const nav =
+    typeof navigator !== 'undefined'
+      ? navigator
+      : typeof window !== 'undefined' && window.navigator
+      ? window.navigator
+      : undefined;
+
+  if (!nav) return true;
```
vicnicius, line 219: Minor: but maybe we could add a warning if there's no navigator present, to alert users trying to use this in other envs.

src/connection.ts
vishalnarkhede: So on RN we use openConnection/closeConnection

src/utils.ts
vicnicius: Same for the other methods in this module. Or alternatively, we could name the methods something else like (sBrowserConnected, isBrowserOnline... not sure about the naming though.

src/connection_fallback.ts
mahboubii: the longpoll responds with empty event list after 20s of no event

ferhatelmas: true, `events := client.ConsumeEvents(10, 500*time.Millisecond, 20*time.Second)`

I remembered 27 for some reason

src/client.ts
vicnicius: maybe `return closePromise.then(this.tokenManager.reset);` would be better for guaranteeing the order of execution?

src/connection.ts
```diff
@@ -95,14 +107,16 @@ export class StableWSConnection<
 
   /**
    * connect - Connect to the WS URL
-   *
+   * the default 15s timeout allows between 2~3 tries
    * @return {ConnectAPIResponse<ChannelType, CommandType, UserType>} Promise that completes once the first health check message is received
    */
-  async connect() {
+  async connect(timeout = 15000) {
```
vicnicius, line 113:
```suggestion
  async connect(options = {timeout: 15000}) {
```
what do you think of an object as argument to give a little bit more clarity to what it does?

src/connection.ts
mahboubii: then it should be OK, but still worth manual testing with both longpoll enabled/disabled situation

src/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+  '69': { name: 'ErrWrongRegion', retryable: false },
+  '70': { name: 'ErrQueryChannelPermissions', retryable: false },
+  '71': { name: 'ErrTooManyConnections', retryable: true },
+  '99': { name: 'AppSuspendedError', retryable: false },
+};
+
+type APIError = Error & { code?: number; isWSFailure?: boolean };
+
+export function isAPIError(error: Error): error is APIError {
+  return (error as APIError).code !== undefined;
```
vicnicius, line 34: why in the type definition do we say code is optional to the APIError type, but in the type guard we say an error is only an APIError if it has a code field?

src/client.ts
```diff
@@ -1436,7 +1449,33 @@ export class StreamChat<
...
+        this.options.enableWSFallback ? this.defaultWSTimeoutWithFallback : this.defaultWSTimeout,
+      );
+    } catch (err) {
+      // run fallback only if it's WS/Network error and not a normal API error
+      // make sure browser is online before even trying the longpoll
+      if (this.options.enableWSFallback && isWSFailure(err) && isOnline()) {
+        this.logger('info', 'client:connect() - WS failed, fallback to longpoll', { tags: ['connection', 'client'] });
+
+        this.wsConnection._destroyCurrentWSConnection();
+        this.wsConnection.disconnect().then(); // close WS so no retry
```
vicnicius, line 1464: Why the `.then()` ?

src/connection_fallback.ts
```diff
@@ -0,0 +1,205 @@
...
+        await sleep(retryInterval(this.consecutiveFailures));
+      }
+    }
+  };
+
+  /**
+   * connect try to open a longpoll request
+   * @param reconnect should be false for first call and true for subsequent calls to keep the connection alive and call recoverState
+   */
+  connect = async (reconnect = false) => {
```
vicnicius, line 153: this connect has `reconnect` as argument, while the other has `timeout`. Maybe better to use an object on both.
```suggestion
  connect = async (options = { reconnect: false }) => {
```

src/client.ts
vicnicius, line 1460: I wonder if we shouldn't remove thhe online check. If you're offline the `this.wsConnection.connect` fails and `isWsFailure` will return false anyways. On the other hand, if it could be the case that you have a wsFailure and you're offline, in RN this check wouldn't work, and we would trigger the fallback while offline. I'm not sure if I'm missing something, though. Does it make sense?

src/utils.ts
mahboubii: warning is a good idea

src/connection.ts
mahboubii: yea i prefer it this way since it's clear to understand the param on the spot, there are more than one place that can benefit from this refactor so let's keep it for another PR

src/errors.ts
mahboubii: good catch, we also need to refactor the way we handle and throw errors in the SDK, worth putting it in the roadmap notion

src/client.ts
mahboubii: disconnect creates and return a promise that's not yet awaited

src/client.ts
vishalnarkhede: Following up on this: so I just tested on RN. If you access `window.navigator` it looks like following:

```
{"product": "ReactNative"}
```

And thus `isOnline()` function will return always `undefined` -> fallback will never be executed. Maybe we can provide a way so that RN SDK can let JS client know about the online status (e.g., we can provide js client with async function to fetch online status).

vishalnarkhede (approved):
LGTM

PR #822, opened 2021-11-25
refactor: simplify StableWSConnection constructor params

mahboubii:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -406,10 +406,10 @@ export class StableWSConnection<
       if (data.error) {
         this.rejectPromise?.(this._errorFromWSEvent(data, false));
         return;
-      } else {
-        this.resolvePromise?.(data);
-        this._setHealth(true);
       }
+
+      this.resolvePromise?.(data);
+      this._setHealth(true);
     }
 
     // trigger the event..
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1,7 +1,7 @@
 import WebSocket from 'isomorphic-ws';
 import { chatCodes, convertErrorToJson, sleep, retryInterval, randomId } from './utils';
 import { buildWsFatalInsight, buildWsSuccessAfterFailureInsight, postInsights } from './insights';
-import { ConnectAPIResponse, ConnectionOpen, LiteralStringForUnion, UR, LogLevel, UserResponse } from './types';
+import { ConnectAPIResponse, ConnectionOpen, LiteralStringForUnion, UR, LogLevel } from './types';
 import { StreamChat } from './client';
 
 // Type guards to check WebSocket error type
@@ -179,8 +179,8 @@ export class StableWSConnection<
    */
   _buildUrl = (reqID?: string) => {
     const params = {
-      user_id: (this.client.user as UserResponse).id,
-      user_details: this.client.user,
+      user_id: this.client.userID,
+      user_details: this.client._user,
       user_token: this.client.tokenManager.getToken(),
       server_determines_connection_id: true,
       device: this.client.options.device,
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1,18 +1,8 @@
 import WebSocket from 'isomorphic-ws';
 import { chatCodes, convertErrorToJson, sleep, retryInterval, randomId } from './utils';
-import { TokenManager } from './token_manager';
-import { buildWsFatalInsight, buildWsSuccessAfterFailureInsight, InsightMetrics, postInsights } from './insights';
-import {
-  BaseDeviceFields,
-  ConnectAPIResponse,
-  ConnectionChangeEvent,
-  ConnectionOpen,
-  LiteralStringForUnion,
-  Logger,
-  UR,
-  UserResponse,
-  LogLevel,
-} from './types';
+import { buildWsFatalInsight, buildWsSuccessAfterFailureInsight, postInsights } from './insights';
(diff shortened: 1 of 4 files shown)


src/connection.ts
```diff
@@ -61,23 +29,19 @@ type Constructor<
...
-  messageCallback: Constructor<ChannelType, CommandType, UserType>['messageCallback'];
-  recoverCallback: Constructor<ChannelType, CommandType, UserType>['recoverCallback'];
-  tokenManager: Constructor<ChannelType, CommandType, UserType>['tokenManager'];
-  user: Constructor<ChannelType, CommandType, UserType>['user'];
-  userAgent: Constructor<ChannelType, CommandType, UserType>['userAgent'];
-  userID: Constructor<ChannelType, CommandType, UserType>['userID'];
-  wsBaseURL: Constructor<ChannelType, CommandType, UserType>['wsBaseURL'];
-  device: Constructor<ChannelType, CommandType, UserType>['device'];
+  // global from constructor
+  client: StreamChat<AttachmentType, ChannelType, CommandType, EventType, MessageType, ReactionType, UserType>;
```
ferhatelmas, line 42: nice touch

ferhatelmas (approved):
LGTM

Amazing simplification 🌮

src/connection.ts
```diff
@@ -61,23 +29,19 @@ type Constructor<
  * - if the servers fails to publish a message to the client, the WS connection is destroyed
  */
 export class StableWSConnection<
+  // CAUTION: generics are out of usual order here
```
vicnicius, line 32: Why can't we have them in the usual order?

src/connection.ts
```diff
@@ -647,30 +576,21 @@ export class StableWSConnection<
...
+    this.connectionOpen = new Promise<WebSocket.MessageEvent>((resolve, reject) => {
+      this.resolvePromise = resolve;
+      this.rejectPromise = reject;
+    }).then((e) => {
+      if (!e.data || typeof e.data !== 'string') return;
+
+      const data = JSON.parse(e.data) as ConnectionOpen<ChannelType, CommandType, UserType> & { error?: unknown };
+      if (data?.error) {
+        throw new Error(JSON.stringify(data.error));
+      }
```
vishalnarkhede, line 590: I think we can also get rid of this, since we already have this check in `onMessage` function ... and thats the only place where resolvePromise gets called.

```tsx
  onmessage = (wsID: number, event: WebSocket.MessageEvent) => {
    if (this.wsID !== wsID) return;

    const data = typeof event.data === 'string' ? JSON.parse(event.data) : null;

    // we wait till the first message before we consider the connection open..
    // the reason for this is that auth errors and similar errors trigger a ws.onopen and immediately
    // after that a ws.onclose..
    if (!this.isResolved && data) {
      this.isResolved = true;
      if (data.error != null) { // <-------- THIS ONE
        this.rejectPromise?.(this._errorFromWSEvent(data, false));
        return;
      } else {
        this.resolvePromise?.(event);
        this._setHealth(true);
      }
    }

    // trigger the event..
    this.lastEvent = new Date();
    this.logger('info', 'connection:onmessage() - onmessage callback', {
      tags: ['connection'],
      event,
      wsID,
    });

    if (data && data.type === 'health.check') {
      this.scheduleNextPing();
    }

    this.messageCallback(event);
    this.scheduleConnectionCheck();
  };
```

vishalnarkhede: Check for error I mean!!

src/connection.ts
vicnicius: I see the previous interface with less args was different. 👍

vishalnarkhede (commented):
I am loving these changes!! Thanks alot @mahboubii ❤️

src/connection.ts
vishalnarkhede: It also opens a possibility of getting rid of this `then` block.

vishalnarkhede: Or even `_setupConnectionPromise`, not 100% sure though ... just wild idea

mahboubii: good stuff, removed duplication here https://github.com/GetStream/stream-chat-js/pull/822/commits/f672fdc8fa9a82199d7b9650eabad4143db291cd

vishalnarkhede (dismissed):

src/connection.ts
mahboubii: yea, but actually I'm thinking we can reorder them correctly, this is a private/internal class and should not be considered a breaking change, @vishalnarkhede wdyt?

mahboubii: also wont matter much since we're going to refactor the generics

src/connection.ts
```diff
@@ -247,16 +179,18 @@ export class StableWSConnection<
...
+      user_token: this.client.tokenManager.getToken(),
       server_determines_connection_id: true,
-      device: this.device,
+      device: this.client.options.device,
       client_request_id: reqID,
     };
     const qs = encodeURIComponent(JSON.stringify(params));
-    const token = this.tokenManager.getToken();
-    return `${this.wsBaseURL}/connect?json=${qs}&api_key=${this.apiKey}&authorization=${token}&stream-auth-type=${this.authType}&X-Stream-Client=${this.userAgent}`;
+    const token = this.client.tokenManager.getToken();
```
vicnicius, line 190: do we really use `token` and `user_token` on this call?

mahboubii: good catch, could be an old mistake

PR #849, opened 2021-12-17
fix: prevent own_capabilities and hidden set to undefined

yetieaterxb1:
prevent own_capabilities and hidden from being set to undefined after a channel.updated event is fired

link to issue: https://github.com/GetStream/stream-chat-react/issues/1310

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1319,8 +1319,8 @@ export class Channel<
         if (event.channel) {
           channel.data = {
             ...event.channel,
-            own_capabilities: channel.data?.own_capabilities,
-            hidden: channel.data?.hidden,
+            hidden: event.channel?.hidden ?? channel.data?.hidden,
+            own_capabilities: event.channel?.own_capabilities ?? channel.data?.own_capabilities,
           };
         }
         break;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1317,7 +1317,11 @@ export class Channel<
         break;
       case 'channel.updated':
         if (event.channel) {
-          channel.data = event.channel;
+          channel.data = {
+            ...event.channel,
+            own_capabilities: channel.data?.own_capabilities,
+            hidden: channel.data?.hidden,
+          };
         }
         break;
       case 'reaction.new':


AnatolyRugalev (approved):

mahboubii:
Could this cause sync issues? Maybe better to keep old data if new channel data is missing it

petyosi:
> Could this cause sync issues? Maybe better to keep old data if new channel data is missing it

@amin How can that happen?

BrianLi101:
Thanks for the fix!

Quick followup, what happens if the `own_capabilities` ends up changing after someone already joined a Channel? For example, they're promoted to a `moderator` and the channel was already watched. This PR technically uses the existing `own_capabilities` on the channel and wouldn't end up updating them off of the new ones from the event.

mahboubii:
> > Could this cause sync issues? Maybe better to keep old data if new channel data is missing it
> 
> @amin How can that happen?

@petyosi Talked with the team and looks like this is not the case now since this event only fired for specific cases. [changed](https://github.com/GetStream/stream-chat-js/pull/849/commits/3f4f8ac21240c453fc73c848a86582db8c74aa51) it a bit to avoid potential future issues

mahboubii:
> Thanks for the fix!
> 
> Quick followup, what happens if the `own_capabilities` ends up changing after someone already joined a Channel? For example, they're promoted to a `moderator` and the channel was already watched. This PR technically uses the existing `own_capabilities` on the channel and wouldn't end up updating them off of the new ones from the event.
> 
> Even with the above possible concerns, it would be great if this could be deployed as soon as possible! We would prefer to use it as a fully supported patch-fix

@BrianLi101 thanks for bringing this up. It is not the case as of now for this event, `own_capabilities` are updated in a different event.

BrianLi101:
@mahboubii Thanks for clarifying! Do you know what the timeline for merging this looks like?

petyosi (approved):

mahboubii:
> @mahboubii Thanks for clarifying! Do you know what the timeline for merging this looks like?

@BrianLi101 thanks for waiting, fix is released in [v5.1.1](https://github.com/GetStream/stream-chat-js/releases/tag/v5.1.1)

BrianLi101:
@mahboubii Thank you!

PR #854, opened 2021-12-28
BREAKING: reorganize type names

AnatolyRugalev:

This PR fixes some type name inconsistencies and also removes obsolete entities

### updateChannel changes

* Rename `UpdateChannelOptions` -> `UpdateChannelTypeOptions`
* Rename `InviteOptions` -> `UpdateChannelOptions`
* Add missing fields to `UpdateChannelOptions`
* Add `AddChannelMember` and `UpdateChannelMember` types
* Removed incomplete `ChannelUpdateOptions`

### Permission changes

* Updated `Role` type to be just `string`
* Renamed `Role` -> `RoleName`
* Added `Role` object type and corresponding response types
* Removed redundant Policy and Permission types 

We can split this into two PRs


src/channel.ts
```diff
@@ -476,41 +476,41 @@ export class Channel<
...
-    options: InviteOptions<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType> = {},
+    options: UpdateChannelOptions<AttachmentType, ChannelType, CommandType, MessageType, UserType> = {},
   ) {
     return await this._update({ accept_invite: true, ...options });
   }
 
   /**
    * rejectInvite - reject invitation to the channel
    *
-   * @param {InviteOptions<AttachmentType, ChannelType, CommandType, MessageType, ReactionType, UserType>} [options] The object to update the custom properties of this channel with
```
madsroskar, line 492: WE don't actually need the jsDoc type definitions anymore, right? Or are there cases TS typings doesn't cut it if you know?

src/permissions.ts
```diff
@@ -1,31 +1,48 @@
...
-  action: RequiredPermissionObject['action'];
-  owner: RequiredPermissionObject['owner'];
-  priority: RequiredPermissionObject['priority'];
-  resources: RequiredPermissionObject['resources'];
-  roles: RequiredPermissionObject['roles'];
+/*
+ * Policies are entities for permissions v1. They are deprecated and will be removed in the future release
+ */
+export class Policy {
+  action?: 'Deny' | 'Allow';
```
madsroskar, line 31: There's a check failing related to permissions, is that related to this change? Idk if it's just a matter of renaming in the test if so, but 🤞🏻

src/permissions.ts
```diff
@@ -36,46 +53,17 @@ export class Permission {
...
 
-export const BuiltinRoles = {
-  Admin: 'admin',
-  Anonymous: 'anonymous',
-  ChannelMember: 'channel_member',
-  ChannelModerator: 'channel_moderator',
-  Guest: 'guest',
-  User: 'user',
-};
+// deprecated
```
madsroskar, line 65:
```suggestion
// @deprecated will be removed in the next major release
```
No idea when it would actually be removed, that's just guessing on my end 😅 But having it a jsdoc comment plays nice with most editors/LSP/whatever I think, does `// deprecated` do it normally as well? I might have to rewrite some config on my end for that :)

madsroskar (commented):
Haven't tried this out yet, so don't consider this a review, but so far looks good to me. I like the changes :) 👍🏻

Note to self, we'll have to update the import of roles we use in the RN SDK (I think), adding it on our notion.

PR #859, opened 2022-01-04
Don't add messages from shadow banned users to state

madsroskar:
## CLA

- [x] Code changes are tested

We have gotten two tickets recently pointing out that we currently
display messages from shadow-banned users to every user, and not only
the shadow-banned user.

The docs specify that `message.shadowed` will not be set on a message
from the backend for the connected user, meaning the JS client won't
need to also check whether the connected user is shadowbanned, so this
only checks whether the message is specified as `shadowed`.

As far as I can see so far, this seems to work with the RN SDK fine, but
if anybody can think of ways this wouldn't be sufficient to keep the
messages from being displayed on the frontend, please don't hesitate to
point me the right direction.

## Changelog

- Added check to ensure messages from shadowbanned users aren't added to a channels state


gumuz (approved):
LGTM

ferhatelmas (approved):
LGTM

PR #860, opened 2022-01-05
[CHAT-2049] Add async user export endpoint support

SiddhantAgarwal:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -353,6 +353,10 @@ export type ExportChannelResponse = {
   task_id: string;
 };
 
+export type ExportUsersResponse = {
+  task_id: string;
+};
+
 export type ExportChannelStatusResponse = {
   created_at?: string;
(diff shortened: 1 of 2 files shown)


AnatolyRugalev (approved):

gumuz (approved):

src/types.ts
```diff
@@ -353,6 +353,10 @@ export type ExportChannelResponse = {
   task_id: string;
 };
 
+export type ExportUsersResponse = {
```
mahboubii, line 356: is there a difference or will be in future between this and `ExportChannelResponse`?

SiddhantAgarwal: Currently there is no difference, but wanted to keep it separate just in case we decide to diverge in the future.

PR #868, opened 2022-01-14
feat: add unblockmessage

ffenix113:
## Changelog

- Add `_unblockMessage` private function

Chat MR: https://github.com/GetStream/chat/pull/2969

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2304,6 +2304,24 @@ export class StreamChat<
   }
 
   /**
+   * unblockMessage - unblocks message blocked by automod
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
+   * @param {string} targetMessageID
+   * @param {string} [options.user_id] currentUserID, only used with serverside auth
+   * @returns {Promise<APIResponse>}
+   */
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -2303,6 +2303,24 @@ export class StreamChat<
     });
   }
 
+  /**
+   * unblockMessage - unblocks message blocked by automod
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
```
mahboubii, line 2312:
```suggestion
   *
   * @private
```

mahboubii (dismissed):

src/client.ts
ffenix113: Thanks for this, will add for other new private methods.

ffenix113: Updated, please check.

mahboubii (approved):

ferhatelmas (approved):

ffenix113:
@peterdeme Seems like `${{ github.event.pull_request.title }}` does not contain latest changes, only initial event, so it will not pass PR name lint if firstly PR name was incorrect. Can we take it into account please?

peterdeme (approved):

PR #874, opened 2022-01-20
ci: fix config for the initiate_release workflow

peterdeme:


madsroskar (approved):
LGTM, but can the title be changed to reflect what part of CI is being changed? In case we need to look for it in the history later. Something like `fix config for the initiate_release workflow` or something

peterdeme:
@madsroskar definitely, I was too lazy with it haha

ferhatelmas (approved):

PR #875, opened 2022-01-20
ci: fix initiate_release job once again

peterdeme:
fix the with/script section of the yaml

### DIFF
--- a/.github/workflows/initiate_release.yml
+++ b/.github/workflows/initiate_release.yml
@@ -41,13 +41,14 @@ jobs:
         uses: actions/github-script@v5
         env:
           VERSION: ${{ github.event.inputs.version }}
-        with: |
-          npm run changelog
-          git config --global user.name 'github-actions' 
-          git config --global user.email 'release@getstream.io'
-          git checkout -q -b "release-$VERSION"
-          git commit -am "chore(release): $VERSION"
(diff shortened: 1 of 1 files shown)


ferhatelmas (approved):

.github/workflows/initiate_release.yml
```diff
@@ -41,13 +41,14 @@ jobs:
...
           VERSION: ${{ github.event.inputs.version }}
-        with: |
-          npm run changelog
-          git config --global user.name 'github-actions' 
-          git config --global user.email 'release@getstream.io'
-          git checkout -q -b "release-$VERSION"
-          git commit -am "chore(release): $VERSION"
-          git push -q -u origin "release-$VERSION"
+        with:
+          script: |
```
madsroskar, line 45: This doesn't seem to be valid syntax in [the docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)? Though I might just not see it

Is it maybe easier to move it into a new step, and have `run: ....` there?

peterdeme: it's valid, we already use it every where like here:
https://github.com/GetStream/stream-chat-net/blob/master/.github/workflows/initiate_release.yml#L24

PR #883, opened 2022-01-20
feat: add xiaomi push provider

ferhatelmas:
Also, drop huawei data template which was never used.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -99,6 +99,7 @@ export type AppSettingsAPIResponse<CommandType extends string = LiteralStringFor
       apn?: APNConfig;
       firebase?: FirebaseConfig;
       huawei?: HuaweiConfig;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -286,13 +287,16 @@ export type ChannelMemberResponse<UserType = UR> = {
 
 export type CheckPushResponse = APIResponse & {
   device_errors?: {
-    error_message?: string;
-    provider?: string;
+    [deviceID: string]: {
```
driver-devel, line 290: Have we had the wrong structure all this time ?

ferhatelmas: Yes, it seems type was wrong. It's not a single one but a map of errors with their ids

driver-devel (approved):

peterdeme (approved):

PR #884, opened 2022-01-20
feat: imports API endpoints

gumuz:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3074,19 +3074,22 @@ export class StreamChat<
     });
   }
 
-  async getImportUploadPath(name: string) {
-    return this.get<APIResponse>(this.baseURL + `/imports/upload_path`, { name });
+  async getImportUploadPath(id: string) {
+    return this.get<APIResponse>(this.baseURL + `/imports/${id}/upload_path`);
   }
 
-  async createImport(path: string) {
-    return this.post<APIResponse>(this.baseURL + `/imports`, { path });
+  async createImport(filename: string) {
+    const { import_task } = await this.post<APIResponse & { import_task: {} }>(this.baseURL + `/imports`, { filename });
+    return import_task;
   }
 
   async getImport(id: string) {
-    return this.get<APIResponse>(this.baseURL + `/imports/${id}`);
+    const { import_task } = await this.get<APIResponse & { import_task: {} }>(this.baseURL + `/imports/${id}`);
+    return import_task;
   }
 
-  async listImports() {
-    return this.get<APIResponse>(this.baseURL + `/imports`);
+  async listImports(options: { limit?: number; offset?: number }) {
+    const { import_tasks } = await this.get<APIResponse & { import_tasks: {}[] }>(this.baseURL + `/imports`, options);
+    return import_tasks;
   }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -3073,4 +3073,8 @@ export class StreamChat<
       ...options,
     });
   }
+
+  async getImportUploadPath(name: string) {
+    return this.get<APIResponse>(this.baseURL + `/imports/upload_path`, { name });
+  }
 }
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -3073,4 +3073,18 @@ export class StreamChat<
       ...options,
     });
   }
+
+  async createImport(filename: string) {
```
ferhatelmas, line 3077: let's add some notes that they are internal and can change any time

ferhatelmas: maybe it might be better to prefix with _ to make explicit the intention

src/client.ts
```diff
@@ -3073,4 +3073,54 @@ export class StreamChat<
...
+   * _listImports - Lists Import Tasks.
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
+   * @private
+   * @param {{ limit?: number; offset?: number }} options pagination options
+   *
+   * @return {APIResponse} An ImportTask
```
peterdeme, line 3121: We should have a specific response type for this method and for the others as well, right?

gumuz: I'll add it

ferhatelmas: this isn't exactly true since it's a richer version of it 

maybe we should define listResponse, getResponse etc. How is it done in other endpoints?

src/types.ts
```diff
@@ -2137,3 +2137,18 @@ export type TruncateOptions<AttachmentType, MessageType, UserType> = {
...
+    created_at: string;
+    next_state: string;
+    prev_state: string;
+  }[];
+  id: string;
+  state: string;
+  updated_at: string;
+  result?: UR;
+  size?: number;
+};
```
peterdeme, line 2154: I guess that's a bit more consistent with other types in this file:

```suggestion
export type ImportTaskHistory = {
    created_at: string;
    next_state: string;
    prev_state: string;
};

export type ImportTask = {
  created_at: string;
  filename: string;
  history: ImportTaskHistory[];
  id: string;
  state: string;
  updated_at: string;
  result?: UR;
  size?: number;
};
```

src/client.ts
```diff
@@ -3073,4 +3074,54 @@ export class StreamChat<
...
+   * _createImport - Create an Import Task.
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
+   * @private
+   * @param {string} filename filename of uploaded data
+   *
+   * @return {APIResponse} An ImportTask
```
peterdeme, line 3088:
```suggestion
   * @return {APIResponse & { import_task: ImportTask; upload_url: string }} An ImportTask
```

src/client.ts
```diff
@@ -3073,4 +3074,54 @@ export class StreamChat<
...
+   * _getImport - Get an Import Task.
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
+   * @private
+   * @param {string} id id of Import Task
+   *
+   * @return {APIResponse} An ImportTask
```
peterdeme, line 3106:
```suggestion
   * @return {APIResponse & { import_task: ImportTask }} An ImportTask
```

src/client.ts
```diff
@@ -3073,4 +3074,54 @@ export class StreamChat<
...
+   * _listImports - Lists Import Tasks.
+   *
+   * Note: Do not use this.
+   * It is present for internal usage only.
+   * This function can, and will, break and/or be removed at any point in time.
+   *
+   * @private
+   * @param {{ limit?: number; offset?: number }} options pagination options
+   *
+   * @return {APIResponse} An ImportTask
```
peterdeme, line 3122:
```suggestion
   * @return {APIResponse & { import_tasks: ImportTask[] }} An ImportTask
```

src/types.ts
ferhatelmas: good one

peterdeme (approved):

peterdeme:
lgtm

ferhatelmas (approved):
LGTM

PR #892, opened 2022-02-01
fix: added transport.changed event to event map

vishalnarkhede:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,5 @@
 import { AxiosRequestConfig } from 'axios';
+import { EVENT_MAP } from './events';
 import { Role } from './permissions';
 
 /**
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -1052,50 +1053,7 @@ export type EventHandler<
...
-  | 'reaction.updated'
-  | 'typing.start'
-  | 'typing.stop'
-  | 'user.banned'
-  | 'user.deleted'
-  | 'user.presence.changed'
-  | 'user.unbanned'
-  | 'user.updated'
-  | 'user.watching.start'
-  | 'user.watching.stop';
```
ferhatelmas, line 1098: this list is bigger than the other

vishalnarkhede: This list is smaller actually than the other. Its missing `channel.kicked` event, which is fixed by this change

vishalnarkhede: This one has 42 types (excluding `all`), and other one has 43.

ferhatelmas: right, it was diff not expanded completely. The place of event might be local ones though

ferhatelmas: otherwise, good reduction in duplication

ferhatelmas (approved):
LGTM

AnatolyRugalev (approved):

PR #900, opened 2022-02-08
fix: fire health.check event after successful connection when using long poll transport

thesyncim:


ferhatelmas (approved):
LGTM

ferhatelmas:
there are breaking changes on master though, release might be troublesome

vishalnarkhede:
> there are breaking changes on master though, release might be troublesome

@ferhatelmas which one is the breaking change?

ferhatelmas:
node v10 drop

PR #904, opened 2022-02-10
fix: unread count to not increment if channel has read_events off

khushal87:
## CLA

- [x] Code changes are tested

If the read events are turned off on the application, then the unread count shouldn’t show up on the channel list component. Currently, when a new message comes in, we increment the unread count for a particular channel. But this increment works even if read events are turned off, which leads to confusing behavior.

## Changelog

- Add the logic for not incrementing channel unread count if the read_events are off.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -923,6 +923,10 @@ export class Channel<
     if (message.user?.id && this.getClient().userMuteStatus(message.user.id)) return false;
     if (message.type === 'system') return false;
 
+    // Return false if channel doesn't allow read events.
+    if (Array.isArray(this.data?.own_capabilities) && !this.data?.own_capabilities.includes('read_events'))
+      return false;
+
     if (this.muteStatus().muted) return false;
 
     return true;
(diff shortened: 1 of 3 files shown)


vishalnarkhede (dismissed):

src/channel.ts
```diff
@@ -863,6 +863,10 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     if (message.user?.id && this.getClient().userMuteStatus(message.user.id)) return false;
     if (message.type === 'system') return false;
 
+    // Return false if channel doesn't allow read events.
+    if (Array.isArray(this.data?.own_capabilities) && !this.data?.own_capabilities.includes('read_events'))
```
AnatolyRugalev, line 867: it's `read-events`

AnatolyRugalev (approved):

src/channel.ts
khushal87: @AnatolyRugalev Fixed 😄

PR #905, opened 2022-02-10
fix: added attribute `pendingDisconnect`

arnautov-anton:
## CLA

- [ ] Code changes are tested

Whenever `disconnectUser` got called in a "cleaning" function of a React component (let's say user got switched) the developer had to store the return value of this method (a Promise object) to be then awaited before new connection could be established - without the said await the effect would attempt to establish a new connection right after cleaning function finished running which resulted in a warning and a connection error:

image
image

This PR adds this "await" functionality to the SDK so the user does not have to do it themselves.
Third parameter could be added to the `connectUser` - `skipDisconnectAwait` to prevent possible breaking changes but I'd expect this to be a default behaviour.

Without the fix:
```ts
const disconnectReference = useRef(Promise.resolve());

useEffect(() => {
	if (!user) return;

	disconnectReference.current.then(() => {
		chatAPIClient.connectUser({ id: user.id }, user.token);
	});

	return () => {
		disconnectReference.current = chatAPIClient.disconnectUser();
	};
}, [user]);
```
With the fix:
```ts
useEffect(() => {
	if (!user) return;

	chatAPIClient.connectUser({ id: user.id }, user.token);

	return () => {
		chatAPIClient.disconnectUser();
	};
}, [user]);
```

## Changelog
- added attribute `pendingDisconnect` to the `StreamChat` class

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -195,6 +195,8 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   insightMetrics: InsightMetrics;
   defaultWSTimeoutWithFallback: number;
   defaultWSTimeout: number;
+  /** promise which gets set on disconnectUser and gets awaited (if set) before creating a new connection */
+  pendingDisconnect: null | Promise<void>;
 
   /**
    * Initialize a client
@@ -282,6 +284,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
 
     this.defaultWSTimeoutWithFallback = 6000;
     this.defaultWSTimeout = 15000;
+    this.pendingDisconnect = null;
 
     /**
      * logger function should accept 3 parameters:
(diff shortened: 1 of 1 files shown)


petyosi (dismissed):
This looks good to me. As a matter of fact, the current behavior can technically be considered a bug. @imtoori - how do you do it in Flutter?

src/client.ts
```diff
@@ -480,6 +483,9 @@ export class StreamChat<
       throw new Error('The "id" field on the user is missing');
     }
 
+    if (this.pendingDisconnect) await this.pendingDisconnect;
```
szuperaz, line 486: I think we should include this promise in the `setUserPromise` that is used by the `queryUsers` and `queryChannels` methods to wait for pending connects

src/client.ts
```diff
@@ -507,14 +510,23 @@ export class StreamChat<
...
+        (result) => result[1], // We only return connection promise;
+      );
+    };
+
+    this.setUserPromise = this.pendingDisconnect
+      ? this.pendingDisconnect.then(() => {
+          this.pendingDisconnect = null;
+          return startConnecting();
+        })
+      : startConnecting();
```
arnautov-anton, line 529: We could do either this or `this.pendingDisconnect` could be always set to some kind of a Promise (`Promise.resolve()` for example) and that way I wouldn't have to check whether it's `null` or not (also `startConnecting` wouldn't have to exist).
@petyosi @szuperaz

PR #914, opened 2022-02-24
fix: channel unreadCount to be set as 0 when notification.mark_read event is dispatched [CRNS - 433]

khushal87:
## CLA

- [x] Code changes are tested

When we use the markAllRead method, we get the event notification.mark_read, which updates the total_unread_count, but not the channel.state.unreadCount. This PR solved the above issue.

## Changelog

- Added the logic to set the activeChannel state's unreadCount to 0.
- Added tests for the same.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1186,7 +1186,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
 
     if (event.type === 'notification.mark_read') {
       const activeChannelKeys = Object.keys(this.activeChannels);
-      activeChannelKeys.map((activeChannelKey) => (this.activeChannels[activeChannelKey].state.unreadCount = 0));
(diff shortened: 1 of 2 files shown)


AnatolyRugalev (dismissed):

src/client.ts
```diff
@@ -1184,6 +1184,11 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       this.mutedUsers = event.me.mutes;
     }
 
+    if (event.type === 'notification.mark_read') {
+      const activeChannelKeys = Object.keys(this.activeChannels);
+      activeChannelKeys.map((activeChannelKey) => (this.activeChannels[activeChannelKey].state.unreadCount = 0));
```
peterdeme, line 1189: Shouldn't this be `.forEach(activeChannelKey  =>`?

khushal87: Yes, that's a better method to handle this case. Thanks, @peterdeme 😄

peterdeme (approved):

PR #927, opened 2022-03-17
feat: added import mode

gumuz:
fixes https://github.com/GetStream/chat/issues/3502

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -132,6 +132,7 @@ import {
   ExportUsersRequest,
   ExportUsersResponse,
   CreateImportResponse,
+  CreateImportOptions,
   CreateImportURLResponse,
   GetImportResponse,
   ListImportsResponse,
@@ -538,10 +539,10 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   _setupConnection = this.openConnection;
 
   /**
-	 * updateAppSettings - updates application settings
-	 *
-	 * @param {AppSettings} options App settings.
-	 * 		IE: {
+   * updateAppSettings - updates application settings
+   *
+   * @param {AppSettings} options App settings.
+   * 		IE: {
 	  			"apn_config": {
 					"auth_type": "token",
 					"auth_key": fs.readFileSync(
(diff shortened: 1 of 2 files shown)


yaziine (approved):

src/client.ts
```diff
@@ -2860,12 +2860,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    *
    * @private
    * @param {string} path path of uploaded data
-   *
+   * @param {string} mode import mode insert|upsert
    * @return {APIResponse & CreateImportResponse} An ImportTask
    */
-  async _createImport(path: string) {
+  async _createImport(path: string, mode: string) {
```
ferhatelmas, line 2866: it might wiser to take second param as an option object so that we can add more flags in the future if needed

ferhatelmas: since it's noted to be changed, not strict though

src/client.ts
```diff
@@ -2864,12 +2865,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    *
    * @private
    * @param {string} path path of uploaded data
-   *
+   * @param {CreateImportOptions} options import options
    * @return {APIResponse & CreateImportResponse} An ImportTask
    */
-  async _createImport(path: string) {
+  async _createImport(path: string, options: CreateImportOptions) {
```
ferhatelmas, line 2871: accept default too

ferhatelmas: `_createImport(path)` should be fine

src/client.ts
```diff
@@ -624,18 +625,18 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   }
 
   /**
-	 * testPushSettings - Tests the push settings for a user with a random chat message and the configured push templates
```
gumuz, line 627: not sure why linter didn't catch this before @ferhatelmas

ferhatelmas: yes interesting

src/client.ts
```diff
@@ -2866,12 +2867,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    *
    * @private
    * @param {string} path path of uploaded data
-   *
+   * @param {CreateImportOptions} options import options
    * @return {APIResponse & CreateImportResponse} An ImportTask
    */
-  async _createImport(path: string) {
+  async _createImport(path: string, options: CreateImportOptions = { mode: 'upsert' }) {
```
ferhatelmas, line 2873: this is good, but we could treat missing as upsert probably

ferhatelmas (approved):

ferhatelmas (approved):

PR #933, opened 2022-03-23
feat: add provider filtering into check push

ferhatelmas:


src/types.ts
```diff
@@ -295,14 +295,15 @@ export type CheckPushResponse = APIResponse & {
   device_errors?: {
     [deviceID: string]: {
       error_message?: string;
-      provider?: string;
+      provider?: PushProvider;
+      provider_name?: string;
```
ferhatelmas, line 299: @driver-devel expose name here under devices

driver-devel (approved):

PR #934, opened 2022-03-23
fix: catch tokenProvider rejection

pyrho:
## CLA

- [x] Code changes are tested

When providing a `tokenProvider` function to `connectUser`, if the `tokenProvider` promise is rejected this will result in an uncaught promise rejection.

This fix adds a `try/catch` around the `await`ing of the user provided `tokenProvider` function as to gracefully handle the error.

## Changelog
- fix: catch tokenProvider rejection

### DIFF
--- a/test/unit/connection.js
+++ b/test/unit/connection.js
@@ -44,6 +44,18 @@ describe('connection', function () {
 
 	after(() => wss.close());
 
+	describe('Connection tokenProvider', () => {
+		it('should handle token provider rejection ', async () => {
+			const client = new StreamChat('apiKey', {
+				allowServerSideConnect: true,
+				baseURL: 'http://localhost:9999', // invalid base url
+			});
+			client.defaultWSTimeout = 20;
+			const tokenProvider = () => Promise.reject(new Error('network failure'));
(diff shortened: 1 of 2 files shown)


test/unit/connection.js
```diff
@@ -44,6 +44,18 @@ describe('connection', function () {
 
 	after(() => wss.close());
 
+	describe('Connection tokenProvider', () => {
+		it('should handle token provider rejection ', async () => {
+			const client = new StreamChat('apiKey', {
+				allowServerSideConnect: true,
+				baseURL: 'http://localhost:9999', // invalid base url
```
arnautov-anton, line 51:
```suggestion
```
Not sure why this is needed, could you elaborate?

arnautov-anton (commented):
Thank you for your contribution, @pyrho! Looks like a no-brainer to me, what do you think, @MartinCupela and @petyosi?

test/unit/connection.js
pyrho: I was not sure what the required parameters were for a proper `StreamChat` instantiation. If it works without it, we can remove it; iirc I just copy/pasted this from another test in this file :sweat_smile:

arnautov-anton (approved):
LGTM :taco:

PR #942, opened 2022-04-05
fix: added missing push and reminders related types 

vishalnarkhede:
## Changelog

- Fixed the typescript for push config
- Added missing types for `automod_threshold`
- Added missing types for `reminders `
- Move the api keys and secrets for typescript test to github secret

### DIFF
--- a/.github/workflows/type.yml
+++ b/.github/workflows/type.yml
@@ -13,3 +13,8 @@ jobs:
       - uses: ./.github/actions/setup-node
       - name: Run API type definition tests
         run: yarn test-types
+        env:
+          API_KEY: ${{ secrets.TS_TEST_API_KEY }}
+          API_SECRET: ${{ secrets.TS_TEST_API_SECRET }}
+          MULTITENANCY_API_KEY: ${{ secrets.TS_TEST_MULTITENANCY_API_KEY }}
+          MULTITENANCY_API_SECRET: ${{ secrets.TS_TEST_MULTITENANCY_API_SECRET }}
--- a/src/types.ts
+++ b/src/types.ts
@@ -75,8 +75,10 @@ export type AppSettingsAPIResponse<StreamChatGenerics extends ExtendableGenerics
     channel_configs: Record<
       string,
       {
+        reminders: boolean;
         automod?: ChannelConfigAutomod;
         automod_behavior?: ChannelConfigAutomodBehavior;
+        automod_thresholds?: ChannelConfigAutomodThresholds;
         blocklist_behavior?: ChannelConfigAutomodBehavior;
         commands?: CommandVariants<StreamChatGenerics>[];
         connect_events?: boolean;
@@ -98,6 +100,7 @@ export type AppSettingsAPIResponse<StreamChatGenerics extends ExtendableGenerics
         url_enrichment?: boolean;
       }
     >;
+    reminders_interval: number;
     async_url_enrich_enabled?: boolean;
     auto_translation_enabled?: boolean;
     before_message_send_hook_url?: string;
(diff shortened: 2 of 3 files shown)


ferhatelmas (approved):

src/types.ts
```diff
@@ -1956,7 +1962,7 @@ export type Policy = {
   owner?: boolean;
   priority?: number;
   resources?: string[];
-  roles?: string[];
+  roles?: string[] | null;
```
peterdeme, line 1965: isn't `?` means it's nullable already?

.github/workflows/type.yml
```diff
@@ -9,7 +9,10 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/setup-node
+      uses: actions/checkout@v2
+      uses: ./.github/actions/setup-node
```
peterdeme, line 13: this syntay isn't valid

src/types.ts
```diff
@@ -75,8 +75,10 @@ export type AppSettingsAPIResponse<StreamChatGenerics extends ExtendableGenerics
     channel_configs: Record<
       string,
       {
+        reminders: boolean;
```
peterdeme, line 78: i think this fails prettier: the properties should be in an alphabetical order

src/types.ts
```diff
@@ -1542,8 +1545,10 @@ export type ChannelConfigAutomodThresholds = null | {
 };
 
 export type ChannelConfigFields = {
+  reminders: boolean;
```
peterdeme, line 1548: i think this fails prettier: the properties should be in an alphabetical order

src/types.ts
AnatolyRugalev: `?` means `string[] | undefined` which is a bit different from `null`. So the result type here is `string[] undefined | null`

.github/workflows/type.yml
vishalnarkhede: haha yeah I am trying to figure out whats wrong exactly

test/typescript/response-generators/client.js
```diff
@@ -174,7 +174,11 @@ async function syncTeam() {
 		text: 'New Event?',
 		user: { id: user1 },
 	});
-	return await client.sync([channel.cid], new Date(Date.now() - 1000 * 60).toISOString());
+	const response = await client.sync([channel.cid], new Date(Date.now() - 1000 * 60).toISOString());
+
+	const serverClient = await utils.disableMultitenancy();
```
AnatolyRugalev, line 179: This won't play well with concurrent runs

src/types.ts
AnatolyRugalev: > yeah but does it make sense here?

That's a good question :) I don't think this should be necessary

vishalnarkhede: not to me ... but backend sent `null` in one of our test case

test/typescript/response-generators/client.js
vishalnarkhede: hmm good point, creating the separate app is the only way then ... for multi-tenancy!!

src/types.ts
vishalnarkhede: I can remove it if we can ensure it never null from backend response

src/types.ts
vishalnarkhede: Its a required property ... they get higher priority over optional ones

src/types.ts
```diff
@@ -98,6 +100,7 @@ export type AppSettingsAPIResponse<StreamChatGenerics extends ExtendableGenerics
         url_enrichment?: boolean;
       }
     >;
+    reminders_interval: number;
```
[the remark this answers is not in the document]
vishalnarkhede: Same ... required property

src/types.ts
vishalnarkhede: Nope ... required vs optional

test/typescript/response-generators/client.js
vishalnarkhede: ok just gonna setup different app

AnatolyRugalev (approved):

PR #963, opened 2022-05-02
fix(delete_user): add task_id to deleteuser response

peterdeme:


src/client.ts
```diff
@@ -1863,7 +1863,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       mark_messages_deleted?: boolean;
     },
   ) {
-    return await this.delete<APIResponse & { user: UserResponse<StreamChatGenerics> }>(
+    return await this.delete<APIResponse & { user: UserResponse<StreamChatGenerics> } & { task_id: string }>(
```
ferhatelmas, line 1866: good catch but it should be optional because it's only returned server side where we run tests. Client side it won't be returned

ferhatelmas:
```suggestion
    return await this.delete<APIResponse & { user: UserResponse<StreamChatGenerics> } & { task_id?: string }>(
```

.gitignore
```diff
@@ -66,5 +66,6 @@ typings/
 
 /.vscode
 /.idea
+.envrc
```
ferhatelmas, line 69: this is something we can think generalization in one place

peterdeme: I could add it to my own gitignore I guess :)

ferhatelmas: I use it too, change is good. In terms of hack weeks, might want to think about how to maintain this and similar changes in an easier way

ferhatelmas (approved):

PR #966, opened 2022-05-10
fix: rename sendAt field to scheduledFor

yaziine:
This is breaking but no one uses this endpoint yet, safe then.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2744,14 +2744,14 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    * scheduleCampaign - Schedule a Campaign
    *
    * @param {string} id Campaign ID
-   * @param {{sendAt: number}} params Schedule params
+   * @param {{scheduledFor: number}} params Schedule params
    *
    * @return {Campaign} Scheduled Campaign
    */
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -2744,14 +2744,14 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    * scheduleCampaign - Schedule a Campaign
    *
    * @param {string} id Campaign ID
-   * @param {{sendAt: number}} params Schedule params
+   * @param {{scheduledFor: number}} params Schedule params
```
gumuz, line 2747: does it have the same meaning as before?

yaziine: yes, a new field is introduced on the API side which reflects this field.

gumuz: isn't it less consistent to use `_for` suffix instead of `_at`, compared to other endpoints?

yaziine: on campaign side we have two fields now:
- scheduled_at: at which time the campaign was scheduled
- scheduled_for: for when the campaign is scheduled (sendAt here)

gumuz (approved):

PR #967, opened 2022-05-11
fix: add TestCampaignResponse type

yaziine:
This PR adds a new type `TestCampaignResponse` to expose either the campaign (on success) or invalid user if there is any.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2241,6 +2241,11 @@ export type Campaign = {
 } & CampaignData &
   CampaignStatus;
 
+export type TestCampaignResponse = {
+  campaign?: Campaign;
+  invalid_users?: Record<string, string>;
+};
+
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -2241,6 +2241,11 @@ export type Campaign = {
 } & CampaignData &
   CampaignStatus;
 
+export type TestCampaignResponse = {
+  campaign?: Campaign;
```
gumuz, line 2245: in which situation would this field be empty?

yaziine: when any user is invalid, the test is not ran

gumuz (approved):

peterdeme (approved):

ferhatelmas (approved):

PR #973, opened 2022-05-25
refactor: add type for general error API response format

MartinCupela:
## CLA

- [X] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [X] Code changes are tested

Add type definitions for error API response payload and for error generated and thrown on error response propagated from the chat client.

This is important for SDKs that use the stream-chat client in order to facilitate work with errors thrown by the client.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -61,5 +61,5 @@ export function isWSFailure(err: APIError): boolean {
 }
 
 export function isErrorResponse(res: AxiosResponse<unknown>): res is AxiosResponse<APIErrorResponse> {
-  return (res.status + '')[0] !== '2';
+  return !res.status || res.status < 200 || 300 <= res.status;
 }
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,3 +1,6 @@
+import { AxiosResponse } from 'axios';
+import { APIErrorResponse } from './types';
+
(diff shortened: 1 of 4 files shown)


src/errors.ts
```diff
@@ -56,3 +59,7 @@ export function isWSFailure(err: APIError): boolean {
     return false;
   }
 }
+
+export function isErrorResponse(res: AxiosResponse<unknown>): res is AxiosResponse<APIErrorResponse> {
+  return (res.status + '')[0] !== '2';
```
peterdeme, line 64: this is very clever but isn't `res.status && res.status > 300` easier to read?

MartinCupela: @peterdeme this was just copied from the original place of use. You are right that it is hacky. I am just wandering, why the author of the one liner included also status 100 among error responses. If I would be translating the original one liner into something more readable, then it would be `!!res.status || res.status < 200 && 300 >= res.status`. Would you be ok with that one?

arnautov-anton (approved):

peterdeme (approved):

PR #1003, opened 2022-07-06
fix: missing app setting types

### LINKED ISSUE #1002: `before_message_send_hook_url` missing from type `AppSettings`
The docs specify that the correct way to enable the before message send webhook is:

```ts
    const client = new StreamChat(apiKey, apiSecret);
    await client.updateAppSettings({
        before_message_send_hook_url: 'http://127.0.0.1:4323',
    });
```

But this generates the following type error:

```
Argument of type '{ before_message_send_hook_url: string; }' is not assignable to parameter of type 'AppSettings'.
  Object literal may only specify known properties, and 'before_message_send_hook_url' does not exist in type 'AppSettings'.ts(2345)
```

Because the [`AppSettings` type](https://github.com/GetStream/stream-chat-js/blob/8ad1303162eec62e52d8526f74d146e9f55f1ecb/src/types.ts#L1456-L1512) in types.ts doesn't list the `before_message_send_hook_url` key.

ferhatelmas:
fixes #1002

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -101,6 +101,7 @@ export type AppSettingsAPIResponse<StreamChatGenerics extends ExtendableGenerics
       }
     >;
     reminders_interval: number;
+    agora_options?: AgoraOptions | null;
     async_url_enrich_enabled?: boolean;
     auto_translation_enabled?: boolean;
     before_message_send_hook_url?: string;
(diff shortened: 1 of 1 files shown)


AnatolyRugalev (approved):

ferhatelmas:
@AnatolyRugalev added a few more

AnatolyRugalev (approved):

PR #1007, opened 2022-07-07
feat!: separate local events and server events

santhoshvai:
## CLA

- [x] Code changes are tested

When supplying event callbacks, this PR helps with stricter typing.

For example,
```
client.on('connection.changed', (event) => console.log(event.online));
// only `online` will be shown by the IDE for code-completion
```
### Breaking change

Clients of the SDK cannot use the fields directly in if statements,
For example,
```
function func(event: Event<StreamChatGenerics>) {
  // this would error out, as TS would complain that event doesn't have online field
  if (event.online) {
    console.log(event.online);
  }
  // the correct way
  if (event.type === 'connection.changed') {
    console.log(event.online)
  }
}
```

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -16,7 +16,7 @@ import {
   DeleteChannelAPIResponse,
   Event,
   EventAPIResponse,
-  EventHandler,
+  EventHandlerInferred,
   EventTypes,
   FormatMessageResponse,
   GetMultipleMessagesAPIResponse,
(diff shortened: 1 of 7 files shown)


src/channel.ts
```diff
@@ -1336,7 +1342,12 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     }
 
     // any event can send over the online count
-    if (event.watcher_count !== undefined) {
+    if (
+      event.type !== 'connection.changed' &&
+      event.type !== 'connection.recovered' &&
+      event.type !== 'transport.changed' &&
```
vishalnarkhede, line 1348: We can extract this into a function `isLocalEvent`

santhoshvai: Great suggestion! Done 👍

vishalnarkhede:
We will pick this up later!

PR #1008, opened 2022-07-12
fix: message update event overrides own reactions

szuperaz:
## CLA

- [x] Code changes are tested

`own_reactions` isn't populated inside WS events, currently if we receive a `message.updated` event the JS client will override the existing `own_reactions` array of the message. This PR fixes that.

## Changelog

-

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -469,13 +469,20 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
 
     // message already exists and not filtered due to timestampChanged, update and return
     if (!timestampChanged && message.id) {
+      let messageIndex;
       if (messageArr[left] && message.id === messageArr[left].id) {
-        messageArr[left] = message;
-        return [...messageArr];
+        messageIndex = left;
       }
 
       if (messageArr[left - 1] && message.id === messageArr[left - 1].id) {
-        messageArr[left - 1] = message;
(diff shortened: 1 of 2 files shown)


src/channel_state.ts
```diff
@@ -469,13 +469,20 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
...
+        messageIndex = left;
       }
 
       if (messageArr[left - 1] && message.id === messageArr[left - 1].id) {
-        messageArr[left - 1] = message;
+        messageIndex = left - 1;
+      }
+
+      if (messageIndex !== undefined) {
+        // own_reactions always be [] in WS events -> ignore for message updates
```
szuperaz, line 482: I'm not happy with this solution, I think the nice solution would be to extend the `event.message` object with `own_reactions` when the `message.updated` [event is received](https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L1250), but that would require an extra iteration of the `channelState.messages` and wasn't sure how performance-critical that method was

szuperaz: Did a second version: https://github.com/GetStream/stream-chat-js/pull/1009 - let me know which one you prefer

AnatolyRugalev: I like the second one better

szuperaz: 👍

PR #1028, opened 2022-08-11
feat: campaign missing pieces

ferhatelmas:
it has some breaking changes but this API had been marked as to be changed.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2695,7 +2695,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    * @return {Segment[]} Segments
    */
   async querySegments(filters: SegmentFilters, options: CampaignQueryOptions = {}) {
-    const { segments } = await this.get<{
+    return await this.get<{
       segments: Segment[];
     }>(this.baseURL + `/segments`, {
       payload: {
@@ -2703,7 +2703,6 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
         ...options,
       },
     });
-    return segments;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -2683,26 +2689,20 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
-   * listSegments - List Campaign Segments
+   * querySegments - Query Campaign Segments
    *
    *
    * @return {Segment[]} Segments
    */
-  async listSegments(options: { limit?: number; offset?: number }) {
-    const { segments } = await this.get<{ segments: Segment[] }>(this.baseURL + `/segments`, options);
-    return segments;
+  async querySegments(filters: SegmentFilters, options: CampaignQueryOptions = {}) {
```
SiddhantAgarwal, line 2697: `CampaignQueryOptions` could be changed to something related to segments options.

src/client.ts
```diff
@@ -2843,6 +2840,27 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     return await this.post<APIResponse & TestCampaignResponse>(this.baseURL + `/campaigns/${id}/test`, { users });
   }
 
+  /**
+   * queryRecipients - Query Campaign Recipient Results
+   *
+   *
+   * @return {Recipient[]} Recipients
+   */
+  async queryRecipients(filters: RecipientFilters, options: CampaignQueryOptions = {}) {
```
SiddhantAgarwal, line 2849: `CampaignQueryOptions` same for query recipients

src/client.ts
ferhatelmas: It's same type but we can provide alias

ferhatelmas: Let me add alias them so that we can diverge if we need later on

SiddhantAgarwal (approved):

PR #1029, opened 2022-08-11
fix: race of query channels and ws connection

vishalnarkhede:
## CLA

- [ ] Code changes are tested

## Changelog

- Fixed the following issue:

```
QueryChannels failed with error: "Watch or Presence requires an active websocket connection, please make sure to include your websocket connection_id"
```

- Fixed handling of multiple client.openConnection calls

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -523,12 +523,11 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       throw Error('User is not set on client, use client.connectUser or client.connectAnonymousUser instead');
     }
 
-    if (this.wsConnection?.isConnecting) {
+    if (this.wsConnection?.isConnecting && this.wsPromise) {
       this.logger('info', 'client:openConnection() - connection already in progress', {
         tags: ['connection', 'client'],
       });
-
-      return Promise.resolve();
+      return this.wsPromise;
     }
 
     if ((this.wsConnection?.isHealthy || this.wsFallback?.isHealthy()) && this._hasConnectionID()) {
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -523,12 +523,11 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
     }
 
-    if (this.wsConnection?.isConnecting) {
+    if (this.wsConnection?.isConnecting && this.wsPromise) {
       this.logger('info', 'client:openConnection() - connection already in progress', {
         tags: ['connection', 'client'],
       });
-
-      return Promise.resolve();
+      return this.wsPromise;
```
vishalnarkhede, line 530: fix: If the connection is in progress, then we need to return the existing promise instead of resolving the promise.

src/client.ts
```diff
@@ -1315,9 +1314,14 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     if (!this.wsConnection && (this.options.warmUp || this.options.enableInsights)) {
       this._sayHi();
     }
-
     // The StableWSConnection handles all the reconnection logic.
-    this.wsConnection = new StableWSConnection<StreamChatGenerics>({ client: this });
+    if (this.options.wsConnection && this.node) {
+      // Intentionally avoiding adding ts generics on wsConnection in options since its only useful for unit test purpose.
+      ((this.options.wsConnection as unknown) as StableWSConnection<StreamChatGenerics>).setClient(this);
+      this.wsConnection = (this.options.wsConnection as unknown) as StableWSConnection<StreamChatGenerics>;
```
vishalnarkhede, line 1321: This is purely to ease unit-testing around connection logic. Currently you can't really properly mock connection class methods since we always create a new instance of `StableWSConnection` class.
Allowing `wsConnection` in options will help with writing more unit tests.

src/connection.ts
```diff
@@ -38,7 +38,7 @@ const isErrorEvent = (res: WebSocket.CloseEvent | WebSocket.Data | WebSocket.Err
  */
 export class StableWSConnection<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics> {
   // global from constructor
-  client: StreamChat<StreamChatGenerics>;
+  client?: StreamChat<StreamChatGenerics>;
```
vishalnarkhede, line 41: Changes in this file are related to https://github.com/GetStream/stream-chat-js/pull/1029/files#r943398458

src/client.ts
```diff
@@ -1466,8 +1470,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     const defaultOptions: ChannelOptions = { state: true, watch: true, presence: false };
 
     // Make sure we wait for the connect promise if there is a pending one
-    await this.setUserPromise;
-
+    await this.wsPromise;
```
vishalnarkhede, line 1473: Wait for actual wsPromise instead of setUserPromise which gets created only during connectUser. But in case of reconnection flow or manual `openConnection` ... setUserPromise would already have resolved resulting in queryChannels call without ws connection.

AnatolyRugalev (approved):

PR #1031, opened 2022-08-17
feat: support limit for load message into state

petyosi:
## CLA

- [x] Code changes are tested

This change allows the SDKs to increase the size of the messages loaded when jumping, thus avoiding multiple requests to the db.

## Changelog
`loadMessageIntoState` accepts an optional `limit` parameter, (default value is 25)
-

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -664,7 +664,7 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
    * @param {string} messageId The id of the message, or 'latest' to indicate switching to the latest messages
    * @param {string} parentMessageId The id of the parent message, if we want load a thread reply
    */
-  async loadMessageIntoState(messageId: string | 'latest', parentMessageId?: string) {
+  async loadMessageIntoState(messageId: string | 'latest', parentMessageId?: string, limit = 25) {
     let messageSetIndex: number;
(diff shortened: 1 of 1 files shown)


src/channel_state.ts
```diff
@@ -664,10 +664,13 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
    * @param {string} messageId The id of the message, or 'latest' to indicate switching to the latest messages
    * @param {string} parentMessageId The id of the parent message, if we want load a thread reply
    */
-  async loadMessageIntoState(messageId: string | 'latest', parentMessageId?: string) {
+  async loadMessageIntoState(messageId: string | 'latest', parentMessageId?: string, limit?: number) {
```
MartinCupela, line 667:
```suggestion
  async loadMessageIntoState(messageId: string | 'latest', parentMessageId?: string, limit = 25) {
```

MartinCupela: Just to avoid the condition:
```
   if (!limit) {
      limit = 25;
    }
```

AnatolyRugalev (approved):

szuperaz (dismissed):
The JSdoc for the `limit` parameter is missing, but the changes look good to me

MartinCupela (dismissed):

Issue #869, opened 2022-01-18
Cannot listen to custom events


andreibarabas:
According to the docs, you can send custom events from the server.

```
 await serverClient.sendUserCustomEvent(user_id, {
        type: 'channel_is_ready',
        abc: 'efg',
      });
```

The problem is though, when you try to listen to those events on the client, you get a "Event not valid error".

```
client.on('channel_is_ready', (event) => {});
```

Caused by a simple check that currently does not support this custom events:
https://github.com/GetStream/stream-chat-js/blob/master/src/events.ts#L49

If we patch the library and add our own custom type to the EVENT_MAP array, it works, so it's a matter of adding support in the sdk

PR #1037, opened 2022-08-19
feat: replaces console.warn instead throwing error when event is not valid

vanGalilea:
## CLA
- [x] Code changes are tested

Motive of this fix is the following issue https://github.com/GetStream/stream-chat-js/issues/869

## Changelog
- replaces console.warn instead throwing an error when event is not valid

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1119,7 +1119,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     const key = callbackOrNothing ? (callbackOrString as string) : 'all';
     const valid = isValidEventType(key);
     if (!valid) {
-      throw Error(`Invalid event type ${key}`);
+      console.warn(`Invalid event type ${key}`);
     }
     const callback = callbackOrNothing ? callbackOrNothing : callbackOrString;
     if (!(key in this.listeners)) {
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -1119,7 +1119,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     const key = callbackOrNothing ? (callbackOrString as string) : 'all';
     const valid = isValidEventType(key);
     if (!valid) {
-      throw Error(`Invalid event type ${key}`);
+      console.warn(`Invalid event type ${key}`);
```
ferhatelmas, line 1122: early return is missing

vishalnarkhede: No, we want to continue executing listeners here. The reason being this `throw error` creates blocker for custom events.

ferhatelmas: It's not executing code path but changing listeners so it will put an invalid key

vishalnarkhede: I am wondering if we should simply remove this console.warn as well.

ferhatelmas: so do you want to put whatever given by user? makes sense to drop then

vishalnarkhede: Yeah!! cc @vanGalilea

vanGalilea: Ready guys!

ferhatelmas (approved):

PR #1039, opened 2022-08-24
feat: async moderation config in app settings

miagilepner:
## CLA

- [ ] Code changes are tested

## Changelog

-


src/types.ts
```diff
@@ -1518,6 +1518,13 @@ export type AppSettings = {
     p12_cert?: string;
     team_id?: string;
   };
+  async_moderation_config?: {
```
ferhatelmas, line 1521: isn't it supposed be in response type too ?

miagilepner: yes it is. fixed now.

ferhatelmas (approved):

PR #1047, opened 2022-10-06
fix: update client to use channel.cid as config keys

arnautov-anton:
## CLA

- [ ] Code changes are tested

Channel configurations can differ from channel to channel so storing configurations under `channel.type` causes overrides for other loaded channels.

Channel `A` of type `messaging` for user `john` with channel role `channel_moderator` has different set of commands in the channel configuration than the channel `B` of type `messaging` where `john` isn't a `channel_moderator`. As channel `B` is after channel `A` in the list and is of the same type, the configuration gets overriden and `john` isn't able to see available commands for channel `A`.

I find using `channel.cid` as the most appropriate store key as the configuration should be bound to user/channel. I might be missing something so I'd like you guys to share your opinions (especially RN team).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -179,9 +179,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   clientID?: string;
   configs: Configs<StreamChatGenerics>;
   key: string;
-  listeners: {
-    [key: string]: Array<(event: Event<StreamChatGenerics>) => void>;
-  };
+  listeners: Record<string, Array<(event: Event<StreamChatGenerics>) => void>>;
   logger: Logger;
   /**
    * When network is recovered, we re-query the active channels on client. But in single query, you can recover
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -1196,7 +1194,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     }
 
     if (event.channel && event.type === 'notification.message_new') {
-      this.configs[event.channel.type] = event.channel.config;
+      this._addChannelConfig({ channel: event.channel });
```
arnautov-anton, line 1197:
```suggestion
      // possible @ts-ignore here instead?
      this._addChannelConfig(event);
```
I had to pass a new object as TS somehow doesn't respect `event.channel` nullish check of the condition above and thinks that `event.channel` could be `undefined`.

madsroskar: It would probably be okay with it if the check is `event.channel !== undefined` and not the vague truthy check that it does now, right? Or does that give the same result?

madsroskar: It seems a bit odd to me to pass the channel to a function that by name expects a channel _config_, but that's not that big a deal

madsroskar (commented):
I think this should be fine for us, I don't foresee it breaking anything for RN

vanGalilea (approved):

madsroskar (approved):

Issue #992, opened 2022-06-18
TypeError: Cannot read properties of undefined (reading 'bind')


nayebare:
Hi, i need someone to kindly look into this error.

```
let  StreamChat = require("stream-chat").StreamChat;

../node_modules/stream-chat/dist/index.js:1284
    return crypto.getRandomValues.bind(crypto);
                                  ^

TypeError: Cannot read properties of undefined (reading 'bind')
    at ../node_modules/stream-chat/dist/index.js:1284:35
    at Object.<anonymous> (../node_modules/stream-chat/dist/index.js:1290:2)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/index.js:6:19)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
```

madsroskar:
Hi, @nayebare. I'm not able to reproduce this, would you be able to provide a minimal example that reproduces the issue? Any more information you could give about the specific environment would help as well. For example, which version of `stream-chat` are you using? Which version of NodeJS are you using?

nayebare:
@madsroskar kindly replicate the error using this code the goal is to run the code as serverless while testing it locally

vanGalilea:
@nayebare does this issue still occur? 
If yes we would like to pinpoint this issue as fast as possible, we would like to ask you to share a minimalistic repo where the problem is reproduced.
The fastest way to achieve this is with the following command to create a new RN project with Stream chat in it:

`npx react-native init MyApp --template stream-chat-react-native-template`

Please then share a Github repo with us, where we can explore the issue ourselves.

nayebare:
@vanGalilea kindly replicate the error using this code the goal is to run the code as serverless while testing it locally https://github.com/nayebare/create-getstream-node-test

khushal87:
Hey @nayebare the code you have shared takes us to a 404 not found page. Still facing this issue?

nayebare:
hey @khushal87 i have made the repository thank you

vanGalilea:
[...]

PR #1053, opened 2022-10-11
fix: crypto.getRandomValues check for older node versions and min nod…

vanGalilea:
## CLA

- [ ] Code changes are tested

This PR fixes issue #992 by checking weather `crypto.getRandomValues` exists and fallbacks to our own implementation of generating a uuid.  `crypto.getRandomValues` is being supported from [node 17.4.0 onwards](https://nodejs.org/api/crypto.html#cryptogetrandomvaluestypedarray)

## Changelog

- Targeting node v16 and higher to prevent future mishaps (APIs that are partially supported) similar to this. We were a bit behind with our engine node version (v14, while v18 will become LTS soon).


test/unit/utils.js
```diff
@@ -1,8 +1,25 @@
...
+
+	it('generates a UUID manually when crypto is unavailable', () => {
+		sinon.spy(Math, 'pow');
+		sinon.spy(Math, 'random');
+		const uuid = generateUUIDv4();
+		expect(uuid).to.be.a('string');
+		expect(uuid.length).to.equal(36);
+		expect(Math.pow.calledWithMatch(2, 8)).to.be.true;
+		expect(Math.random.callCount).to.be.equal(16);
+	});
```
vanGalilea, line 20: Globally mocking crypto was a tough chore, due to chai/mocha's logic. (With Jest it would've been a different story, way easier :|)
I couldn't figure it out at the moment. Therefore we test the fallback scenario only

vishalnarkhede (approved):

PR #1074, opened 2022-11-30
feat: add batch method for deactivate and reactivate users

yaziine:


src/client.ts
```diff
@@ -1961,7 +1962,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    *
    * @param {string[]} user_ids which users to restore
    *
-   * @return {APIResponse} A task ID
+   * @return {APIResponse} An API response
```
vishalnarkhede, line 1965:
```suggestion
   * @return {APIResponse & { user: UserResponse<StreamChatGenerics> }} An API response
```

vishalnarkhede (dismissed):

yaziine:
@ferhatelmas @vishalnarkhede are we good here?

src/types.ts
```diff
@@ -884,6 +889,17 @@ export type QueryMembersOptions = {
   user_id_lte?: string;
 };
 
+export type ReactivateUserOptions = {
+  created_by_id?: string;
+  name?: string;
```
ferhatelmas, line 894: we don't support name in batch version?

src/client.ts
```diff
@@ -1961,35 +1964,68 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
 
+  /**
+   * deactivateUsers - Deactivate many users asynchronously
+   *
+   * @param {string[]} user_ids which users to deactivate
+   * @param {DeactivateUsersOptions} [options]
+   *
+   * @return {TaskResponse} A task ID
+   */
+  async deactivateUsers(user_ids: string[], options?: DeactivateUsersOptions) {
```
ferhatelmas, line 2025: nit: to align with others, might want to rename it to `userIDs` and internally do `user_ids: userIDs`

src/types.ts
yaziine: we don't, might be easier to handle batches where there is no conflict, and handle conflict using the simple endpoint. I don't think it will be a common scenario. Can be upgraded later if needed.

src/client.ts
yaziine: if you look at other endpoints, we use `user_ids`

src/types.ts
ferhatelmas: agreed

src/client.ts
ferhatelmas: my point is difference between single vs batch ones

ferhatelmas (approved):

src/client.ts
yaziine: yeah but since others use `user_ids`, i don't see the real value honestly

ferhatelmas: value is clear, mixing casing is unprofessiona and not user intuitive but should be fixed globally

PR #1115, opened 2023-04-17
Khushal87/new message channel hidden

khushal87:
## CLA

- [x] Code changes are tested

According to this documentation, the channel should not stay hidden if a new message is received or `message.new` event is triggered. Without this, the channel is still shown in the channel list even if the filter matches the `channel.data.hidden` always remains as true. This needs to be updated when a new message is received. Hence this change is introduced.

In React Native, even if `channelRenderFilterFn` is used, the hidden remains true, and in an environment of a two-channel (one with filter as hidden true and the other with false) list, both channels receive the update, which seems buggy.

## Changelog

-

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1385,14 +1385,6 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
           channelState.clearMessages();
         }
         break;
-      case 'channel.visible':
-        if (event.channel) {
-          channel.data = {
-            ...event.channel,
-            hidden: false,
-          };
-        }
-        break;
       default:
     }
 
@@ -1436,7 +1428,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
   _checkInitialized() {
     if (!this.initialized && !this.offlineMode && !this.getClient()._isUsingServerAuth()) {
       throw Error(
-        `Channel ${this.cid} hasn't been initialized yet. Make sure to call .watch() and wait for it to resolve ${this.initialized}, ${this.offlineMode}`,
+        `Channel ${this.cid} hasn't been initialized yet. Make sure to call .watch() and wait for it to resolve`,
       );
     }
   }
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1276,6 +1276,8 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
           const ownMessage = event.user?.id === this.getClient().user?.id;
           const isThreadMessage = event.message.parent_id && !event.message.show_in_channel;
 
+          channel.data = { ...channel.data, hidden: false };
+
           if (this.state.isUpToDate || isThreadMessage) {
             channelState.addMessageSorted(event.message, ownMessage);
           }
@@ -1383,6 +1385,14 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
           channelState.clearMessages();
         }
         break;
+      case 'channel.visible':
+        if (event.channel) {
+          channel.data = {
+            ...event.channel,
+            hidden: false,
+          };
+        }
+        break;
       default:
     }
 
@@ -1426,7 +1436,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
   _checkInitialized() {
     if (!this.initialized && !this.offlineMode && !this.getClient()._isUsingServerAuth()) {
       throw Error(
-        `Channel ${this.cid} hasn't been initialized yet. Make sure to call .watch() and wait for it to resolve`,
+        `Channel ${this.cid} hasn't been initialized yet. Make sure to call .watch() and wait for it to resolve ${this.initialized}, ${this.offlineMode}`,
       );
     }
   }


src/channel.ts
```diff
@@ -1279,6 +1279,15 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
           const ownMessage = event.user?.id === this.getClient().user?.id;
           const isThreadMessage = event.message.parent_id && !event.message.show_in_channel;
 
+          if (channel.data?.hidden) {
```
MartinCupela, line 1282: Shouldn't we be rather calling `channel.show()` to be in sync with the back-end? Also the documentation mentions this method to show the channel again.

image

khushal87: I tried that, but this would need significant refactoring on the client side since the method is async and returns a promise, and most of the code here is sync. Also, we didn't handle `channel.visible` before; that was a big issue too.

khushal87: My wild guess is the backend calls the `channel.hide()` API internally when a new message is received for a hidden channel.

MartinCupela: @khushal87 you think the backend calls `channel.hide()` on already hidden channel when a new message arrives?

khushal87: Not this exactly, but they locally update the hidden entity(if they have it in the DB) when the new message is received or execute the controller of the hidden logic.

MartinCupela: I tested the following workflow without the changes included in this PR:

1. A hides channel Boulder
2. B sends message to Boulder
3. Channel Boulder appears in the A's channel list again

@khushal87 Is this what you are trying to achieve by this PR?

https://ctrlv.tv/C0fw

khushal87: Yes @MartinCupela

khushal87: Lets say there are two channel list one with filter hidden true and the other with hidden false. If boulder was in channel list with filter hidden false and was hidden it shall move to channel with filter hidden true and once message.new is triggered, the channel should move back to channel list with hidden false.

MartinCupela: That is exactly what is happening on the video

khushal87: Oh sorry, I couldn't play the video for some reason. So do you think this PR can solve the issue?

MartinCupela: @khushal87 the video demo is not using your fix. So my concern is, whether this fix is actually needed.

khushal87: Can you show me the `message.new` event listener in the React SDK? Maybe you handle this case explicitly. Even we do it in RN SDK, and adding this change removes [the following lines](https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/ChannelList/hooks/listeners/useNewMessage.ts#L48).

vishalnarkhede: As discussed, I think instead of dispatching this event (which is wrong since its a backend event) ... we should just switch the hidden property to false here.

khushal87: Cool, I have removed the dispatch, but it will be great for backend to emit the `channel.visible` event, which I think is not emitted at all currently.

MartinCupela: @khushal87 I have pushed the change that  instead of dispatching the `channel.visible` event just switches the hidden property to false. As @vishalnarkhede proposed.

src/channel.ts
```diff
@@ -1084,7 +1084,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
 
   /**
    * removes the hidden status for a channel
-   *
+   * Has to be called before Channel.sendMessage() in order WS event `channel.visible` is emitted.
```
szuperaz, line 1087: I don't understand this comment, backend will automatically mark a channel as visible when we send a new message. This might be required some SDK workaround, but in that case that should live in an SDK specific documentation.

szuperaz:
I'm not sure what's the problem we want to solve, but the main issue seems to be that the `hidden` flag is controlled by the backend, but the client isn't always notified about changes to that flag. 

In any case, updating the `hidden` flag locally on `channel.visible` event seems to be a safe change, I'm not sure it will solve all problems around the `hidden` flag, but if it solves a specific problem we can merge this PR.

src/channel.ts
khushal87: I am also in favour of it and talked about it with @MartinCupela. The BE team cannot pick this up right now because of the lack of resources I think.

khushal87: This whole thing should ideally happen on the backend.

MartinCupela: @szuperaz Backend marks the channel as visible with a new message, but does not send the channel.visible event unless you explicitly call `channel.show()` before `channel.sendMessage()`. If we want to keep it in the SDK docs, then all the JS SDKs should add this not into the documentation. I can remove the comment. Please confirm, if you insist it should not be mentioned here.

src/channel.ts
```diff
@@ -1387,6 +1387,9 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
           channelState.clearMessages();
         }
         break;
+      case 'channel.visible':
+        channel.data = { ...channel.data, hidden: false };
```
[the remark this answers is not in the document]
MartinCupela: @szuperaz true, have fixed that.

src/channel.ts
szuperaz: It doesn't relevant to all JS SDKs, for example Angular will show the channel even without calling `channel.show()`, so I think it's misleading to have it in the client. Also,the comment doesn't mention what is the problem we solve with calling `channel.show()` so I don't really think an integrator would understand the comment on their own anyway, I'm also not sure how would they find that comment in the code

MartinCupela: @szuperaz I have removed the doc string

szuperaz: Thank you

szuperaz (approved):

santhoshvai (approved):

vanGalilea (approved):

PR #1116, opened 2023-04-20
feat: do not delete the active channel if offlineMode is true

khushal87:
## CLA

- [x] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1572,17 +1572,6 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       channels.push(c);
     }
 
-    if (!offlineMode) {
-      // If the channels are coming from server, then clear out the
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -1572,17 +1572,6 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
-      // If the channels are coming from server, then clear out the
-      // previously help offline channels.
-      for (const key in this.activeChannels) {
-        const channel = this.activeChannels[key];
-        if (channel.offlineMode) {
-          delete this.activeChannels[key];
-        }
-      }
-    }
-
```
santhoshvai, line 1585: It seems like the removal of this part is unnecessary.. because `channel.offlineMode` should be false when a successful BE query happens.

PR #1138, opened 2023-06-20
fix: add missing properties to ExportChannelStatusResponse

rchl:
## CLA

- [x] Code changes are tested

Those properties were missing in the type returned from `getExportChannelStatus` API.

Some references:
 - Node.js API - https://getstream.io/chat/docs/node/exporting_channels/?language=javascript#requesting-a-channel-export
 - REST API - https://getstream.io/chat/docs/rest/#channels-getexportchannelsstatus

Unfortunately there is no documentation as to what exact values the `status` can have so it's just typed as `string`.

The `error` and `result` objects could also have better types but apart from REST documentation having some info on those, it's not 100% clear on which properties are optional.

EDIT: Added types for `result` also, making best effort with available information.

## Changelog

- fix: add missing properties to ExportChannelStatusResponse


rchl:
Also created `ExportChannelsResult` type to improve slightly more.

src/types.ts
```diff
@@ -381,10 +381,18 @@ export type ExportUsersResponse = {
...
+};
+
 export type ExportChannelStatusResponse = {
+  status: string;
+  task_id: string;
   created_at?: string;
-  error?: {};
-  result?: {};
+  error?: ExportChannelsResult | null;
+  result?: {} | null;
```
santhoshvai, line 395: `result` should use `ExportChannelsResult`

and `error` should use `ErrorResult`

where `ErrorResult` is

```
type?: string;
description?: string;
stacktrace?: string;
version?: string;
```

rchl: Thanks. That was embarrassing :)

Applied your suggestion with the difference that I've called the error type `ExportChannelError` since I thought `ErrorResult` would be too generic. Can change if you actually want it to be that generic.

santhoshvai: Thanks, also I think `null` is unnecessary here . `ExportChannelError` sounds good to me 👍

santhoshvai: and we would need to make it

```
 status?: string;
 task_id?: string;
```

rchl: Don't fully understand. Add those to the `ExportChannelError` or change required to optional in `ExportChannelStatusResponse`?

> and we would need to make it
> 
> ```
>  status?: string;
>  task_id?: string;
> ```

santhoshvai: Sorry, to be lear, what I meant is that all the properties in `ExportChannelStatusResponse` must be made optional, like below

```
export type ExportChannelStatusResponse = {
  status?: string;
  task_id?: string;
  created_at?: string;
  error?: ExportChannelError;
  result?: ExportChannelsResult | null;
  updated_at?: string;
};
```

santhoshvai: can you also remove the null here please 

```
result?: ExportChannelsResult | null;
```

rchl:
```
Status {
  task_id: 'c1e57bb1-04f1-4284-af61-455c3990b631',
  status: 'waiting',
  created_at: '2023-11-30T12:42:46.266232Z',
  updated_at: '2023-11-30T12:42:46.266232Z',
  duration: '16.19ms',
  result: null
}
Status {
  task_id: 'c1e57bb1-04f1-4284-af61-455c3990b631',
  status: 'pending',
  created_at: '2023-11-30T12:42:46.266232Z',
  updated_at: '2023-11-30T12:42:46.501436Z',
  duration: '3.45ms',
  result: null
}
Status {
  task_id: 'c1e57bb1-04f1-4284-af61-455c3990b631',
  status: 'completed',
  created_at: '2023-11-30T12:42:46.266232Z',
  updated_at: '2023-11-30T12:42:46.689775Z',
  duration: '2.00ms',
  result: {
    url: 'https://...'
  }
}
```

rchl: BTW. I wonder what makes you think that `task_id` and `status` are optional. Did you verify internally that it can be the case that those are missing? I would think that the response would be pretty useless without those.

rchl: We're closing on a year here.
If you don't care then I can just close it.

PR #1165, opened 2023-09-04
feat: make unblockMessage method public

gumuz:
made _unblockMessage method public and kept alias for backwards compatibility

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2248,23 +2248,21 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   }
 
   /**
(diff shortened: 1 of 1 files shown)


khushal87:
Implementation wise this looks good. Can you brief me about the purpose behind the change? Thanks 😄

gumuz:
thanks @khushal87 this was a method that was previously used only internally by the dashboard that we now want to expose to the user. Hence the rename and dropping of the @public marker, I've left the _unblockMessage alias there for backwards compatibility.

khushal87 (approved):

PR #1179, opened 2023-09-29
fix: queue channel WS events until the channel is initialized

MartinCupela:
## CLA

- [ X ] Code changes are tested

A channel is flagged as `initialized` only after `channel.query` call successfully returns inside `channel.watch()` method. The WS events for a given channel can however arrive sooner to the client than `channel.query` call returns. Processing WS events before the `channel.query` call returns leads to checking whether the channel has been initialized. If the channel was not flagged as initialized, then the code throws error.

To prevent processing WS events before the initialization has finished, this PR introduces a WS event queue for a channel. There will all the event objects be gathered until the channel is flagged as initialized. Once flagged as initialized, the queue is flushed.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1576,7 +1576,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
 
   _flushWsEventQueue() {
     while (this.wsEventQueue.length) {
-      const event = this.wsEventQueue.pop();
+      const event = this.wsEventQueue.shift();
       if (event) this.getClient().dispatchEvent(event);
     }
   }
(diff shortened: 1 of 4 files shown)


src/channel.ts
```diff
@@ -1555,4 +1573,11 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     this.disconnected = true;
     this.state.setIsUpToDate(false);
   }
+
+  _flushWsEventQueue() {
+    while (this.wsEventQueue.length) {
+      const event = this.wsEventQueue.pop();
```
vishalnarkhede, line 1579: I think FIFO would make more sense than LIFO.

vishalnarkhede: `unshift` is expensive compared to `pop`, but I can't imagine this queue being huge.

vishalnarkhede (approved):

oliverlaz (approved):

PR #1183, opened 2023-10-05
fix: evaluate channel.lastRead when channel is not initialized

MartinCupela:

Currently `channel.lastRead()` throws an error if a channel is uninitialized.  Call to `channel.countUnreadMentions()` would throw an error in such case too. This complicates the work in an UI with uninitialized `Channel` instance which is a valid use-case. This PR proposes to return valid values like 0 for unread mentions count and `null` for last read data and not to break the UI.

### DIFF
--- a/test/unit/channel.js
+++ b/test/unit/channel.js
@@ -592,7 +592,7 @@ describe('Uninitialized Channel', () => {
 	});
 
 	it('returns 0 mentions in unread messages', () => {
-		expect(channel.countUnreadMentions()).to.eq(0);
+		expect(channel.countUnreadMentions()).to.eq(-1);
 	});
 
 	it('reports no lastRead data', () => {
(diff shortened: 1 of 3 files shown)


test/unit/channel.js
```diff
@@ -569,6 +576,30 @@ describe('Channel _handleChannelEvent', function () {
...
+		client.user = user;
+		client.userID = user.id;
+		client.userMuteStatus = (targetId) => targetId.startsWith('mute');
+		channel = client.channel('messaging', 'id');
+		channel.initialized = false;
+		channel.offlineMode = false;
+	});
+
+	it('returns 0 mentions in unread messages', () => {
+		expect(channel.countUnreadMentions()).to.eq(0);
```
oliverlaz, line 595: out of curiosity, isn't it better to return `-1` when the channel is not yet initialized? 

`0` can be easily confused with:
- I have no unread mentions
- My channel isn't initialized

MartinCupela: Yes, it gives more granularity to decision making. Also, it should not break any UI as until now, in this case the error would be thrown.

oliverlaz (dismissed):

oliverlaz (approved):

oliverlaz (approved):

vishalnarkhede (approved):

PR #1196, opened 2023-11-14
fix: reload channel state if frozen flag changed

szuperaz:
## CLA

- [x] Code changes are tested

Problem: when the `frozen` flag changes on channels, the backend updates the user's capabilities to reflect the channel's frozen state. However, the backend doesn't send the updated `own_capabilities` via WS (due to performance reasons), so the client has to reload the channel state to update `own_capabilities`. 

To help SDKs detect when a state update is necessary, the channel now emits `capabilities.changed` event.

## Changelog

- fix: reload channel state if frozen flag changed


oliverlaz (approved):

oliverlaz (approved):

src/channel.ts
```diff
@@ -1351,6 +1362,10 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
         break;
       case 'channel.updated':
         if (event.channel) {
+          const isFrozenChanged = event.channel?.frozen !== undefined && event.channel.frozen !== channel.data?.frozen;
+          if (isFrozenChanged) {
```
vishalnarkhede, line 1366:
```suggestion
          const isFrozen = event.channel?.frozen !== undefined && event.channel.frozen !== channel.data?.frozen;
          if (isFrozen) {
```

Since we only query when channel is frozen, not the other way round.

szuperaz: We need to query if frozen changes from `false` -> `true`, but also if it changed from `true` -> `false`. So we want to detect state changes for the `frozen` flag, so I think the name is correct here.

vishalnarkhede: Should messages limit be 0? It will just wipe the messages from UI. Is that expected?

szuperaz: It doesn't wipe messages, at least not from `channel.state`, it only adds new messages from the query result: https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L1474

szuperaz: messages will be an empty array in the `channels.queried` event's payload: https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L1035 - but not in `channel.state`

szuperaz: The reason `limit: 0` is set is to avoid unnecessarily loading the same messages that are already in the state, I've only tested with Angular SDK, but I didn't spot any issue there, let me know if you see it otherwise.

szuperaz: I went though the `_initializeState` and this is the only wiping out possibility I've spotted:https://github.com/GetStream/stream-chat-js/pull/1196/commits/63fdfabcd1f071bd17e619bf381e3abf2961d9f6

vishalnarkhede (approved):

src/channel.ts
```diff
@@ -1480,7 +1495,9 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     if (state.pending_messages) {
       this.state.pending_messages = state.pending_messages;
     }
-    this.state.watcher_count = state.watcher_count || 0;
+    if (state.watcher_count !== undefined) {
+      this.state.watcher_count = state.watcher_count || 0;
```
MartinCupela, line 1499:
```suggestion
      this.state.watcher_count = state.watcher_count;
```

test/unit/channel.js
```diff
@@ -478,6 +481,82 @@ describe('Channel _handleChannelEvent', function () {
 		expect(channel.data.hidden).eq(true);
 	});
 
+	it('should update the frozen flag and reload channel state to update `own_capabilities`', () => {
+		const event = {
+			channel: { frozen: true },
+			type: 'channel.updated',
+		};
+		channel.data.frozen = false;
+		sinon.spy(channel, 'query');
```
MartinCupela, line 490: @szuperaz shouldn't we restore the spy? I could not find any restore call here.

MartinCupela (approved):

test/unit/channel.js
szuperaz: I've checked and the spy only lives in the scope of the given `it` function, so it doesn't affect other tests.

PR #1204, opened 2023-12-19
feat: threads 2.0 

vishalnarkhede:
## CLA

- [ ] Code changes are tested

So we are introducing threads 2.0, which will help SDKs build "Threads" section like slack.

### Basic Usage

```ts
const client = StreamChat.getInstance('api_key');
await client.connectUser({ id: 'user_id' }, 'user_token');
await threads = await client.queryThreads({ watch: true, limit: 20, participant_limit: 5 });

for (const thread of threads) {
  console.log(thread.participants);
  console.log(thread.latestReplies);
  console.log(thread.parentMessage.text)
}
```

[QA Tests](https://github.com/GetStream/chat/blob/threads/qa/tests/thread_list.ts) for thread list might be helpful to understand the usage as well

### Basic usage of queryThreads in sdk:

- `client.queryThreads` endpoint returns list of threads sorted in descending order of latest reply date time.
- Its important to note that you can provide `watch` param to this endpoint. This will subscribe the user to corresponding channels of the threads that are returned. So channel may not exist in ChannelList but user will be watching this channel due to threads endpoint. 

```tsx
const [threads, setThreads] = useState([]);
const [nextCursor, setNextCursor] = useState(undefined);

const loadThreads = async () => {
  const { next, threads } = await client.queryThreads();
  setThreads(threads);
  setNextCursor(next);
}

const loadNextPage = async () => {
  const { next, threads: newThreads } = await client.queryThreads({ next: nextCursor });
  setThreads((threads) => [...threads, ...newThreads]);
  setNextCursor(next);
};

// Render the threads in UI
```

### Handling a new reply on thread

Only participants of the thread receive an event whenever there is a new reply on the thread - `notification.thread_message_new`

```js
client.on('notification.thread_message_new', (event) => {
  const message = event.message;
  if (!message?.parent_id) return threads;

  const parentInThreadsIdx = threads.findIndex((thread) => thread.id === message.parent_id);
  // If the parent message is not in threads, then don't do anything.
  // Slack doesn't append a new thread dynamically to the list, only shows a notification dot.
  // But one can use `client.getThread()` to fetch the new thread.
  if (parentInThreadsIdx === -1) {
    return;
  }

  setThreads((threads) => {
    const newThreads = [...threads];
    newThreads[parentInThreadsIdx].addReply(message);
    return newThreads;
  });
});

```

### Handling message updated on thread

Updates to messages should be handled as usual via `message.updated` and `message.deleted` event:

```js
client.on('message.updated', handleEvent);
client.on('message.deleted', handleEvent);
client.on('reaction.new', handleEvent);
client.on('reaction.deleted', handleEvent);

const handleEvent = (event) => {
  const message = event.message;
  if (message.parent_id) {
    // check if one of the thread replies was deleted or updated
    const thread = threads.find((thread) => thread.message.id === message.parent_id);
    if (!thread) return;
    thread.updateReply(message);
  } else {
    // check if one of the thread parent messages was deleted or updated
    const thread = threads.find((thread) => thread.message.id === message.id);
    if (!thread) return;
    thread.message = formatMessage(message);
  }

  setThreads([...threads]);
};

```

## Changelog

-


tbarbugli (approved):

oliverlaz (dismissed):

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+    const opts = {
+      limit: 10,
+      participant_limit: 10,
+      reply_limit: 3,
+      watch: true,
+      ...options,
+    };
+
+    const res = await this.post<QueryThreadsAPIResponse<StreamChatGenerics>>(this.baseURL + `/threads`, opts);
+    const threads: Thread<StreamChatGenerics>[] = [];
```
MartinCupela, line 2620: Could the following lines be simplified a bit?
```suggestion
    return {
      threads: res.threads.map((thread) => new Thread(this. thread)),
      next: res.next,
    };
```

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     );
   }
 
+  async queryThreads(options?: {
+    limit?: number;
```
MartinCupela, line 2603: Could you please document the `options`?

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+
+    return {
+      threads,
+      next: res.next,
+    };
+  }
+
+  async getThread(
+    messageId: string,
+    options: { participant_limit?: number; reply_limit?: number; watch?: boolean } = {},
```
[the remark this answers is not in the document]
MartinCupela: Could these be documented please?

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+      ...options,
+    };
+
+    const res = await this.get<GetThreadAPIResponse<StreamChatGenerics>>(this.baseURL + `/threads/${messageId}`, opts);
+
+    return new Thread<StreamChatGenerics>(this, res.thread);
+  }
+
+  async partialUpdateThread(messageId: string, partialThreadObject: PartialThreadUpdate) {
+    if (!messageId) {
```
MartinCupela, line 2650: Shouldn't the same error be thrown in case of `getThread()` method?

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+
+    for (const key in { ...partialThreadObject.set, ...partialThreadObject.unset }) {
+      if (reservedThreadFields.includes(key)) {
+        throw Error(
+          `You cannot set ${key} field. ${key} is reserved for server-side use. Please omit ${key} from your set object.`,
+        );
+      }
+    }
+
+    return await this.patch<GetThreadAPIResponse<StreamChatGenerics>>(this.baseURL + `/threads/${messageId}`, {
```
MartinCupela, line 2676:
```suggestion
    return await this.patch<GetThreadAPIResponse<StreamChatGenerics>>(this.baseURL + `/threads/${messageId}`, partialThreadObject)
```

src/thread.ts
```diff
@@ -0,0 +1,115 @@
...
+  MessageResponse,
+  ThreadResponse,
+  ChannelResponse,
+  FormatMessageResponse,
+  ReactionResponse,
+  UserResponse,
+} from './types';
+import { addToMessageList, formatMessage } from './utils';
+
+type ThreadReadStatus<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics> = Record<
```
MartinCupela, line 14: Where is the thread read status updated / based on what WS event?

src/thread.ts
```diff
@@ -0,0 +1,115 @@
...
+    last_read_message_id: string;
+    unread_messages: number;
+    user: UserResponse<StreamChatGenerics>;
+  }
+>;
+export class Thread<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics> {
+  id: string;
+  latestReplies: FormatMessageResponse<StreamChatGenerics>[] = [];
+  participants: ThreadResponse['thread_participants'] = [];
+  message: FormatMessageResponse<StreamChatGenerics>;
```
MartinCupela, line 27: Is this the thread parent message?

src/client.ts
```diff
@@ -2595,6 +2599,85 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     );
   }
 
+  async queryThreads(options?: {
```
MartinCupela, line 2602: Is there a plan to update client.activeChannels[key].state.threads with the latest date with this query?

src/thread.ts
vishalnarkhede: yes

src/thread.ts
vishalnarkhede: hmm right now I am delegating this part to SDK implementation. But I will think about it for 2nd round, maybe we can handle it by default within js client.

src/client.ts
vishalnarkhede: latest date?

src/thread.ts
MartinCupela: don't we want to call it parentMessage as the thread reply contains parent_id?

vishalnarkhede: I actually called it `parentMessage` before, but somehow felt it easier to just call it a message. I mean it is assumed that thread will have 1 message and multiple replies.

But feel free to change it later during sdk integration if everyone agrees on it.

src/client.ts
vishalnarkhede: Good point. If needed we can, didn't feel any need at the moment.

vishalnarkhede:
@MartinCupela I am going to merge this one to unblock backend PR. But feel free to suggest more changes, happy to do them on a separate PR.

MartinCupela:
> @MartinCupela I have made the suggested changes and I am going to merge this one to unblock backend PR. But feel free to suggest more changes, happy to do them on a separate PR. During SDK integration we will need probably bunch new methods and logic for event handlign.

In general I think we will need to 

1. solve the merging of the data queried from the API to the client state
2. WS event handling (notification.thread_message_new vs. message.new, message.deleted, message.updated,  
3. thread read state notification from BE and updates

PR #1206, opened 2024-01-03
fix: deleteUsers - add `pruning` to options enums

arnautov-anton:

Adjust `deleteUsers` to match [the documentation](https://getstream.io/chat/docs/react/update_users/?language=javascript&q=pruning#deleting-many-users);
- adds missing `pruning` option to enums
- adjusts conditions
- makes `options` truly optional
- adds tests

### DIFF
--- a/test/unit/client.js
+++ b/test/unit/client.js
@@ -328,6 +328,49 @@ describe('Detect node environment', () => {
 	});
 });
 
+describe('Client deleteUsers', () => {
+	it('should allow completely optional options', async () => {
+		const client = await getClientWithUser();
+
+		client.post = () => Promise.resolve();
+
+		await expect(client.deleteUsers(['_'])).to.eventually.equal();
+	});
+
+	it('delete types - options.conversations', async () => {
+		const client = await getClientWithUser();
+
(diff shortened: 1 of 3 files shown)


oliverlaz (dismissed):

test/unit/client.js
```diff
@@ -328,6 +328,48 @@ describe('Detect node environment', () => {
...
+		await expect(client.deleteUsers(['_'])).to.eventually.equal();
+	});
+
+	it('delete types - options.conversations', async () => {
+		const client = await getClientWithUser();
+
+		client.post = () => Promise.resolve();
+
+		await expect(client.deleteUsers(['_'], { conversations: 'hard' })).to.eventually.equal();
+		await expect(client.deleteUsers(['_'], { conversations: 'soft' })).to.eventually.equal();
```
MartinCupela, line 346: Shouldn't we test for option `'pruning'` too?

arnautov-anton: That is not supported by `conversations` option but I can add it as an extra case just in case? 🤔

oliverlaz (approved):

MartinCupela (approved):

PR #1210, opened 2024-01-16
fix: prevent channel unread count reset to 0 when sending message and on new own or thread messages

MartinCupela:

A user's unread count for a channel is not reset on the back-end when:
1. the user sends a new message
2. a thread reply is received
3. message.new event for own message is received

Therefore neither the client should reset the unread count to 0. Keeping the logic that resets the count to 0 in the above cases would mean that if I channel is marked unread and one of the above actions happen (we receive a thread reply), then the channel unread count would locally be reset to 0, but on the BE it would be kept non-zero.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -170,18 +170,10 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
    * @return {Promise<SendMessageAPIResponse<StreamChatGenerics>>} The Server Response
    */
   async sendMessage(message: Message<StreamChatGenerics>, options?: SendMessageOptions) {
-    const sendMessageResponse = await this.getClient().post<SendMessageAPIResponse<StreamChatGenerics>>(
-      this._channelURL() + '/message',
-      {
-        message,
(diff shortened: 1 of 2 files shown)


szuperaz (commented):
I've tested with the Angular SDK - everything is ok there. However, I don't think the old behavior caused any issues - probably because the SDK does a lot of `markRead` calls which will reset the unread count on the backend as well. Do I understand it correctly that these problems surfaced in connection of the mark unread feature?

MartinCupela:
> I've tested with the Angular SDK - everything is ok there. However, I don't think the old behavior caused any issues - probably because the SDK does a lot of `markRead` calls which will reset the unread count on the backend as well. Do I understand it correctly that these problems surfaced in connection of the mark unread feature?

Yes @szuperaz that is correct. Imagine you mark some messages unread and you want to quickly send a message to that given channel. A side effect would be, that you see 0 unread messages, but the BE would keep unread count non-zero.

szuperaz (approved):

PR #1211, opened 2024-01-16
feat: register first unread message id on message.new & when initializing Channel state

MartinCupela:

To achieve consistency in the UI concerning the display of unread messages indicators, we need to provide the `first_unread_message_id` value consistently. The value is not provided when querying channels, neither when a new message arrives. Therefore, we have to calculate it client-side.
In this PR we register the `first_unread_message_id` in `message.new` handler and in `Channel._initializeState()`. The latter is invoked every time a `Channel` instance is created or channel data is queried from the BE. In `Channel._initializeState()` we inspect the whole message set and not only the recently queried messages.

This PR also removes dead code, that was anyways overwritten on the following lines.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1270,8 +1270,9 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
         break;
       case 'message.new':
         if (event.message) {
+          const ownUserId = this.getClient().user?.id;
           /* if message belongs to current user, always assume timestamp is changed to filter it out and add again to avoid duplication */
-          const ownMessage = event.user?.id === this.getClient().user?.id;
+          const ownMessage = event.user?.id === ownUserId;
           const isThreadMessage = event.message.parent_id && !event.message.show_in_channel;
 
           if (this.state.isUpToDate || isThreadMessage) {
@@ -1297,6 +1298,9 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
                 };
               } else {
                 channelState.read[userId].unread_messages += 1;
+              }
+              if (userId === ownUserId && channelState.read[ownUserId].unread_messages === 1) {
+                channelState.read[ownUserId].first_unread_message_id = event.message.id;
               }
             }
           }
@@ -1481,7 +1485,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     state: ChannelAPIResponse<StreamChatGenerics>,
     messageSetToAddToIfDoesNotExist: MessageSetType = 'latest',
   ) {
-    const { state: clientState, user, userID } = this.getClient();
+    const { state: clientState, user } = this.getClient();
 
     // add the Users
     if (state.members) {
@@ -1520,20 +1524,6 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
       }
     }
 
-    // initialize read state to last message or current time if the channel is empty
-    // if the user is a member, this value will be overwritten later on otherwise this ensures
-    // that everything up to this point is not marked as unread
-    if (userID != null) {
-      const last_read = this.state.last_message_at || new Date();
-      if (user) {
-        this.state.read[user.id] = {
-          user,
-          last_read,
-          unread_messages: 0,
-        };
-      }
-    }
-
     // apply read state if part of the state
     if (state.read) {
       for (const read of state.read) {
@@ -1545,6 +1535,17 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
         };
 
         if (read.user.id === user?.id) {
+          if (this.state.read[user.id].last_read) {
+            for (let i = 0; i < messageSet.messages.length; i++) {
+              const msg = messageSet.messages[i];
+              const isOwnMsg = msg.user?.id === user.id;
+              const isRead = !msg.created_at || new Date(msg.created_at) <= this.state.read[user.id].last_read;
+              if (isOwnMsg || isRead) continue;
+              this.state.read[read.user.id].first_unread_message_id = msg.id;
+              break;
+            }
+          }
+
           this.state.unreadCount = this.state.read[read.user.id].unread_messages;
         }
       }
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1299,6 +1299,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
               } else {
                 channelState.read[userId].unread_messages += 1;
               }
+              // register the first unread msg id for the current user if this new message is the only one unread
               if (userId === ownUserId && channelState.read[ownUserId].unread_messages === 1) {
                 channelState.read[ownUserId].first_unread_message_id = event.message.id;
               }
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1528,15 +1528,19 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
     // apply read state if part of the state
     if (state.read) {
       for (const read of state.read) {
-        this.state.read[read.user.id] = {
+        const previousReadState = this.state.read[read.user.id];
+        const newReadState = {
           last_read: new Date(read.last_read),
           last_read_message_id: read.last_read_message_id,
           unread_messages: read.unread_messages ?? 0,
           user: read.user,
         };
+        this.state.read[read.user.id] = newReadState;
 
         if (read.user.id === user?.id) {
-          if (this.state.read[user.id].last_read) {
+          const lastReadChanged =
+            newReadState.last_read && newReadState.last_read.getTime() !== previousReadState?.last_read.getTime();
+          if (lastReadChanged) {
             for (let i = 0; i < messageSet.messages.length; i++) {
               const msg = messageSet.messages[i];
               const isOwnMsg = msg.user?.id === user.id;
@@ -1545,6 +1549,8 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
               this.state.read[read.user.id].first_unread_message_id = msg.id;
               break;
             }
+          } else {
+            this.state.read[read.user.id].first_unread_message_id = previousReadState.first_unread_message_id;
           }
 
           this.state.unreadCount = this.state.read[read.user.id].unread_messages;
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -170,18 +170,10 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
    * @return {Promise<SendMessageAPIResponse<StreamChatGenerics>>} The Server Response
    */
   async sendMessage(message: Message<StreamChatGenerics>, options?: SendMessageOptions) {
-    const sendMessageResponse = await this.getClient().post<SendMessageAPIResponse<StreamChatGenerics>>(
-      this._channelURL() + '/message',
-      {
-        message,
-        ...options,
-      },
-    );
-
-    // Reset unreadCount to 0.
-    this.state.unreadCount = 0;
-
-    return sendMessageResponse;
+    return await this.getClient().post<SendMessageAPIResponse<StreamChatGenerics>>(this._channelURL() + '/message', {
+      message,
+      ...options,
+    });
   }
 
   sendFile(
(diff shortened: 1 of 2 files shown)


szuperaz (commented):
Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
- `message.read` event
- `notification.mark_unread` event
- `_initializeState` function

MartinCupela:
> Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> 
> * `message.read` event
> * `notification.mark_unread` event
> * `_initializeState` function

I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.

szuperaz:
> > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > 
> > * `message.read` event
> > * `notification.mark_unread` event
> > * `_initializeState` function
> 
> I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.

Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.

It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .

MartinCupela:
I 

> > > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > > 
> > > * `message.read` event
> > > * `notification.mark_unread` event
> > > * `_initializeState` function
> > 
> > 
> > I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.
> 
> Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.
> 
> It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .

I personally have some read states in queried channels without `last_read_message_id` + the type is declared with it as optional whereas `last_read` is mandatory.

image

szuperaz:
> I
> 
> > > > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > > > 
> > > > * `message.read` event
> > > > * `notification.mark_unread` event
> > > > * `_initializeState` function
> > > 
> > > 
> > > I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.
> > 
> > 
> > Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.
> > It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .
> 
> I personally have some read states in queried channels without `last_read_message_id` + the type is declared with it as optional whereas `last_read` is mandatory.
> 
> image

But this is expected, and the iOS algorithm handles it (`Scenario 3`). I think the `last_read_message_id` is missing in this specific example because you haven't opened the channel since the `last_read_message_id` field has been implemented.

MartinCupela:
> > I
> > > > > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > > > > 
> > > > > * `message.read` event
> > > > > * `notification.mark_unread` event
> > > > > * `_initializeState` function
> > > > 
> > > > 
> > > > I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.
> > > 
> > > 
> > > Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.
> > > It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .
> > 
> > 
> > I personally have some read states in queried channels without `last_read_message_id` + the type is declared with it as optional whereas `last_read` is mandatory.
> > image
> 
> But this is expected, and the iOS algorithm handles it (`Scenario 3`). I think the `last_read_message_id` is missing in this specific example because you haven't opened the channel since the `last_read_message_id` field has been implemented.

So you are suggesting to use `last_read` as a fallback in case `last_read_message_id` does not exist?

szuperaz:
> > > I
> > > > > > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > > > > > 
> > > > > > * `message.read` event
> > > > > > * `notification.mark_unread` event
> > > > > > * `_initializeState` function
> > > > > 
> > > > > 
> > > > > I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.
> > > > 
> > > > 
> > > > Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.
> > > > It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .
> > > 
> > > 
> > > I personally have some read states in queried channels without `last_read_message_id` + the type is declared with it as optional whereas `last_read` is mandatory.
> > > image
> > 
> > 
> > But this is expected, and the iOS algorithm handles it (`Scenario 3`). I think the `last_read_message_id` is missing in this specific example because you haven't opened the channel since the `last_read_message_id` field has been implemented.
> 
> So you are suggesting to use `last_read` as a fallback in case `last_read_message_id` does not exist?

No, I'm suggesting to use the algorithm that iOS has implemented. In the case `last_read_message_id` is not defined it uses `Scenario 3`: 

Screenshot 2024-01-19 at 09 53 24

I'm trying to understand why we aren't using the algorithm specified in the Notion document. You mentioned the edge case of `last_read_message_id` not being defined, but the iOS algorithm handles that specific edge case (see above).

MartinCupela:
> > > > I
> > > > > > > Can't we use this algorithm:https://www.notion.so/stream-wiki/Mark-Unread-Developer-Specs-3b05e4fb069a4e2b9355216a5e6085a2?pvs=4#04bc470a33e249c6ba9e7e6c8d6c9279? That algorithm computes `first_unread_message_id` based on `last_read_message_id` so we would have to run that algorithm whenever the `last_read_message_id` can change, which are these places based on the code:
> > > > > > > 
> > > > > > > * `message.read` event
> > > > > > > * `notification.mark_unread` event
> > > > > > > * `_initializeState` function
> > > > > > 
> > > > > > 
> > > > > > I have tested that `last_read_message_id` is not guaranteed to be provided. I actually had an implementation with that one, but as it does not have to be always available I decided to go the safe path.
> > > > > 
> > > > > 
> > > > > Did you find a scenario where the implementation based on `last_read_message_id` wasn't working? I'm asking this because iOS uses that algorithm ([code](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift#L1220)) and mark unread feature is already available there. So if the code works for them, it can work for us too. It feels safer to copy an algorithm already used in an SDK, than to create a new one.
> > > > > It also feels more consistent, even though it's possible your algorithm does the same as iOS but it's hard to compare the two .
> > > > 
> > > > 
> > > > I personally have some read states in queried channels without `last_read_message_id` + the type is declared with it as optional whereas `last_read` is mandatory.
> > > > image
> > > 
> > > 
> > > But this is expected, and the iOS algorithm handles it (`Scenario 3`). I think the `last_read_message_id` is missing in this specific example because you haven't opened the channel since the `last_read_message_id` field has been implemented.
> > 
> > 
> > So you are suggesting to use `last_read` as a fallback in case `last_read_message_id` does not exist?
> 
> No, I'm suggesting to use the algorithm that iOS has implemented. In the case `last_read_message_id` is not defined it uses `Scenario 3`:
> 
> Screenshot 2024-01-19 at 09 53 24
> 
> I'm trying to understand why we aren't using the algorithm specified in the Notion document. You mentioned the edge case of `last_read_message_id` not being defined, but the iOS algorithm handles that specific edge case (see above).

Ok, let me adjust it

src/channel.ts
```diff
@@ -1547,6 +1535,17 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
         };
 
         if (read.user.id === user?.id) {
+          if (this.state.read[user.id].last_read) {
```
szuperaz, line 1538: Can't we only run this code if the underlying data has changed?

src/channel.ts
```diff
@@ -1547,6 +1535,17 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
...
+            for (let i = 0; i < messageSet.messages.length; i++) {
```
szuperaz, line 1539: `findLast` only supported since node v18, nice catch @MartinCupela

src/channel.ts
```diff
@@ -1300,12 +1299,13 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
               } else {
                 channelState.read[userId].unread_messages += 1;
               }
+              if (userId === ownUserId && channelState.read[ownUserId].unread_messages === 1) {
```
szuperaz, line 1302: Can you please add a comment here @MartinCupela ? It's a bit hard to understand

src/channel.ts
MartinCupela: Fixed and added tests

src/channel.ts
MartinCupela: Added a comment.

szuperaz (approved):

MartinCupela:
It will be unnecessary to calculate first unread message id. If not available at the moment of jumping to the first unread message, it will be determined as a message following the last read message.

PR #1212, opened 2024-01-24
feat: add batch unread endpoint

JimmyPettersson85:
## CLA

- [ ] Code changes are tested

add support for the new batch unread endpoint

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -96,6 +96,7 @@ import {
   GetMessageAPIResponse,
   GetRateLimitsResponse,
   GetUnreadCountAPIResponse,
+  GetUnreadCountBatchAPIResponse,
   ListChannelResponse,
   ListCommandsResponse,
   ListImportsPaginationOptions,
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -1688,10 +1689,28 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
 
+  /**
+   * getUnreadCountBatch - Returns unread counts for multiple users at once. Only works server side.
+   *
+   * @param {string[]} [userIDs] List of user IDs to fetch unread counts for.
+   *
+   * @return {<GetUnreadCountBatchAPIResponse>}
+   */
+  async getUnreadCountBatch(userIDs: string[]) {
+    return await this.post<GetUnreadCountBatchAPIResponse>(this.baseURL + '/unread_batch', { user_ids: userIDs });
```
MartinCupela, line 1711: @JimmyPettersson85 just out of curiosity, why for a GET request is used POST method?

JimmyPettersson85: For future proofing the backend we chose `POST` for this endpoint as we might hit querystring limits if we increase the limit of number of users. Currently the limit is 100 user IDs, and with a max length of 255 for a user ID that means a querystring of ~25k chars.

In going with `POST` we can handle > 100 user IDs in the future as well as support compression.

src/types.ts
```diff
@@ -513,6 +513,10 @@ export type GetUnreadCountAPIResponse = APIResponse & {
   total_unread_count: number;
 };
 
+export type GetUnreadCountBatchAPIResponse = APIResponse & {
+  counts_by_user: GetUnreadCountAPIResponse[];
```
akupila, line 517: Should this be a map (object), instead of an array? AFAIK, the key of the map is the user id

I think this should work (_not tested_)

```suggestion
  counts_by_user: { [string]: GetUnreadCountAPIResponse };
```

JimmyPettersson85: good catch, the other parts of this code seems to be using `Record` so maybe
`counts_by_user: Record<string, GetUnreadCountAPIResponse>;`?

akupila: Yeah probably, but my Typescript-foo isn't strong enough to say for sure 😆 @MartinCupela probably knows

oliverlaz: Jumping in here... `counts_by_user: Record<string, GetUnreadCountAPIResponse>;` should be good.
But probably it will be better if we are a bit more explicit and have:
`{ [userId: string]: GetUnreadCountAPIResponse }`

JimmyPettersson85: thanks for the input, pushed a fix for this

oliverlaz (approved):

PR #1216, opened 2024-01-30
feat: add new api for send a refresh-token

guerinoni:
## CLA

- [ ] Code changes are tested

## Changelog

-


oliverlaz:
hey @guerinoni, on second thought, we can make this integration even easier. 
What if the `client` handles the `token.expiring` event internally and whenever it arrives, it invokes the `tokenProvider()` and automatically sends the new token to the backend? This way, all of our customers will automatically get this feature.

src/client.ts
```diff
@@ -3366,4 +3372,17 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   async commitMessage(id: string) {
     return await this.post<APIResponse & MessageResponse>(this.baseURL + `/messages/${id}/commit`);
   }
+
+  /**
+   * send over a WS the new token to the server, from server you would expect `token.ack` event in case of success.
+   * @param token the new token
+   *
+   */
+  refreshToken() {
```
oliverlaz, line 3381: It would be good to support an optional token as an argument for the "static" token use cases.
Or, in case a customer wants to refresh it long before it expires.

What do you think @szuperaz @guerinoni?

szuperaz: - Static tokens are static by design, if we allow integrators to update them, then they can confuse those with token providers, so I think it's easier not to allow that
- As for refreshing the token long before it expires: I've updated the implementation to allow it: https://github.com/GetStream/stream-chat-js/pull/1216/commits/3a5c24ceeea6d00c7e02218ed0645cd1f962ecc0 - I also added a remark to let integrators know, they only should do it in special cases

oliverlaz: fair point! thank you!

oliverlaz (dismissed):

oliverlaz (approved):

PR #1237, opened 2024-02-22
feat: add a way to undelete a deleted messages

akupila:
## CLA

- [ ] Code changes are tested

This adds support for calling a new undelete endpoint, which can be used for undeleting previously soft deleted messages. The API requires this to be called from serverside clients and will return a 403 if called from the browser.

The API will also publish a new `message.undeleted` event when the message has been undelete. I'm not sure if more changes are required here beyond adding it to the event map.

## Changelog

- Add `undeleteMessage` for bringing back soft deleting messages

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2603,6 +2603,27 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     );
   }
 
+  /**
+   * undeleteMessage - Undelete a message
+   *
+   * undeletes a message that was previous soft deleted. Hard deleted messages
+   * cannot be undeleted. This is only allowed to be called from server-side
+   * clients.
+   *
+   * @param {string} messageID
+   *
+   * @return {{ message: MessageResponse<StreamChatGenerics> }} Response that includes the message
+   */
+  async undeleteMessage(messageID: string) {
+    if (!this._isUsingServerAuth()) {
+      throw Error('Messages can only be deleted with a server-side client');
+    }
+
+    return await this.post<APIResponse & { message: MessageResponse<StreamChatGenerics> }>(
+      this.baseURL + `/messages/${messageID}/undelete`,
+    );
+  }
+
   async getMessage(messageID: string, options?: GetMessageOptions) {
     return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(
       this.baseURL + `/messages/${encodeURIComponent(messageID)}`,
--- a/src/client.ts
+++ b/src/client.ts
@@ -2616,7 +2616,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    */
   async undeleteMessage(messageID: string) {
     if (!this._isUsingServerAuth()) {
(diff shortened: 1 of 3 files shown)


akupila:
The tests [are failing](https://github.com/GetStream/stream-chat-js/actions/runs/8008694295/job/21875676530?pr=1237#step:4:103) here, but i'm not entirely sure why. Happy to fix if somebody can provide some guidance on what needs to be done :

src/client.ts
```diff
@@ -2599,6 +2599,27 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+   * cannot be undeleted. This is only allowed to be called from server-side
+   * clients.
+   *
+   * @param {string} messageID
+   *
+   * @return {{ message: MessageResponse<StreamChatGenerics> }} Response that includes the message
+   */
+  async undeleteMessage(messageID: string) {
+    if (!this._isUsingServerAuth()) {
+      throw Error('Messages can only be deleted with a server-side client');
```
oliverlaz, line 2615:
```suggestion
      throw new Error('Messages can only be undeleted with a server-side client');
```

Nitpicking 😅

oliverlaz (approved):

oliverlaz (dismissed):

oliverlaz:
> The tests [are failing](https://github.com/GetStream/stream-chat-js/actions/runs/8008694295/job/21875676530?pr=1237#step:4:103) here, but i'm not entirely sure why. Happy to fix if somebody can provide some guidance on what needs to be done 🙂

I see `threads` there. Perhaps @vishalnarkhede knows more about this one :)

src/client.ts
akupila: Whoops, good one! 😅

oliverlaz (dismissed):

arnautov-anton:
We're missing a WS event handler here - is that expected?

akupila:
> We're missing a WS event handler here - is that expected?

Sorry, I'm not very familiar with this codebase. I didn't deliberately leave it out if it's required. Could you point me in the right direction, and I'll try to fix it?

arnautov-anton:
> > We're missing a WS event handler here - is that expected?
> 
> Sorry, I'm not very familiar with this codebase. I didn't deliberately leave it out if it's required. Could you point me in the right direction, and I'll try to fix it?

Oh, no worries! I'm not sure what the specs are here but it should look something like this:
https://github.com/GetStream/stream-chat-js/blob/cf275838bab1f5b5912082cc97cb016584126bc7/src/channel.ts#L1272-L1284

akupila:
@arnautov-anton thanks, that makes a lot of sense. AFAIK, the handler should be basically the same as for when a message is updated. I did some manual testing with this event sent from the API instead, and messages re-appeared. I added this ebb589f454de7b90a24f8f3c35defbe4ee199599

arnautov-anton (approved):

vishalnarkhede (approved):
Ignore the failing type test. I will fix those separately. You can even disable them for now.

PR #1252, opened 2024-03-04
feat: add support for show_deleted_message in `getMessage`

guerinoni:
## CLA

- [x] Code changes are tested

This allow the optional query parameter `show_deleted_message` in API `getMessage` in order to show text of deleted msg.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -172,6 +172,7 @@ import {
   SegmentTargetsResponse,
   QuerySegmentTargetsFilter,
   SortParam,
+  GetMessageOptions,
 } from './types';
 import { InsightMetrics, postInsights } from './insights';
 import { Thread } from './thread';
@@ -2602,10 +2603,14 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     );
   }
 
-  async getMessage(messageID: string) {
-    return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(
-      this.baseURL + `/messages/${encodeURIComponent(messageID)}`,
-    );
+  async getMessage(messageID: string, options?: GetMessageOptions) {
+    const url = this.baseURL + `/messages/${messageID}`;
+    let params = {};
+    if (options?.show_deleted_message) {
+      params = { show_deleted_message: true };
+    }
+
+    return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(url, params);
   }
 
   /**
--- a/src/types.ts
+++ b/src/types.ts
@@ -2218,6 +2218,10 @@ export type UpdateMessageOptions = {
   skip_enrich_url?: boolean;
 };
 
+export type GetMessageOptions = {
+  show_deleted_message?: boolean;
+};
+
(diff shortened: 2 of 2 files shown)


src/client.ts
```diff
@@ -2602,10 +2602,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     );
   }
 
-  async getMessage(messageID: string) {
-    return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(
-      this.baseURL + `/messages/${encodeURIComponent(messageID)}`,
-    );
+  async getMessage(messageID: string, show_deleted_message?: boolean) {
```
totalimmersion, line 2605: Instead of a single parameter, I would add it as an options object:
`options: GetMessageOptions`
```suggestion
  async getMessage(messageID: string, options?: GetMessageOptions) {
  
  type GetMessageOptions = {
    show_deleted_message?: boolean
  }
```

This way it's more easily extendable in the future in case you want to add another option / parameter

totalimmersion (changes requested):

src/client.ts
```diff
@@ -2602,10 +2602,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+    let url = this.baseURL + `/messages/${messageID}`;
```
MartinCupela, line 2606: Please check the [signature of `StreamChat.get()` method](https://github.com/GetStream/stream-chat-js/blob/a80813f6cc1ec83fcfc770d4196146db3bbcef55/src/client.ts#L1019). It allows you to pass params object so that you do not have to concatenate the URL search params in this way.

src/client.ts
guerinoni: Yes nice one! Done

src/client.ts
guerinoni: Oh thanks, fixed it :D

src/client.ts
```diff
@@ -2602,10 +2603,14 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
-      this.baseURL + `/messages/${encodeURIComponent(messageID)}`,
-    );
+  async getMessage(messageID: string, options?: GetMessageOptions) {
+    const url = this.baseURL + `/messages/${messageID}`;
+    let params = {};
+    if (options?.show_deleted_message) {
+      params = { show_deleted_message: true };
+    }
+
+    return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(url, params);
```
MartinCupela, line 2613:
```suggestion
  async getMessage(messageID: string, options?: GetMessageOptions) {
    return await this.get<GetMessageAPIResponse<StreamChatGenerics>>(
      this.baseURL + `/messages/${encodeURIComponent(messageID)}`,
        {...options}
    );
```

guerinoni: Thanks! done

totalimmersion (approved):

MartinCupela (approved):

vishalnarkhede (approved):

vishalnarkhede:
@guerinoni you can merge it, I will fix the type test separately. Its not related to this PR

PR #1277, opened 2024-04-17
feat: user privacy settings

myandrienko:
Adds support for user-level privacy settings. A new field `privacy_settings` can be passed in a user object in `connectUser()` and `upsertUser()`.

Generally, just passing this data to backend should be enough (backend will discard read/typing events accordingly). However, we go one step further and also prevent the client from sending `typing.start` and `typing.stop` events at all.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -55,7 +55,6 @@ import {
   QueryChannelAPIResponse,
   PollVoteData,
   SendMessageOptions,
-  UserPrivacySettings,
 } from './types';
 import { Role } from './permissions';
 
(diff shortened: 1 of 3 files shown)


src/channel.ts
```diff
@@ -1630,6 +1631,18 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
...
+    if (!this.getConfig()?.typing_events) {
+      return false;
+    }
+
+    const userPrivacySettings: Partial<UserPrivacySettings> = this.getClient().user?.privacy_settings || {};
+    const defaultPrivacySettings: Partial<UserPrivacySettings> =
+      (await this.getClient().getAppSettings()).app?.chat_default_user_privacy_settings || {};
+
+    return userPrivacySettings.typing_indicators?.enabled ?? defaultPrivacySettings.typing_indicators?.enabled ?? true;
+  }
```
myandrienko, line 1644: 🍒

src/client.ts
```diff
@@ -700,9 +701,15 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+   * @param [options={}] App settings are cached by default. Pass { force: true } to force cache refresh.
    */
-  async getAppSettings() {
-    return await this.get<AppSettingsAPIResponse<StreamChatGenerics>>(this.baseURL + '/app');
+  async getAppSettings(options: { force?: boolean } = {}) {
+    if (options.force || !this._appSettings) {
+      this._appSettings = await this.get<AppSettingsAPIResponse<StreamChatGenerics>>(this.baseURL + '/app');
+    }
+
+    return this._appSettings;
```
myandrienko, line 712: In case user-level privacy settings are not provided, the client should fall back to app-level settings. We don't want to fetch app settings on every keystroke event, so adding caching here.

In practice this is not a big change, because the React SDK already does caches app settings on component level. (That caching can be removed and after this PR is merged.)

oliverlaz: Can you check how other JS SDKs handle this? This could be potentially a breaking change.

oliverlaz:
See #1283

PR #1286, opened 2024-04-30
feat: add reaction groups fallback

myandrienko:
Adds fallback for the new `reaction_groups` property on messages.

Since `reaction_groups` will not be available for older messages, we use the data from `reaction_counts` and `reaction_scores` to provide reactions data in a similar shape as a fallack.

The `reaction_groups` property is marked as optional.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -627,7 +627,6 @@ export type MessageResponse<
 export type MessageResponseBase<
   StreamChatGenerics extends ExtendableGenerics = DefaultGenerics
 > = MessageBase<StreamChatGenerics> & {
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -651,6 +650,7 @@ export type MessageResponseBase<
   pinned_by?: UserResponse<StreamChatGenerics> | null;
   poll?: PollResponse<StreamChatGenerics>;
   reaction_counts?: { [key: string]: number } | null;
+  reaction_groups?: { [key: string]: ReactionGroupResponse } | null;
```
myandrienko, line 653: 🍒 Now optional.

oliverlaz (approved):

PR #1287, opened 2024-04-30
fix: fix filter type for query reactions

myandrienko:
According to the [implementation](https://github.com/GetStream/chat/blob/master/utils/mq/reaction/reaction.go#L38-L53), reactions filter only supports the following fields:

- user_id
- type
- created_at

So no need to include any other keys in the filter type. Adding all other keys from the `Reaction` type breaks the filter type, because `Reaction` type is overridable via generic params.

Also, adds pagination `next` field to response.


oliverlaz (approved):

guerinoni (approved):
Thanks, not super skilled in JS :)

PR #1326, opened 2024-07-03
chore: allow string arrays on sort parameters

xernobyl:
## CLA

- [x] Code changes are tested

Added support to Sort by arrays of strings:

```
const { members } = await channel.queryMembers(
    {},
    { channel_role: ['channel_moderator', 'channel_member'], name: 1 },
);
```

Now users should be able to query members sorted by role.

## Changelog

-

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1790,7 +1790,7 @@ export type PinnedMessagesSort = PinnedMessagesSortBase | Array<PinnedMessagesSo
 export type PinnedMessagesSortBase = { pinned_at?: AscDesc };
 
 export type Sort<T> = {
-  [P in keyof T]?: AscDesc;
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -1793,13 +1793,17 @@ export type Sort<T> = {
   [P in keyof T]?: AscDesc;
 };
 
+export type SortMember<T> = {
```
MartinCupela, line 1796: Why this type contains Member in its name? Seems like generic that just differs from Sort in that it allows string[]. What string values can be passed here?

xernobyl: I'm adding support to sort members by role, and the idea is to pass something like {channel_role:["channel_admin", "channel_member"]}

MartinCupela: 👍

PR #1330, opened 2024-07-25
feat: threads 2.0

arnautov-anton:
This PR adds two new classes: ThreadManager and Thread. These classes are designed to be used as "component backends" - the components being ThreadList and Thread respectively. To achieve easy integration with the UI framework, these classes expose `state` - a reactive state store you can subscribe to. Ideally, components contain no business logic and just subscribe to relevant parts of the state.

## StateStore

This is an immutable store you can subscribe to. Get the latest value with `getLatestValue()`, update value with `next` or `partialNext`, subscribe with `subscribe` or `subscribeWithSelector`.

## Thread

Thread loads, stores and paginates thread replies. It also keeps track of read states.

For some level of consistency with [channel messages pagination indicators](https://github.com/GetStream/stream-chat-js/pull/1332), we have a `pagination` object which stores cursors and loading indicators.

After calling `registerSubscriptions`, Thread instance starts listening to channel events. See a list of `subscribe...` methods to see which exact events we are listening to.

Thread can be active or inactive. This is a UI state - an integrator should mark visible/opened threads as active. Active thread immediately marks all the incoming messages as read.

Although backend supports a lot of query options for thread replies, we only support querying latest replies in chronological order, and then paginating through them. If we were to support all query options, we'd have to:

1. Support message sets for replies like Channel does
2. Implement [complex pagination indicator logic](https://github.com/GetStream/stream-chat-js/pull/1332/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R402-R629)

You are not likely to construct a Thread instance by yourself - you get by calling `channel.getThread()` instead.

## ThreadManager

Loads, stores and paginates threads.

ThreadManager does its best to keep the order of threads in a list stable (which is a business requirement for the ThreadList component). So when a reply arrives in a thread which is not currently in a list, we don't load the thread - instead, we add it to the list of "unseen threads". The UI should then prompt user to reload the list if they require so.

Similarly, when a reply arrives in a thread which is already in the list, we don't reorder the list, but set a `isThreadOrderStale` flag.

ThreadManager makes sure that all Thread instances it stores are as up-to-date as possible.


src/thread.ts
```diff
@@ -1,142 +1,873 @@
...
+      this.updateParentMessageOrReplyLocally(event.message);
+    };
+
+    ['message.updated', 'message.deleted', 'reaction.new', 'reaction.deleted'].forEach((eventType) => {
+      this.unsubscribeFunctions.add(this.client.on(eventType, handleMessageUpdate).unsubscribe);
+    });
+  };
+
+  public deregisterSubscriptions = () => {
+    this.unsubscribeFunctions.forEach((cleanupFunction) => cleanupFunction());
```
myandrienko, line 366: Shall we `this.unsubscribeFunctions.clear()` here?

arnautov-anton: Yes, good point. :)

src/thread_manager.ts
```diff
@@ -0,0 +1,334 @@
...
+    const { nextCursor, loadingNextPage } = this.state.getLatestValue();
+
+    if (nextCursor === null || loadingNextPage) return;
+
+    const optionsWithNextCursor: QueryThreadsOptions = {
+      ...options,
+      next: nextCursor,
+    };
+
+    this.state.next((current) => ({ ...current, loadingNextPage: true }));
```
MartinCupela, line 318: Could be here used `this.state.partialNext({ loadingNextPage: true })`?

src/thread_manager.ts
```diff
@@ -33,14 +32,17 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
   public readonly state: StateStore<ThreadManagerState<SCG>>;
   private client: StreamChat<SCG>;
   private unsubscribeFunctions: Set<() => void> = new Set();
+  private threadsByIdGetterCache: {
+    threads: ThreadManagerState<SCG>['threads'];
+    threadsById: { [key: string]: Thread<SCG> };
```
myandrienko, line 37: I would advise `{ [key: string]: Thread<SCG> | undefined }` (or just `Record<string, Thread<SCG> | undefined>`), because without `noUncheckedIndexAccess` this is just more true :)

myandrienko (dismissed):

src/thread_manager.ts
arnautov-anton: Good point, fixing right away.

myandrienko (dismissed):

oliverlaz (dismissed):

myandrienko (approved):

PR #1332, opened 2024-08-06
feat: add channel messages pagination indicators

MartinCupela:

1. The React SDK (and probably other SDKs using `stream-chat`) performs redundant channel queries to retrieve more non-existent messages. 
2. Also, as the SDK components do not know, whether there are more messages to be loaded after the initial `client.queryChannels()` call, the SDK's flag determining if there are more messages to load (`hasMore`) is optimistically set to `true`. This breaks customer UIs, which would like to display components based on the value of `hasMore` when it is false in channels with few messages (show component if `hasMore` is `false`).

It makes sense to keep the pagination information in the client instead of the component SDK as the client performs the queries and knows the query options.

The pagination information has to be stored for every message set separately. The pagination is determined separately for `created_at_around`, `id_around` and the rest of the query params. If the query options object contains multiple params, then the pagination is determined as if there was only one of them in the following order: 1. `created_at_around`, 2. `id_around` and 3. the rest of the query params.

### Usage in components SDK:

```js
const hasMore = channel.state.messagesPagination.hasPrev;
const hasMoreNewer = channel.state.messagesPagination.hasNext;
```

## Questions I could not answer

Why the thread messages are not stored in message sets as it is done for channel messages even though they are queried with `MessagePaginationOptions`? Probably jumping to a reply is not possible?
Why pinned messages are not stored in message sets as it is done for channel messages even though they are queried with `PinnedMessagePaginationOptions`? Probably jumping to a pinned message is not possible?


src/client.ts
```diff
@@ -1638,26 +1640,33 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
         c.state.clearMessages();
-        c._initializeState(channelState, 'latest');
+        const { messageSet} = c._initializeState(channelState, 'latest');
+        updatedMessagesSet = messageSet;
+      }
+
+      if (updatedMessagesSet) {
+        updatedMessagesSet.pagination = {
+          ...updatedMessagesSet.pagination,
+          ...messageSetPagination({requestedPageSize: queryChannelsOptions?.message_limit || DEFAULT_QUERY_CHANNELS_MESSAGE_LIST_PAGE_SIZE, returnedPageSize: channelState.messages.length})
```
szuperaz, line 1668: Shouldn't `channelState.messages` be `updatedMessagesSet.messages`? `channelState.messages` will be the active message set, but I think it's possible to update a non-active message set here

szuperaz:
"Why the thread messages are not stored in message sets as it is done for channel messages even though they are queried with MessagePaginationOptions? Probably jumping to a reply is not possible?" -> Thread replies should be stored in message sets, this is a bug. You can jump to thread replies, we have a public method for this, jumping will work fine, but due to the lack of message sets scrolling won't work in a lot of cases. This was an oversight from me when I implemented this years ago, I'm pretty sure I had this reported somewhere before, but I couldn't find it, so created a new issue for it: https://stream-io.atlassian.net/jira/software/c/projects/PBE/boards/168?selectedIssue=PBE-5451

 "Why pinned messages are not stored in message sets as it is done for channel messages even though they are queried with PinnedMessagePaginationOptions? Probably jumping to a pinned message is not possible?" -> There are two things here: jumping to a pinned message is possible, it's works just like jumping to any channel/thread message. The message set would be necessary if someone were to build a message list that only contains pinned messages, and want to jump inside that list, that is not possible, nor do I think we ever wanted to support that, so it's fine that pinned messages don't have a message sets.

szuperaz (dismissed):

oliverlaz (dismissed):
It looks good from my side as well.
One nitpick I have is, that I believe the prettier settings were not picked up by your IDE and the formatting is a bit non-standard. Can you please fix this?

src/types.ts
```diff
@@ -2900,6 +2900,12 @@ export type ImportTask = {
 };
 
 export type MessageSetType = 'latest' | 'current' | 'new';
+export type MessageSet<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics> = {
+  isCurrent: boolean;
+  isLatest: boolean;
+  messages: FormatMessageResponse<StreamChatGenerics>[];
+  pagination: { hasNext?: boolean; hasPrev?: boolean };
```
myandrienko, line 2907: If I understand the semantics correctly:
- `true` means there's _maybe_ more to load
- `false` means there's _definitely_ nothing else to load
- `undefined` means nothing has been queried yet

Because both `true` and `undefined` mean "try querying", an easy mistake to make will be writing `if (messages.pagination.hasNext) { ... }`. Maybe we can make this type a bit more verbose:

```
pagination: { maybeHasNext: boolean; maybeHasPrev: boolean }
```

And always initialize with `{ maybeHasNext: true; maybeHasPrev: true }`

MartinCupela: Makes sense to keep it true. But I would keep `maybe` out of the name as it reveals our internals that can change in the future (maybe once this will be fixed on the BE side and each response will provide information about the next page availability :)).

MartinCupela: `hasOlder` - I am not sure whether the messages have to be always ordered by creation date. Also if we wanted to paginate other types of items, they may not be sorted by date - that is why I think it does not give us added value and the integrators should know, how they sort their items.

myandrienko (approved):

src/channel_state.ts
```diff
@@ -846,6 +854,14 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
         sources.forEach((messageSet) => {
           target.isLatest = target.isLatest || messageSet.isLatest;
           target.isCurrent = target.isCurrent || messageSet.isCurrent;
+          target.pagination.hasPrev =
```
MartinCupela, line 857: Merging pagination objects when merging message sets.

src/client.ts
```diff
@@ -1601,7 +1603,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       },
     });
 
-    return this.hydrateActiveChannels(data.channels, stateOptions);
+    return this.hydrateActiveChannels(data.channels, stateOptions, options);
```
MartinCupela, line 1606: Spaghetti code architecture forced me to do this.

src/client.ts
```diff
@@ -1638,26 +1640,37 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+        updatedMessagesSet = messageSet;
       } else if (!skipInitialization.includes(channelState.channel.id)) {
         c.state.clearMessages();
-        c._initializeState(channelState, 'latest');
+        const { messageSet } = c._initializeState(channelState, 'latest');
+        updatedMessagesSet = messageSet;
+      }
+
+      if (updatedMessagesSet) {
+        updatedMessagesSet.pagination = {
```
MartinCupela, line 1666: Pagination indicators are calculated once the requested page is merged into the parent message set.

src/utils.ts
```diff
@@ -396,3 +398,232 @@ function maybeGetReactionGroupsFallback(
...
+
+export const messageSetPagination = <StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>(
+  params: MessagePaginationUpdatedParams<StreamChatGenerics>,
+) => {
+  if (params.parentSet.messages.length < params.returnedPage.length) {
+    console.error('Corrupted message set state: parent set size < returned page size');
+    return params.parentSet.pagination;
+  }
+
+  if (params.messagePaginationOptions?.created_at_around) {
```
MartinCupela, line 622: If the query options object contains multiple params, then the pagination is determined as if there was only one of them in the following order: 1. created_at_around, 2. id_around and 3. the rest of the query params.

src/utils.ts
```diff
@@ -396,3 +398,232 @@ function maybeGetReactionGroupsFallback(
...
+    newPagination.hasNext = hasNext;
+
+  return newPagination;
+};
+
+export const messageSetPagination = <StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>(
+  params: MessagePaginationUpdatedParams<StreamChatGenerics>,
+) => {
+  if (params.parentSet.messages.length < params.returnedPage.length) {
+    console.error('Corrupted message set state: parent set size < returned page size');
```
MartinCupela, line 618: TODO: Log with client logger

oliverlaz (dismissed):

oliverlaz (approved):

PR #1347, opened 2024-08-27
fix: fix most TODOs and tests for Threads class

myandrienko:
Notable changes:

- Removed throttling from all subscriptions, should work properly without any throttling.
- Fixed handling read states (see `subscribeNewReplies` and `markAsRead`). Got rid of the `incrementOwnUnreadCount` - it's now part of a more generic update performed on `message.new`.
- Fixed a bug caused by typo in pagination logic.
- Updated most tests. Replaced spying on methods (which tests implementation details) with checking resulting state.
- Added tests for pagination logic.
- All tests for Thread are now green.
- Either fixed TODOs that were fixable, or removed TODOs that we are not going to get to any time soon.

Don't expect green CI in this PR. These changes are in a separate PR so that reviewing is at least feasible :)

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -12,7 +12,7 @@ import type {
   ThreadResponse,
   UserResponse,
 } from './types';
-import { addToMessageList, findIndexInSortedArray, formatMessage } from './utils';
+import { addToMessageList, findIndexInSortedArray, formatMessage, throttle } from './utils';
 
 type QueryRepliesOptions<SCG extends ExtendableGenerics> = {
   sort?: { created_at: AscDesc }[];
@@ -63,7 +63,7 @@ export type ThreadReadState<SCG extends ExtendableGenerics = DefaultGenerics> = 
 
 const DEFAULT_PAGE_LIMIT = 50;
 const DEFAULT_SORT: { created_at: AscDesc }[] = [{ created_at: -1 }];
-export const DEFAULT_MARK_AS_READ_THROTTLE_DURATION = 1000;
+const MARK_AS_READ_THROTTLE_TIMEOUT = 1000;
 
 export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
   public readonly state: StateStore<ThreadState<SCG>>;
@@ -181,14 +181,15 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
     this.unsubscribeFunctions.add(this.subscribeMessageUpdated());
   };
 
-  private subscribeMarkActiveThreadRead = () =>
-    this.state.subscribeWithSelector(
+  private subscribeMarkActiveThreadRead = () => {
+    return this.state.subscribeWithSelector(
       (nextValue) => [nextValue.active, ownUnreadCountSelector(this.client.userID)(nextValue)],
       ([active, unreadMessageCount]) => {
         if (!active || !unreadMessageCount) return;
-        this.markAsRead();
+        this.throttledMarkAsRead();
       },
     );
+  };
 
   private subscribeReloadActiveStaleThread = () =>
     this.state.subscribeWithSelector(
@@ -218,6 +219,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
       }
 
       const isOwnMessage = event.message.user?.id === this.client.userID;
+      const { active, read } = this.state.getLatestValue();
 
       this.upsertReplyLocally({
         message: event.message,
@@ -226,27 +228,40 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
         timestampChanged: isOwnMessage,
       });
 
-      const { read } = this.state.getLatestValue();
-      const readUpdate: ThreadReadState = {};
+      if (active) {
+        this.throttledMarkAsRead();
+      }
+
+      const nextRead: ThreadReadState = {};
 
       for (const userId of Object.keys(read)) {
         if (read[userId]) {
-          readUpdate[userId] = {
-            ...read[userId],
-            ...(userId === event.user?.id
-              ? {
-                  lastReadAt: event.created_at ? new Date(event.created_at) : new Date(),
-                  user: event.user,
-                  unreadMessageCount: 0,
-                }
-              : {
-                  unreadMessageCount: read[userId].unreadMessageCount + 1,
-                }),
-          };
+          let nextUserRead: ThreadUserReadState = read[userId];
+
+          if (userId === event.user?.id) {
+            // The user who just sent a message to the thread has no unread messages
+            // in that thread
+            nextUserRead = {
+              ...nextUserRead,
+              lastReadAt: event.created_at ? new Date(event.created_at) : new Date(),
+              user: event.user,
+              unreadMessageCount: 0,
+            };
+          } else if (active && userId === this.client.userID) {
+            // Do not increment unread count for the current user in an active thread
+          } else {
+            // Increment unread count for all users except the author of the new message
+            nextUserRead = {
+              ...nextUserRead,
+              unreadMessageCount: read[userId].unreadMessageCount + 1,
+            };
+          }
+
+          nextRead[userId] = nextUserRead;
         }
       }
 
-      this.state.partialNext({ read: readUpdate });
+      this.state.partialNext({ read: nextRead });
     }).unsubscribe;
 
   private subscribeRepliesRead = () =>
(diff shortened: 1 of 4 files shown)


src/thread.ts
```diff
@@ -140,226 +146,200 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
-        (nextValue) => [nextValue.active, this.client.userID && nextValue.read[this.client.userID]?.unreadMessageCount],
-        ([active, unreadMessagesCount]) => {
-          if (!active || !unreadMessagesCount) return;
+    this.unsubscribeFunctions.add(this.subscribeMarkActiveThreadRead());
+    this.unsubscribeFunctions.add(this.subscribeReloadActiveStaleThread());
+    this.unsubscribeFunctions.add(this.subscribeMarkThreadStale());
+    this.unsubscribeFunctions.add(this.subscribeNewReplies());
+    this.unsubscribeFunctions.add(this.subscribeRepliesRead());
+    this.unsubscribeFunctions.add(this.subscribeReplyDeleted());
+    this.unsubscribeFunctions.add(this.subscribeMessageUpdated());
```
arnautov-anton, line 181: image

src/thread.ts
```diff
@@ -140,226 +146,200 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
-          // mark thread as read whenever thread becomes active or is already active and unread messages count increases
-          throttledMarkAsRead();
-        },
-      ),
+  private subscribeMarkActiveThreadRead = () =>
+    this.state.subscribeWithSelector(
+      (nextValue) => [nextValue.active, ownUnreadCountSelector(this.client.userID)(nextValue)],
+      ([active, unreadMessageCount]) => {
+        if (!active || !unreadMessageCount) return;
+        this.markAsRead();
```
arnautov-anton, line 189: Throttling in this case might be necessary as in a thread that is pretty active (read "getting spammed by other people") this could be a cause for rate-limitting.

src/thread.ts
```diff
@@ -140,226 +146,200 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
-    };
-
-    ['message.updated', 'message.deleted', 'reaction.new', 'reaction.deleted'].forEach((eventType) => {
-      this.unsubscribeFunctions.add(this.client.on(eventType, handleMessageUpdate).unsubscribe);
-    });
+    return () => unsubscribeFunctions.forEach((unsubscribe) => unsubscribe());
   };
 
-  public deregisterSubscriptions = () => {
+  public unregisterSubscriptions = () => {
```
arnautov-anton, line 295: I was following [this post](https://english.stackexchange.com/a/40095) on how to name this one, I'm fine with either though (we user `subscribe`/`unsubscribe` anyway).

src/thread.ts
```diff
@@ -140,226 +146,200 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+        // Message from current user could have been added optimistically,
+        // so the actual timestamp might differ in the event
+        timestampChanged: isOwnMessage,
+      });
+
+      const { read } = this.state.getLatestValue();
+      const readUpdate: ThreadReadState = {};
+
+      for (const userId of Object.keys(read)) {
+        if (read[userId]) {
```
arnautov-anton, line 233: Not sure I understand this part - if we can target specific `userId` (based on `event.user`) in a pre-formatted read map then why going over all of the other users?

src/thread.ts
myandrienko: Good point. I'm going to denounce the underlying request.

src/thread.ts
```diff
@@ -401,33 +382,47 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+          ...current,
+          read: {
+            ...current.read,
+            [currentUserId]: {
+              ...currentUserRead,
+              unreadMessageCount: 0,
+            },
+          },
+        }));
+      }
```
arnautov-anton, line 402: Is this an optimistic update of sorts? We're reacting to `message.read` event to reset this and it's such a value where optimistic update could be omitted in my opinion.

src/thread.ts
```diff
@@ -481,3 +476,9 @@ const repliesPaginationFromInitialThread = (thread: ThreadResponse): ThreadRepli
     isLoadingPrev: false,
   };
 };
+
+const ownUnreadCountSelector = (currentUserId: string | undefined) => <
+  SCG extends ExtendableGenerics = DefaultGenerics
+>(
+  state: ThreadState<SCG>,
+) => (currentUserId && state.read[currentUserId]?.unreadMessageCount) || 0;
```
arnautov-anton, line 484: I like this. 💯

src/thread.ts
myandrienko: That's because a new message should be considered unread by everybody (we know of) until the `message.read` event arrives.

`event.user` will be the author of the message, and we have to increment the unread count for everybody else.

See how Channel handles `message.new`.

src/thread.ts
myandrienko: The optimistic update is useful for active threads. It marks an active thread read immediately, which I think is nicer than having a short period when an active thread is unread.

src/thread.ts
myandrienko: I'm trying to remember why it was renamed - I think for consistency with `unsubscribe` somewhere else :) I'll give it another thought before merging.

arnautov-anton: Let's keep unregister for consistency, I think it's fine. :)

src/thread.ts
arnautov-anton: Let's add a TODO then to handle reset to a previous state in case `markRead` request fails. I think it's highly unlikely but a possibility nonetheless.

test/unit/threads.test.ts
```diff
@@ -5,575 +5,778 @@ import { generateChannel } from './test-utils/generateChannel';
...
+          const message = generateMsg({ parent_id: thread.id }) as MessageResponse;
+          const stateBefore = thread.state.getLatestValue();
+          expect(stateBefore.replies).to.have.lengthOf(0);
 
-          thread.upsertReplyLocally({ message: newMessage as MessageResponse });
+          thread.upsertReplyLocally({ message });
 
-          expect(thread.state.getLatestValue().latestReplies).to.have.lengthOf(1);
-          expect(thread.state.getLatestValue().latestReplies[0].id).to.equal(newMessage.id);
+          const stateAfter = thread.state.getLatestValue();
```
arnautov-anton, line 77: This reads much better. Sorry and thank you! :)

test/unit/threads.test.ts
```diff
@@ -5,575 +5,778 @@ import { generateChannel } from './test-utils/generateChannel';
...
           });
 
-          expect(deleteReplyLocallySpy.calledOnce).to.be.true;
-          expect(updateParentMessageOrReplyLocallySpy.called).to.be.false;
+          const stateAfter = thread.state.getLatestValue();
+          expect(stateAfter.replies).to.have.lengthOf(4);
+          expect(stateAfter.replies.find((reply) => reply.id === messageToDelete.id)).to.be.undefined;
+
+          thread.unregisterSubscriptions();
         });
```
arnautov-anton, line 762: Re-testing delete seemed redundant in this case but maybe I'm missing something?

myandrienko: I wanted to replace the test which tested if `deleteReplyLocally` was called with something that tests state. That's why it's very similar to the test of `deleteReplyLocally`, except that it is triggered by an event.

arnautov-anton: But apart from one extra condition `message.deleted` handler esentially only calls the `deleteReplyLocally` which is being tested higher meaning state-wise it's already tested - no?

src/thread.ts
```diff
@@ -401,33 +382,47 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+    sort = DEFAULT_SORT,
     ...otherOptions
   }: QueryRepliesOptions<SCG> = {}) => {
-    if (!this.channel) throw new Error('queryReplies: This Thread intance has no channel bound to it');
-
     return this.channel.getReplies(this.id, { limit, ...otherOptions }, sort);
   };
 
-  public loadPage = async (count: number) => {
+  public loadPage = async (count: number = DEFAULT_PAGE_LIMIT) => {
```
arnautov-anton, line 418: Just tested this method and while I get why it exists I think it might be confusing to our integrators (pass -50 to count parameter to load previous fifty) - I suspect that this is a method that is going to be used within something like `loadPrevPage` and `loadNextPage` to deduplicate code? Also - in current state it always appends new messages no matter the direction.

myandrienko: I'll keep loadNextPage/loadPrevPage, this method can be private instead.

I'll look into appending order!

test/unit/threads.test.ts
myandrienko: (Decided it's fine to keep both tests. Technically they are checking slightly different things.)

src/thread.ts
myandrienko: Fixed in https://github.com/GetStream/stream-chat-js/pull/1347/commits/805ad1809e5650c1a40059cb90e38ac19ac55641.

1. `markAsRead` no longer optimistically updates read state for current user.
2. We don't count new replies in an active thread towards unread count of the current user.
3. `markAsRead` calls in subscriptions are throttled.

src/thread.ts
myandrienko: Fixed in https://github.com/GetStream/stream-chat-js/commit/805ad1809e5650c1a40059cb90e38ac19ac55641.

arnautov-anton (approved):

PR #1348, opened 2024-08-28
fix: fix most TODOs and tests for ThreadManager class

myandrienko:
Notable changes:

- No longer keep track of loaded list ids that updated while ThreadManager was inactive; looks like it's enough to set a single flag `isThreadOrderStale` instead.
- Replaced usage of `sync` for state recovery after connection drop with a regular reload.
- Streamlined management of subscriptions on underlying thread instances. Thread subscriptions are also unregistered when ThreadManager subscriptions are unregistered, otherwise it's a memory leak.
- Pagination state is now in line with Thread.
- Test should be green.

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -114,7 +114,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
     this.state.partialNext({ active: false });
   };
 
-  public loadState = async () => {
+  public reload = async () => {
     if (this.state.getLatestValue().isLoading) {
       return;
     }
@@ -196,7 +196,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
       (nextValue) => [nextValue.active, nextValue.isStateStale],
       ([active, isStateStale]) => {
         if (active && isStateStale) {
-          this.loadState();
+          this.reload();
         }
       },
     );
--- a/src/thread_manager.ts
+++ b/src/thread_manager.ts
@@ -1,5 +1,4 @@
 import type { StreamChat } from './client';
-import type { Handler } from './store';
 import { StateStore } from './store';
 import type { Thread } from './thread';
 import type { DefaultGenerics, Event, ExtendableGenerics, QueryThreadsOptions } from './types';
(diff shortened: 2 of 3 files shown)


src/thread.ts
```diff
@@ -114,7 +114,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
     this.state.partialNext({ active: false });
   };
 
-  public loadState = async () => {
+  public reload = async () => {
```
myandrienko, line 117: Renamed for consistency with ThreadManager.

src/thread_manager.ts
```diff
@@ -34,17 +38,18 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
     this.client = client;
     this.state = new StateStore<ThreadManagerState<SCG>>({
       active: false,
-      // existing threads which have gotten recent activity during the time the manager was inactive
-      existingReorderedThreadIds: [],
+      isThreadOrderStale: false,
       threads: [],
       threadIdIndexMap: {},
```
arnautov-anton, line 43:
```suggestion
```
I feel like this can be dropped if it's no longer being used anywhere.

src/thread_manager.ts
```diff
@@ -10,20 +9,25 @@ const MAX_QUERY_THREADS_LIMIT = 25;
 
 export type ThreadManagerState<SCG extends ExtendableGenerics = DefaultGenerics> = {
   active: boolean;
-  existingReorderedThreadIds: string[];
-  lastConnectionDownAt: Date | null;
-  loadingNextPage: boolean;
+  isThreadOrderStale: boolean;
+  lastConnectionDropAt: Date | null;
+  pagination: ThreadManagerPagination;
+  ready: boolean;
```
arnautov-anton, line 15: Is this a direct replacement of a
```ts
if (!threads.length && nextCursor !== null) return;
```
in new message handler? I'm not sure if extra property is needed only for this case, the value could be derived from the existing values. Addin this also means that there's an extra one to take care of manually.

src/thread_manager.ts
arnautov-anton: Also - what was the reason for the drop? I feel like with large influx of replies across multiple loaded threads could make a dent in performance with all the slow-ish `findIndex` ops - no (talking about reply handler)?

src/thread_manager.ts
```diff
@@ -56,279 +61,204 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
...
-      if (existsLocally && !existingReorderedThreadIds.includes(parentId) && !active) {
-        return this.state.next((current) => ({
+  private subscribeRecoverAfterConnectionDrop = () => {
+    const unsubscribeConnectionDropped = this.client.on('connection.changed', (event) => {
+      if (event.online === false) {
+        this.state.next((current) => ({
           ...current,
-          existingReorderedThreadIds: current.existingReorderedThreadIds.concat(parentId),
+          lastConnectionDropAt: current.lastConnectionDropAt ?? new Date(),
         }));
```
arnautov-anton, line 136: We're re-creating state here even if the `lastConnectionDropAt` has possibly been already set.

myandrienko: Fixed in https://github.com/GetStream/stream-chat-js/pull/1348/commits/caf81d8f9f3052e25adf772e56e83a7788767f63.

src/thread_manager.ts
myandrienko: Restored the lookup table, but now it's updated together with threads: https://github.com/GetStream/stream-chat-js/pull/1348/commits/ab0db2fee91ce332598cb2619562489da55b5bf9.

src/thread_manager.ts
```diff
@@ -56,279 +61,227 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+  state: ThreadManagerState<SCG>,
+  threads: Thread<SCG>[],
+): Partial<ThreadManagerState<SCG>> {
+  if (threads === state.threads) {
+    return {};
+  }
+
+  return {
+    threads,
+    threadsById: threads.reduce<Record<string, Thread<SCG>>>((acc, thread) => {
```
arnautov-anton, line 282: nit:
```suggestion
    threadsById: threads.reduce<Record<string, Thread<SCG>>>((newThreadsById, thread) => {
```

arnautov-anton (approved):

src/thread_manager.ts
myandrienko: Fixed in https://github.com/GetStream/stream-chat-js/pull/1348/commits/07e588099c83a7fa2f24440a9bfe3ebb5a848c8a :)

PR #1352, opened 2024-09-06
fix(threads): restore reply soft-deletion

arnautov-anton:

Restores soft-deletion of thread replies within Thread instance.


isekovanic (approved):
Cool ! Thanks 🚀

oliverlaz (approved):

myandrienko (approved):

PR #1366, opened 2024-09-16
fix(Thread): parentMessage delete & initial read object

arnautov-anton:
### This PR solves three issues:

- adds handler for the `reaction.updated` event
- adds handler for deletion of the parent message (and itself)
- adds "patchy" placeholder `read` object with current user from which the `unreadMessageCount` can be incremented on `message.new` 

### Noteworthy changes:

- `generateThread` function has been renamed to `generateThreadResponse`
- `updateParentMessageLocally` takes object with `message` property to align with the rest of the methods

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -4,6 +4,7 @@ import { StateStore } from './store';
 import type {
   AscDesc,
   DefaultGenerics,
+  EventTypes,
   ExtendableGenerics,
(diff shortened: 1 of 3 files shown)


isekovanic (approved):
Looks cool !

oliverlaz (approved):

PR #1370, opened 2024-09-19
fix: replace binarySearchByDateEqualOrNearestGreater with findIndexInSortedArray \w midpoint match

arnautov-anton:
This PR removes and replaces `binarySearchByDateEqualOrNearestGreater` with universal `findIndexInSortedArray` function and enables exact matching by [checking midpoint](https://github.com/GetStream/stream-chat-js/pull/1370/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R356). 

Changes to `channel` and `channel_state` tests are to ensure proper order of generated messages (added `created_at` to each generated message with proper offset).

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -310,12 +310,18 @@ export function formatMessage<StreamChatGenerics extends ExtendableGenerics = De
 
 export const findIndexInSortedArray = <T, L>({
   needle,
+  returnOnMidMatch,
   sortedArray,
   selectValueToCompare = (e) => e,
   sortDirection = 'ascending',
 }: {
   needle: T;
   sortedArray: readonly T[];
+  /**
+   * Returns the index of the midpoint if it matches the target value.
+   * Should be enabled only if the searched array cannot contain duplicates.
+   */
+  returnOnMidMatch?: boolean;
   /**
    * In array of objects (like messages), pick a specific
    * property to compare needle value to.
@@ -353,7 +359,7 @@ export const findIndexInSortedArray = <T, L>({
 
     const comparableMiddle = selectValueToCompare(sortedArray[middle]);
 
-    if (comparableNeedle === comparableMiddle) return middle;
+    if (returnOnMidMatch && comparableNeedle === comparableMiddle) return middle;
 
     if (
       (sortDirection === 'ascending' && comparableNeedle < comparableMiddle) ||
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -353,7 +353,7 @@ export const findIndexInSortedArray = <T, L>({
 
     const comparableMiddle = selectValueToCompare(sortedArray[middle]);
 
-    // if (comparableNeedle === comparableMiddle) return middle;
+    if (comparableNeedle === comparableMiddle) return middle;
 
     if (
       (sortDirection === 'ascending' && comparableNeedle < comparableMiddle) ||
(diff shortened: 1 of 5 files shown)


test/unit/channel_state.js
```diff
@@ -110,8 +110,8 @@ describe('ChannelState addMessagesSorted', function () {
 		expect(state.messages).to.have.length(10);
 		state.addMessagesSorted([generateMsg({ id: 'id', date: `2020-01-01T00:00:00.007Z` })]);
 		expect(state.messages).to.have.length(11);
-		expect(state.messages[7].id).to.be.equal('7');
-		expect(state.messages[8].id).to.be.equal('id');
+		expect(state.messages[7].id).to.be.equal('id');
+		expect(state.messages[8].id).to.be.equal('7');
```
arnautov-anton, line 114: One hurdle of the midpoint check addition is that message with the same `created_at` will be added before existing one but since the existence of the messages with the same `created_at` in real-world applications is *almost* impossible - I adjusted the test. 

Even without midpoint check, the previous solution - in case of multiple messages with the same `created_at` - the new message would be placed at most one index after the already existing one which is still incorrect.

test/unit/utils.js
```diff
@@ -2690,23 +2690,3 @@ describe('messageSetPagination', () => {
...
-		{ created_at: '2024-08-05T08:55:07.199808Z', id: '7' },
-		{ created_at: '2024-08-05T08:55:08.199808Z', id: '8' },
-	];
-	it('finds the nearest newer item', () => {
-		expect(binarySearchByDateEqualOrNearestGreater(messages, new Date('2024-08-05T08:55:02.299808Z'))).to.eql(3);
-	});
-	it('finds the nearest matching item', () => {
-		expect(binarySearchByDateEqualOrNearestGreater(messages, new Date('2024-08-05T08:55:07.199808Z'))).to.eql(7);
-	});
-});
```
arnautov-anton, line 2712: Moved to `utils.test.ts` with additions.

test/unit/utils.test.ts
```diff
@@ -93,15 +93,91 @@ describe('addToMessageList', () => {
...
+    expect(messagesBefore[3].text).to.not.equal(newMessage.text);
+    expect(messagesBefore[3]).to.not.equal(newMessage);
 
-    const messagesAfter = addToMessageList(messagesBefore, newMessage, false, 'created_at', false);
+    const messagesAfter = addToMessageList(messagesBefore, newMessage, false);
 
     expect(messagesAfter).to.have.length(5);
-    expect(messagesAfter[4]).to.equal(newMessage);
+    expect(messagesAfter[3]).to.equal(newMessage);
+  });
```
arnautov-anton, line 106: I messed up here previously.

src/utils.ts
```diff
@@ -310,12 +310,18 @@ export function formatMessage<StreamChatGenerics extends ExtendableGenerics = De
...
   selectValueToCompare = (e) => e,
   sortDirection = 'ascending',
 }: {
   needle: T;
   sortedArray: readonly T[];
+  /**
+   * Returns the index of the midpoint if it matches the target value.
+   * Should be enabled only if the searched array cannot contain duplicates.
+   */
+  returnOnMidMatch?: boolean;
```
MartinCupela, line 324:
```suggestion
  returnOnFirstMatch?: boolean;
```

MartinCupela (approved):

PR #1372, opened 2024-09-27
fix: thread manager bugs

isekovanic:
## CLA

- [x] Code changes are tested

This PR fixes 2 leftover issues in Threads v2:

- `ThreadManager` state being stale after logging out in an app
- Channel deletion not having a realtime effect on `ThreadManager` state

## Changelog

-


src/client.ts
```diff
@@ -808,6 +808,8 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     this.activeChannels = {};
     // reset client state
     this.state = new ClientState();
+    // reset thread manager
+    this.threads = new ThreadManager({ client: this });
```
arnautov-anton, line 812: This feels "leaky" and might cause issues because we're mutating the manager in place, appropriate components might not register this change until next re-render. The active threads within will stay connected to the dispatcher and will keep listening and updating their states.

I'd rather do something like: 
```suggestion
    this.threads.state.next(INITIAL_THREAD_MANAGER_STATE) // default would need to be created
    // or with new API
    this.threads.resetLocalState(); // or initializeLocalState() - not sure about naming, I'd let Matvei in on this, he's good at these kinds of things :)
```

src/thread_manager.ts
```diff
@@ -117,6 +118,19 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
     return () => unsubscribeFunctions.forEach((unsubscribe) => unsubscribe());
   };
 
+  private subscribeChannelDeleted = () => {
+    const unsubscribeFunctions = ['notification.channel_deleted', 'channel.deleted'].map((e) => {
```
arnautov-anton, line 122:
```suggestion
    const unsubscribeFunctions = ['notification.channel_deleted'].map((e) => {
```
One should suffice, notification-type events are always delivered (with less information) but in this case it's fine. With two in the array we're going to run the callback twice.

src/thread_manager.ts
```diff
@@ -117,6 +118,19 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+      return this.client.on(e, (event) => {
+        const { cid } = event;
+        const { threads } = this.state.getLatestValue();
+
+        const newThreads = threads.filter((t) => t.channel.cid !== cid);
```
arnautov-anton, line 127: nit:
```suggestion
        const newThreads = threads.filter((thread) => thread.channel.cid !== cid);
```

src/thread_manager.ts
```diff
@@ -7,6 +7,20 @@ import type { DefaultGenerics, Event, ExtendableGenerics, OwnUserResponse, Query
 
 const DEFAULT_CONNECTION_RECOVERY_THROTTLE_DURATION = 1000;
 const MAX_QUERY_THREADS_LIMIT = 25;
+export const INITIAL_STATE = {
```
arnautov-anton, line 10:
```suggestion
const INITIAL_STATE = {
```
I'd keep it private for now (name collision could be also an issue).

isekovanic: Yeah, it was my initial instinct as well but I wanted to actually compare against this in the tests to make sure we have no leftovers - I could do the check manually but then I think the test would be too easy to break if we add additional stuff to the state at some point. How do you propose we proceed ?

src/thread_manager.ts
```diff
@@ -76,6 +77,11 @@ export class ThreadManager<SCG extends ExtendableGenerics = DefaultGenerics> {
     return threadsById;
   }
 
+  public resetState = () => {
+    this.unregisterSubscriptions();
```
arnautov-anton, line 81:
```suggestion
    this.unregisterSubscriptions();
```
Hmmm - good point but now I'm conflicted. I strongly believe that this should happen within the SDK, not here as it's an unexpected side-effect an SDK might not know nothing about. I'm not even sure whether it's even needed to disable the subscriptions here?

isekovanic: Well - if that's the case I'm worried of the fact that they'll stay registered while they really shouldn't, ultimately resulting in double the subscriptions when you log in with another user (and this would keep going on as many times you `connect`/`disconnect`). Whenever the list of `threads` changes within the `ThreadManager`, we anyway unsubscribe from the specific `thread` listeners - so it's probably a good idea to flush everything out on the `ThreadManager` level as well.

But you do have a good point of some unexpected side-effects, anything specific you have in mind ?

src/thread_manager.ts
arnautov-anton: Ah -  then `THREAD_MANAGER_INITIAL_STATE` should do; it's a mouthfull but works well. :)

myandrienko (approved):

MartinCupela (approved):

src/thread_manager.ts
myandrienko: Hey @arnautov-anton, notification events are not delivered for watched channels AFAIK. So we need to keep both.

arnautov-anton: Just tested it just to make sure and I receive both notification and normal event even when the channel is watched though funny enough, I receive two notification events - I guess just to make 100% sure we handle it. :)

image

PR #1373, opened 2024-10-02
feat: polls rewrite

MartinCupela:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/poll.ts
+++ b/src/poll.ts
@@ -208,10 +208,7 @@ export class Poll<SCG extends ExtendableGenerics = DefaultGenerics> {
 
     if (isOwnVote) {
       if (isVoteAnswer(event.poll_vote)) {
-        latestAnswers = [
-          event.poll_vote,
-          ...latestAnswers.filter((answer) => answer.id !== event.poll_vote.id),
-        ];
+        latestAnswers = [event.poll_vote, ...latestAnswers.filter((answer) => answer.id !== event.poll_vote.id)];
         ownAnswer = event.poll_vote;
       } else if (event.poll_vote.option_id) {
         if (event.poll.enforce_unique_votes) {


src/poll_manager.ts
```diff
@@ -0,0 +1,149 @@
...
+    }).unsubscribe;
+  };
+
+  private subscribeVoteRemoved = () => {
+    return this.client.on('poll.vote_removed', (event) => {
+      if (event.poll?.id) {
+        this.fromState(event.poll.id)?.handleVoteRemoved(event);
+      }
+    }).unsubscribe;
+  };
```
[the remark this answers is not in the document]
arnautov-anton: (including `subscribePollUpdated`)

src/poll_manager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  constructor({ client }: { client: StreamChat<SCG> }) {
+    this.client = client;
+  }
+
+  get data(): Map<string, Poll<SCG>> {
+    return this.pollCache;
+  }
+
+  public fromState = (id: string) => {
```
arnautov-anton, line 28:
```suggestion
  public fromCache = (id: string) => {
```

src/poll_manager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  public getPoll = async (id: string) => {
+    const { poll } = await this.client.getPoll(id);
+
+    this.setOrOverwriteInCache(poll, true);
+
+    return this.fromState(id);
+  };
+
+  public queryPolls = async (filter: QueryPollsFilters, sort: PollSort = [], options: QueryPollsOptions = {}) => {
```
arnautov-anton, line 65: `StreamChat.queryThreads` returns list of `Thread` instances - maybe we could do the same here for general API consistency?

The same goes for state [hydration](https://github.com/GetStream/stream-chat-js/blob/71153c8818347e3a7780d57f1df67f75e6f27a0b/src/thread.ts#L146):

```tsx
poll.hydrateState(differentPollInstanceButSameId)
```

src/poll_manager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    return new Poll({ client: this.client, poll: createdPoll });
+  };
+
+  public getPoll = async (id: string) => {
+    const { poll } = await this.client.getPoll(id);
+
+    this.setOrOverwriteInCache(poll, true);
+
+    return this.fromState(id);
```
arnautov-anton, line 62: Why are we reaching for the request first instead of cache - I'd expect we'd be reaching for request in case of cache-miss?

src/poll_manager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  public unregisterSubscriptions = () => {
+    this.unsubscribeFunctions.forEach((cleanupFunction) => cleanupFunction());
+    this.unsubscribeFunctions.clear();
+  };
+
+  public createPoll = async (poll: CreatePollData<SCG>) => {
+    const { poll: createdPoll } = await this.client.createPoll(poll);
+
+    return new Poll({ client: this.client, poll: createdPoll });
```
arnautov-anton, line 54: Why is the newly created poll not being cached?

isekovanic: It's not since presence in the cache is directly tied to the poll actually being sent as a message somewhere (i.e appearing when we `queryChannels`, or a `message.new` event arrives with a `poll` inside of it etc). But it might be a good point to perhaps not even create a poll instance, since in this case we're really only interested in the ID of the poll that got created.

src/poll_manager.ts
isekovanic: Because we've decided that `queryPolls` and `getPoll` are always going to reinitialize the state within the cache. We want them to be used independently from the rest and be able to overwrite whatever's in the cache (not by reference though, only state) when they're queried. But, perhaps we can simply always use `getPoll` and provide a fallback for whenever it's not in the cache I suppose. Wanted to keep those 2 separate though.

src/poll_manager.ts
isekovanic: Well, `client.queryPolls` should return the raw response in my opinion in case someone needs it (they'll have a way to do it). If they're reaching for reactive state though, they would fetch through `pollManager.queryPolls(...)` - this was the general idea. We can of course move the instantiation there too and simply make sure the cache is populated in `pollManager`, but it seemed like losing an API that way.

src/poll_manager.ts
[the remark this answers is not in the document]
isekovanic: That's a no-go, it was indeed our initial approach however there are a few issues with it:

- Having `3000` polls essentially means having `3000 * 5` listeners at a certain point, all of which would be invoked when a new event arrives (and `2999` of them would simply skip because the event does not concern them)
- It doesn't particularly scale well with more polls being loaded through pagination
- This way, we only have a single set of listeners that consume one event at a time, updating the correct poll from within the cache directly (while maintaining referential integrity of the poll instances at all times)

So, in other words - they are indeed handled in the `Poll` instances themselves it's just not the `Poll` instances that listen to the events but rather the `manager` does.

src/poll_manager.ts
isekovanic: Deliberated quite a bit over this as well, but in the end I decided to go with `fromState` since the `cache` is the manager's state in some sense. It feels kind of wrong exposing something called `fromCache` to integrators since it might lead them to believe it's something else, but I might just be overthinking this :D

src/poll_manager.ts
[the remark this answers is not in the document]
arnautov-anton: So esentially - poll instance cannot properly live without the manager - correct? You always have to rely on manager to keep your poll instance up to date so "detached" instances are "broken" in a way. 

I think it'd be good if we explored the idea of targeted events deeper - it should be very simple to do for polls as most of the events are poll-related and carry message/poll id's and we could benefit from this performance enhancement in the future. Poll could be fully encapsulated with logic it should "own" allowing it also to live in a detached (from the manager) state.

src/poll_manager.ts
arnautov-anton: I'm worried that this might be a bit misleading either way as integrators might think that manager has a state which they can subscribe to as the API becomes the norm across the client though I understand your point too.

src/poll_manager.ts
arnautov-anton: There shouldn't be a need for it though, no? If the poll is in the cache then that means it's been receiving updates through event subscriptions meaning its state should closely represent the one that backend has - hydrating it at this point seems excessive.

The other way would be to return it right away and do the state hydration behind the scenes - since the state of the poll is reactive, it should not matter that it changes after it gets returned, no?

src/poll_manager.ts
isekovanic: Maybe something like `byId` would also be good ? Not implying anything specifically that way, we're just saying that we're asking for a `poll` with a certain ID. What do you think ?

src/poll_manager.ts
isekovanic: There is a particular need for this for offline storage in RN for example. Let us consider the following scenario:

- You create a `poll` and are actively watching its progress (some votes, answers come in - w/e)
- You close the app and do something else, while in the meantime more votes and/or answers arrive
- You come back, and the first "optimistic" thing that loads is the data from the local DB (which is now out of date since those WS events never got consumed while you weren't using the app)
- If you're offline, this is no biggie since we don't expect that WS events are handled in the background
- If you're online however, the point at which a query is made for all of the channels (and their messages) you'll get all of the updated data; and it should indeed overwrite the local DB data (and store this agaiån)

Unfortunately, the `sync` API does not yet support `poll` events and so we can't particularly rely on it to do the heavy lifting for us. Whenever these potentially get added in the future, perhaps we can remove the overwriting mechanism (it's anyway hidden behind the second argument). For now, however - I believe we need it.

If it seems sketchy to you maybe we can lift the overwriting condition higher so that we can specifically trigger this only for the RN SDK (since the rest don't have offline support) ?

src/poll_manager.ts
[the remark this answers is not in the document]
isekovanic: If I understand you correctly, you're looking for something like:

- The `pollManager` is still listening for `poll` events
- It re-dispatches a targeted event (for example, `poll.vote_casted-<poll_id>` or something`)
- The poll itself is only listening to `poll.vote_casted-<this.id>` events and handles them

Is this correct ? I totally get your concerns and something like this was briefly considered too, however I do have some concerns about this approach too:

- Each `poll` having its own listener can potentially be a ton of extra listeners
- Not every listener would indeed be triggered (only the specific ones), but their event's `dispatch` would still be coupled with the `poll` manager, which would act as a `router` of sorts in order to orchestrate the traffic - doing these changes through events seems to me like a bit of an overhead considering that we can just update inline
- If `poll` instances are indeed to listen, they have to be subscribed to events; since we need all of these to be up to date at all times we would basically need to subscribe to the new events once again whenever `messages` actually arrive within the state `client.queryChannels` or `channel.query` for example)

With that said, I do see some really good things with this as well:

- The ability to hook up to specific `poll` events from anywhere (for example, if some custom integrator logic depends on events they can simply hook up to `poll.vote_casted-<id>` and not have to do an ID sanity check each time one of these events is fired)
- Having them detached is a very, very good point too

So, I'm not particularly sure which way we should go - perhaps we should sit down and discuss this more thoroughly. Adding targeted events shouldn't be difficult as far as I can see (pretty much the same logic).

PR #1382, opened 2024-10-25
fix: change subscribeWithSelector API declaration

isekovanic:
## CLA

As discussed with @oliverlaz and @arnautov-anton , the current API of the state store seems very brittle for integrators and something that is very prone to errors (that we cannot easily catch). Because of that, we've decided to change the state store's API so that it no longer returns an array of sorted properties, but rather an object with named properties. 

The way the API changes is as follows:

Previously: 
```
const selector = (nextValue: SomeState) => [nextValue.nameFromState] as const;
```

```
const selector = (nextValue: SomeState) => ({ explicitName: nextValue.nameFromState } as const);
```

and then use it by:

```
const { explicitName } = { ... }
```

This way, we will never be completely dependant on keeping the order of the array proper and not have to worry about adding new properties to maintain the order. 

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -36,13 +36,23 @@ export class StateStore<T extends Record<string, unknown>> {
     };
   };
 
-  public subscribeWithSelector = <O extends readonly unknown[]>(selector: (nextValue: T) => O, handler: Handler<O>) => {
+  public subscribeWithSelector = <O extends Readonly<Record<string, unknown>> & { length?: never }>(
+    selector: (nextValue: T) => O,
+    handler: Handler<O>,
+  ) => {
     // begin with undefined to reduce amount of selector calls
     let selectedValues: O | undefined;
 
     const wrappedHandler: Handler<T> = (nextValue) => {
       const newlySelectedValues = selector(nextValue);
-      const hasUpdatedValues = selectedValues?.some((value, index) => value !== newlySelectedValues[index]) ?? true;
+
+      let hasUpdatedValues = !selectedValues;
+
+      for (const key in selectedValues) {
+        if (selectedValues[key] === newlySelectedValues[key]) continue;
+        hasUpdatedValues = true;
+        break;
+      }
 
       if (!hasUpdatedValues) return;
 
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -197,8 +197,11 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
 
   private subscribeMarkActiveThreadRead = () => {
     return this.state.subscribeWithSelector(
-      (nextValue) => [nextValue.active, ownUnreadCountSelector(this.client.userID)(nextValue)],
-      ([active, unreadMessageCount]) => {
+      (nextValue) => ({
+        active: nextValue.active,
+        unreadMessageCount: ownUnreadCountSelector(this.client.userID)(nextValue),
+      }),
+      ({ active, unreadMessageCount }) => {
         if (!active || !unreadMessageCount) return;
         this.throttledMarkAsRead();
       },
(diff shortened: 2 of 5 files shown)


src/store.ts
```diff
@@ -36,13 +36,29 @@ export class StateStore<T extends Record<string, unknown>> {
     };
   };
 
-  public subscribeWithSelector = <O extends readonly unknown[]>(selector: (nextValue: T) => O, handler: Handler<O>) => {
+  public subscribeWithSelector = <O extends Readonly<Record<string, unknown>> & { length?: never }>(
```
arnautov-anton, line 39:
```suggestion
  public subscribeWithSelector = <O extends Readonly<Record<string, unknown>>>(
```
Let's not keep this here; it was fine as a temporary measure (to easily find all the selectors and change outputs to objects).

isekovanic: I thought about removing it, but I'd actually rather keep it so that TS complains on the integrator's side as well (it won't otherwise). They'll get the warning, sure - but it won't "force" them to change; if they use TS this rule would.

isekovanic: Nevermind, the SDKs will anyway be complaining about this. Will remove it, thanks.

src/store.ts
```diff
@@ -36,13 +36,29 @@ export class StateStore<T extends Record<string, unknown>> {
...
+        console.warn(
+          'The API of our state store has changed. Instead of returning an array in the selector, please return a named object of properties.',
+        );
+      }
+
+      for (const key in selectedValues) {
+        if (selectedValues[key] === newlySelectedValues[key]) continue;
+        hasUpdatedValues = true;
+        break;
+      }
```
oliverlaz, line 61: Do we need this optimization? AFAIK, this comparison is broken, as the order of keys in the object isn't guaranteed.
This code will say that `{ a: 10, b: 20 }` is not equal to ` { b: 20, a: 10 }`.

@arnautov-anton, @isekovanic

isekovanic: Hmm, but the order doesn't matter right ? We're just really looking for the keys and then extracting them from both places. We're just comparing the values for the specific key - we're not saying that `selectedValues[selectedValueKey]` needs to be equal to `newlySelectedValues[newlySelectedValuesKey]`.

isekovanic: The only time this might fail would be if `selectedValues` and `newlySelectedValues` have different keys, which would never be the case because they're handled entirely by how the `selector` is defined. Or am I missing something ?

oliverlaz: Ignore me! You are completely right. Maybe loop over `newlySelectedValues` to make sure there aren't keys added or removed.

arnautov-anton: Correct, the store is large and can hold multitude of other values that are selected but did not change - each selection produces new object and that object is being passed to `setState` - to prevent running `setState` each time unrelated change happens (value that changed has not been selected), we bail by comparing actually selected values.

isekovanic: But, if you guys want I can also add something like `Object.keys(selectedValues).length === Object.keys(newlySelectedValues).length` just to be safe. We didn't have that check before either and I agree with @arnautov-anton that it would never happen, but maybe we should be safe.

isekovanic: That check should be enough because if their sizes differ then they're by default different and an update should be triggered. If they're the same, only then do we check if all of the values are the same for the respective keys (if some keys differ it would be caught here).

arnautov-anton: Removing or adding keys is obviously not recommended - it'll just trigger a re-render as either previous or newest wouldn't have the keys thus `if (selectedValues[key] === newlySelectedValues[key]) continue;` is no longer true.

arnautov-anton: We can test just to be sure but I don't think adding an extra check is necessary.

isekovanic: Yep agreed. I don't think we need to do it.

src/store.ts
```diff
@@ -36,13 +36,29 @@ export class StateStore<T extends Record<string, unknown>> {
...
     // begin with undefined to reduce amount of selector calls
     let selectedValues: O | undefined;
 
     const wrappedHandler: Handler<T> = (nextValue) => {
       const newlySelectedValues = selector(nextValue);
-      const hasUpdatedValues = selectedValues?.some((value, index) => value !== newlySelectedValues[index]) ?? true;
+
+      let hasUpdatedValues = !selectedValues;
+
+      if (selectedValues && Array.isArray(selectedValues)) {
```
isekovanic, line 51: Added a truthy check here too, since `Array.isArray(undefined` is `false`.

PR #1412, opened 2024-12-07
fix(search): missing thread_participants in message

kanat:
## Link to issue

- https://linear.app/stream/issue/CHA-365/missing-thread-participants-data-in-streamio-query

## Related PR

- https://github.com/GetStream/chat/pull/7674

## Summary

Enable the `Channel.search` to enrich `thread_participants` in found messages.

## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1713,6 +1713,10 @@ export type MessageFilters<StreamChatGenerics extends ExtendableGenerics = Defau
     }
 >;
 
+export type MessageOptions = {
+  include_thread_participants?: boolean;
+};
+
 export type PrimitiveFilter<ObjectType> = ObjectType | null;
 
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):

src/types.ts
```diff
@@ -1713,6 +1713,10 @@ export type MessageFilters<StreamChatGenerics extends ExtendableGenerics = Defau
     }
 >;
 
+export type MessageOptions = {
```
MartinCupela, line 1716: I think that only `MessageOptions` would not be clear enough to convey the fact that these are actually search payload options.

```suggestion
export type SearchMessageOptions = {
```

kanat: Yeah I though about it as well, however, technically, in the future, it could be used in other places where the `MessageFilters` is also used.

PR #1421, opened 2024-12-17
fix: message duplication when some messages have the same creation timestamp

myandrienko:
To reproduce an existing bug:
1. Have several (3+) messages with the same creation timestamp in a channel (you can send them quickly with the server-side API)
2. Set up a situation when channels are queried twice (e.g. `<StrictMode />` is enabled and the effect in `ChannelList` runs twice)
3. Observe duplicating messages

The issue is that when inserting items into a sorted array with duplicating creation date, the insertion index points next to a fairly random message with the same creation date. That is not necessarily the message with same id, it could be any other message with the same creation id. So, instead of updating an existing message, we add it again.

This is fixed by performing an additional look-around when inserting a message, searching for the message with the same creation date and the same id. It adds very little overhead, since the search range is very limited. If the message with the same key is found, it is returned as an insertion index.

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -354,6 +354,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
       sortedArray: replies,
       sortDirection: 'ascending',
       selectValueToCompare: (reply) => reply.created_at.getTime(),
+      selectKey: (reply) => reply.id,
     });
 
     const actualIndex =
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -311,6 +311,7 @@ export function formatMessage<StreamChatGenerics extends ExtendableGenerics = De
 export const findIndexInSortedArray = <T, L>({
   needle,
   sortedArray,
+  selectKey,
   selectValueToCompare = (e) => e,
   sortDirection = 'ascending',
 }: {
@@ -325,6 +326,7 @@ export const findIndexInSortedArray = <T, L>({
    * selectValueToCompare: (message) => message.created_at.getTime()
    * ```
    */
+  selectKey?: (arrayElement: T) => string;
   selectValueToCompare?: (arrayElement: T) => L | T;
   /**
    * @default ascending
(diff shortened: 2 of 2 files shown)


src/utils.ts
```diff
@@ -353,18 +355,28 @@ export const findIndexInSortedArray = <T, L>({
...
+  // point to possibly find an element with the same key. If found, prefer it.
+  // This, for example, prevents duplication of messages with the same creation date.
+  if (selectKey) {
+    const needleKey = selectKey(needle);
+    for (let i = left; sortedArray[i] === comparableNeedle; i += sortDirection === 'ascending' ? -1 : +1) {
+      if (selectKey(sortedArray[i]) === needleKey) {
+        return i;
+      }
+    }
+  }
```
myandrienko, line 378: 🍒

src/utils.ts
```diff
@@ -410,19 +422,18 @@ export function addToMessageList<T extends FormatMessageResponse>(
...
-  if (!timestampChanged && newMessage.id) {
-    if (newMessages[insertionIndex] && newMessage.id === newMessages[insertionIndex].id) {
-      newMessages[insertionIndex] = newMessage;
-      return newMessages;
-    }
-
-    if (newMessages[insertionIndex - 1] && newMessage.id === newMessages[insertionIndex - 1].id) {
-      newMessages[insertionIndex - 1] = newMessage;
-      return newMessages;
-    }
```
myandrienko, line 425: We no longer need to check this case: if a message needs to be updated, `insertionIndex` is guaranteed to point at it.

src/utils.ts
```diff
@@ -325,6 +326,7 @@ export const findIndexInSortedArray = <T, L>({
    * selectValueToCompare: (message) => message.created_at.getTime()
    * ```
    */
+  selectKey?: (arrayElement: T) => string;
```
arnautov-anton, line 329: Could you please add a description for this property?

myandrienko: Added in https://github.com/GetStream/stream-chat-js/pull/1421/commits/41275d9bef77bcac6caec589031d11932dc2c6d6.

src/thread.ts
```diff
@@ -354,6 +354,7 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
       sortedArray: replies,
       sortDirection: 'ascending',
       selectValueToCompare: (reply) => reply.created_at.getTime(),
+      selectKey: (reply) => reply.id,
     });
 
     const actualIndex =
```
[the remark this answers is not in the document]
myandrienko: Good point, fixed in https://github.com/GetStream/stream-chat-js/pull/1421/commits/51ed4f8bfd5f2a3ea1adc478d4d84aa2cc0284f9.

src/thread.ts
```diff
@@ -354,17 +354,15 @@ export class Thread<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+      selectKey: (reply) => reply.id,
     });
 
-    const actualIndex =
-      replies[index]?.id === message.id ? index : replies[index - 1]?.id === message.id ? index - 1 : null;
-
-    if (actualIndex === null) {
+    if (replies[index]?.id !== message.id ? index : null) {
       return;
     }
```
arnautov-anton, line 362: I think we can ditch this whole thing, no?

myandrienko: I don't think so, the index returned by `findIndexInSortedArray` is not necessarily pointing at an exact match.

arnautov-anton: The confusing part is the rest of the condition (`? index : null`) - in this case it'd go to the early return if the message id's match and the index is non-zero. 🤔

myandrienko: Oh man, for real. Missed that :)

myandrienko: Fixed in https://github.com/GetStream/stream-chat-js/pull/1421/commits/17f8269c543266c59939915df0fa8849995bb122.

MartinCupela (approved):

PR #1423, opened 2024-12-19
feat: add static location and live location support

arnautov-anton:
### This PR adds

- `LocationComposer` class - allows to compose a message with a shared_location
- local events `live_location_sharing.started`, `live_location_sharing.stopped`
- `Channel.sendSharedLocation` - sends a message with a new `shared_location` field
- `Channel.stopLiveLocationSharing` - dispatches local `live_location_sharing.stopped` event to notify `LiveLocationManager` instances (should probably just update message attachment with `stopped_sharing` instead)
- `LiveLocationManager` class - reports location (`client.updateSharedLocation`) updates to the server for all the messages
- location related types


src/LiveLocationManager.ts
```diff
@@ -0,0 +1,256 @@
...
+
+      withCancellation(LiveLocationManager.symbol, async () => {
+        const promises: Promise<SharedLocationResponse>[] = [];
+        await this.assureStateInit();
+
+        for (const [messageId, location] of this.messages) {
+          if (location.latitude === latitude && location.longitude === longitude)
+            continue;
+          const promise = this.client.updateLocation({
+            created_by_device_id: this.deviceId,
```
khushal87: If the `this.deviceId` is passed here the `updateLocation` API fails with an error as below. I think it should be `location.created_by_device_id`.

Screenshot 2025-07-18 at 12 29 55 PM

MartinCupela: Yes, you are right Khushal, this should be taken from the location object.

src/LiveLocationManager.ts
```diff
@@ -0,0 +1,267 @@
...
+  MessageResponse,
+  SharedLiveLocationResponse,
+  SharedLocationResponse,
+} from './types';
+import type { Coords } from './messageComposer';
+
+export type WatchLocationHandler = (value: Coords) => void;
+
+// type Unsubscribe = () => void;
+type WatchLocation = (handler: WatchLocationHandler) => Unsubscribe;
```
khushal87: Can we export it as well?

MartinCupela: done

MartinCupela:
> I am unable to send a location message in a thread. Is it in the spec? I see there's an early exit in the code for `sendLocation`

I think this stems from the spec API definition - there is no place for thread id info in the methods signatures

khushal87 (approved):

PR #1428, opened 2025-01-05
feat(threads): handle custom data

arnautov-anton:

- ~allow querying thread-related data as pure objects instead of instances of `Thread` class~ (discarded)
- handle `thread.updated` within `Thread` class and consider custom fields

### Custom fields

`Thread` class now handles custom data by sifting through `ThreadResponse` keys, skipping reserved ones and storing custom keys with associated data within `ThreadState.custom` object.  For type support, declaration file has to be created and `ThreadResponseCustomData` has to be extended:

```ts
// yourFile.d.ts

import 'stream-chat';

declare module 'stream-chat' {
  interface ThreadResponseCustomData {
    myCustomProperty: string;
    myOtherCustomProperty: { meta: Array<{ key: number }> };
  }
}
```

When the `Thread` is instantiated, the custom properties can be accessed from the state:

```ts
const thread = await client.getThread('messageId');

const { custom } = thread.state.getLatestValue();

console.log(custom.myCustomProperty, custom.myOtherCustomProperty);
```

Fixes: GetStream/stream-chat-react#2592

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2754,20 +2754,10 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
    * @param {boolean} options.watch Subscribes the user to the channels of the threads.
    * @param {number}  options.participant_limit Limits the number of participants returned per threads.
    * @param {number}  options.reply_limit Limits the number of replies returned per threads.
-   * @param {boolean} asJSON If set to true, returns JSON response instead of array of Thread instances.
    *
    * @returns {{ threads: Thread<StreamChatGenerics>[], next: string }} Returns the list of threads and the next cursor.
    */
-
-  async queryThreads(
-    options: QueryThreadsOptions | undefined,
-    asJSON: true,
-  ): Promise<QueryThreadsAPIResponse<StreamChatGenerics>>;
-  async queryThreads(
-    options?: QueryThreadsOptions,
-    asJSON?: false,
-  ): Promise<{ next: string | undefined; threads: Thread<StreamChatGenerics>[] }>;
-  async queryThreads(options: QueryThreadsOptions = {}, asJSON = false) {
+  async queryThreads(options: QueryThreadsOptions = {}) {
     const optionsWithDefaults = {
       limit: 10,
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -2751,23 +2754,42 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+   * @param {boolean} asJSON If set to true, returns JSON response instead of array of Thread instances.
    *
    * @returns {{ threads: Thread<StreamChatGenerics>[], next: string }} Returns the list of threads and the next cursor.
    */
-  async queryThreads(options?: QueryThreadsOptions) {
-    const opts = {
+
+  async queryThreads(
+    options: QueryThreadsOptions | undefined,
+    asJSON: true,
```
isekovanic, line 2764: *nitpick*
```suggestion
    asObject: true,
```
maybe something like this ? This is technically not JSON, so perhaps more fitting. Totally not important, just caught my eye

arnautov-anton: I'll be dropping these changes altogether - I don't think we should support this at all.

MartinCupela (approved):

szuperaz (approved):

src/thread.ts
```diff
@@ -68,19 +68,42 @@ export type ThreadReadState<SCG extends ExtendableGenerics = DefaultGenerics> =
...
 // TODO: remove this once we move to API v2
-const constructCustomDataObject = <SCG extends ExtendableGenerics>(
-  customPropertyKeys: (keyof ThreadResponseCustomData)[],
-  threadData: ThreadResponse<SCG>,
-) =>
-  customPropertyKeys.reduce<ThreadResponseCustomData>((custom, key) => {
-    if (threadData[key]) {
-      custom[key] = threadData[key];
+const constructCustomDataObject = <SCG extends ExtendableGenerics>(threadData: ThreadResponse<SCG>) => {
+  const custom: ThreadResponseCustomData = {};
```
MartinCupela, line 93: Where is `ThreadResponseCustomData` imported from?

PR #1434, opened 2025-01-16
feat: add SearchController

MartinCupela:

This is an addition of a new classes that control the search logic.

1. SearchController

- Presents a public API for coordination of search activities btw search sources.
- Is not aware of what data is being searched
- Delegates the data retrieval logic to search sources
- Has own reactive state used by integrators in the UI components

2. Search sources
- Take care of retrieving the searched information from the back-end
- Contain data-specific logic
- Handle pagination
- Have own reactive state related to pagination over search results

Advantages:
- Data retrieval logic and search parameters are fully customizable by integrators
- Integrators can subscribe to data changes of the reactive states => no need for onEvent props in the UI components


test/unit/search_controller.test.js
```diff
@@ -0,0 +1,510 @@
+import { expect } from 'chai';
```
arnautov-anton, line 1: Not sure if this is still WIP but let's test new things with TS too.

MartinCupela: @arnautov-anton Does it make sense to have ts declarations if all would be mocks?

arnautov-anton: I mean I saw that the controller takes some generic parameter so it'd be good to test that too. Also tests if functions which rely on this generic accept specific parameter types correctly and what not. Having fully-typed tests just adds extra level of safety even if fully or partially mocked - meaning changing some types after the fact should be caught too.

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+  search(text?: string): void;
+  searchDebounced: DebouncedExecQueryFunction;
+  readonly searchQuery: string;
+  setDebounceOptions(options: DebounceOptions): void;
+  readonly state: StateStore<SearchSourceState<T>>;
+  readonly type: SearchSourceType;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type SearchSourceState<T = any> = {
```
arnautov-anton, line 50:
```suggestion
export type SearchSourceState<T> = {
```

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+
+export type SearchSourceType = 'channels' | 'users' | 'messages' | (string & {});
+export type QueryReturnValue<T> = { items: T[]; next?: string };
+export type DebounceOptions = {
+  debounceMs: number;
+};
+type DebouncedExecQueryFunction = DebouncedFunc<(searchString?: string) => Promise<void>>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface SearchSource<T = any> {
```
arnautov-anton, line 29:
```suggestion
export interface SearchSource<T> {
```

test/unit/search_controller.test.js
MartinCupela: I still am not sure, whether TS version would make sense. All the search source classes are based on a single abstract class that does not care about the generic type - there is no impact on logic. Could you give me an example please?

src/search_controller.ts
MartinCupela: I thought that putting the default `any` would be simpler than adding it to all the mentions in the SearchController.

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+  readonly state: StateStore<SearchSourceState<T>>;
+  readonly type: SearchSourceType;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type SearchSourceState<T = any> = {
+  hasMore: boolean;
+  isActive: boolean;
+  isLoading: boolean;
+  items: T[] | undefined;
```
arnautov-anton, line 54:
```suggestion
  items: T[] | undefined;
```
Why `undefined`? Can't it be an empty array instead?

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+
+  setDebounceOptions = ({ debounceMs }: DebounceOptions) => {
+    this.searchDebounced = debounce(this.executeQuery.bind(this), debounceMs);
+  };
+
+  activate = (sourceStateOverride?: Partial<SearchSourceState>) => {
+    if (this.isActive) return;
+    if (this.searchQuery) {
+      this.search();
+    }
```
arnautov-anton, line 142:
```suggestion
    if (this.searchQuery) {
      this.search();
    }
```

Maybe it'd be better to hide this side-effect behind a state-chage subscription of sorts or is it something you'd want to avoid in this case?

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+
+  protected abstract query(searchQuery: string): Promise<QueryReturnValue<T>>;
+
+  protected abstract filterQueryResults(items: T[]): T[] | Promise<T[]>;
+
+  setDebounceOptions = ({ debounceMs }: DebounceOptions) => {
+    this.searchDebounced = debounce(this.executeQuery.bind(this), debounceMs);
+  };
+
+  activate = (sourceStateOverride?: Partial<SearchSourceState>) => {
```
arnautov-anton, line 138: Why the partial state here? I'd say `activate` should flip `active` to `true` and all the other side-effects should happen elsewhere. Multiple `StateStore.next` calls seem better than unexpected state changes seemingly unrelated to the method call.

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+    this.state.partialNext({ isActive: false });
+  };
+
+  async executeQuery(newSearchString?: string) {
+    const hasNewSearchQuery = typeof newSearchString !== 'undefined';
+    const searchString = newSearchString ?? this.searchQuery;
+    if (!this.isActive || this.isLoading || (!this.hasMore && !hasNewSearchQuery) || !searchString) return;
+
+    if (hasNewSearchQuery) {
+      this.resetState({ isActive: this.isActive, isLoading: true, searchQuery: newSearchString });
```
arnautov-anton, line 157: To keep the behavior similar across the newly released classes, `resetState` shouldn't take any value but rather set the store to its initial state (at class construction). Any other modification should happen after the fact:

```suggestion
    if (hasNewSearchQuery) {
      this.resetState();
      this.state.partialNext({ isLoading: true, searchQuery: newSearchString })
```

arnautov-anton: Or if you want to avoid two calls:

```ts
if (hasNewSearchQuery) {
    this.state.next((current) => {
        ...current,
        ...this.initialState,
        // overrides
        isActive: current.isActive,
        isLoading: true,
        searchQuery: newSearchString
    });
```

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+        stateUpdate.offset = (this.offset ?? 0) + items.length;
+        stateUpdate.hasMore = items.length === this.pageSize;
+      }
+
+      stateUpdate.items = await this.filterQueryResults(items);
+    } catch (e) {
+      stateUpdate.lastQueryError = e as Error;
+    } finally {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      this.state.next(({ lastQueryError, ...prev }: SearchSourceState<T>) => ({
```
arnautov-anton, line 181:
```suggestion
      this.state.next((current) => ({
```
Also, `stateUpdate` seems to get unwrapped after the "prev" (current state) and unless it (`stateUpdate`) contains reset value it does not really seem necessary to exclude it (reset value is not being set in the try block).

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+        ...prev,
+        ...stateUpdate,
+        isLoading: false,
+        items: [...(prev.items ?? []), ...(stateUpdate.items || [])],
```
arnautov-anton, line 185: Eliminating the `undefined` from `T[] | undefined` would reduce the unnecessary complexity here.

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+      hasMore: true,
+      isActive: false,
+      isLoading: false,
+      items: undefined,
+      lastQueryError: undefined,
+      next: undefined,
+      offset: 0,
+      searchQuery: '',
+      ...stateOverrides,
+    });
```
arnautov-anton, line 205: Where `initialState` would get set at the beginning (during construction) and would be `private readonly` property.

```suggestion
    this.state.next(this.initialState)
```

src/search_controller.ts
```diff
@@ -0,0 +1,484 @@
...
+  };
+
+  cancelSearchQueries = () => {
+    this.activeSources.forEach((s) => s.searchDebounced.cancel());
+  };
+
+  clear = () => {
+    this.cancelSearchQueries();
+    this.sources.forEach((source) => source.resetState({ isActive: source.isActive }));
+    this.state.next((prev) => ({
```
arnautov-anton, line 466: Nit:
```suggestion
    this.state.next((current) => ({
```

src/search_controller.ts
MartinCupela: The value `undefined` says that no search has been executed yet. This is handy to know if we want to differ btw the state where will tell the user to start search vs telling the user there are not results for the given search query.

We could think that maybe the same could be inferred from the combination of the length of the search query and the length of the items array. The problem is, when you start typing, for a split of a second, the value of loading is not set. In that case you have a search query of non-zero lenght, loading is false and results are an empty array. That leads to showing "No results found" or whatever the empty results component is. So I would say we need an indicator that tells us, whether the items value is the result of a search query.

src/search_controller.ts
MartinCupela: Yes, I will remove the side effect from both the source and controller activate methods.

MartinCupela: https://github.com/GetStream/stream-chat-js/pull/1434/commits/29cd049d2592381a20681c048a4ad86b02e37a22

src/search_controller.ts
```diff
@@ -0,0 +1,480 @@
...
+  state: StateStore<SearchSourceState<T>>;
+  protected pageSize: number;
+  abstract readonly type: SearchSourceType;
+  searchDebounced!: DebouncedExecQueryFunction;
+
+  protected constructor(options?: SearchSourceOptions) {
+    const { debounceMs, isActive, pageSize } = { ...DEFAULT_SEARCH_SOURCE_OPTIONS, ...options };
+    this.pageSize = pageSize;
+    this.state = new StateStore<SearchSourceState<T>>({
+      hasMore: true,
```
MartinCupela, line 84:
```suggestion
      hasNext: true,
```

MartinCupela: https://github.com/GetStream/stream-chat-js/pull/1434/commits/b285199e9761b5c26e0d96c47929599a69081924

src/search_controller.ts
MartinCupela: Removed partial state update on activation: https://github.com/GetStream/stream-chat-js/pull/1434/commits/072930d0afcba044b8d854b0fdb1ddc8e35c0d32

src/search_controller.ts
MartinCupela: The reason for leaving the `lastQueryError` from the previous state out of the next update is to prevent carrying over errors relevant to previous query if the current query hasn't end up in an error. If you have done 2 queries, in the second query you stored the error, then if the 3rd query has not end up in an error, you do not want to keep the error related to the 2nd query.

src/search_controller.ts
MartinCupela: renamed: https://github.com/GetStream/stream-chat-js/pull/1434/commits/9f76e9fb39591bce76ff8b77bec415e3cd667bbf

src/search_controller.ts
```diff
@@ -0,0 +1,488 @@
...
+export abstract class BaseSearchSource<T> implements SearchSource<T> {
+  state: StateStore<SearchSourceState<T>>;
+  protected pageSize: number;
+  abstract readonly type: SearchSourceType;
+  searchDebounced!: DebouncedExecQueryFunction;
+
+  protected constructor(options?: SearchSourceOptions) {
+    const { debounceMs, isActive, pageSize } = { ...DEFAULT_SEARCH_SOURCE_OPTIONS, ...options };
+    this.pageSize = pageSize;
+    this.state = new StateStore<SearchSourceState<T>>({ ...this.initialState, isActive });
```
arnautov-anton, line 84: I'd expect `this.initialState` to be constructed above:
```suggestion
    this.initialState = { 
        hasNext: true,
        isActive,
        isLoading: false,
        items: undefined,
        lastQueryError: undefined,
        next: undefined,
        offset: 0,
        searchQuery: '',
    }
    this.state = new StateStore<SearchSourceState<T>>(this.initialState);
```

src/search_controller.ts
```diff
@@ -0,0 +1,488 @@
...
+      hasNext: true,
+      isActive: false,
+      isLoading: false,
+      items: undefined,
+      lastQueryError: undefined,
+      next: undefined,
+      offset: 0,
+      searchQuery: '',
+    };
+  }
```
arnautov-anton, line 119: This then does not have to exist.
```suggestion
```

src/search_controller.ts
MartinCupela: I have removed `isActive` from the constructor https://github.com/GetStream/stream-chat-js/pull/1434/commits/e145dae3dbb5d08f5152e4dae5eb3e61e033893d

arnautov-anton (approved):

src/search_controller.ts
```diff
@@ -0,0 +1,486 @@
...
+};
+
+export type SearchControllerOptions = {
+  config?: Partial<SearchControllerConfig>;
+  sources?: SearchSource[];
+};
+
+export class SearchController<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics> {
+  // Keeps data that is subject to change. Not intended for direct use by integrators.
+  internalState: StateStore<InternalSearchControllerState<StreamChatGenerics>>;
```
arnautov-anton, line 371:
```suggestion
  /**
   * Not intended for direct use by integrators, might be removed without notice resulting in
   * broken integrations.
   */
  _internalState: StateStore<InternalSearchControllerState<StreamChatGenerics>>;
```

isekovanic (approved):

PR #1442, opened 2025-01-21
fix: add cache disabling option

isekovanic:
## CLA

- [ ] Code changes are tested

Fixes the issues mentioned in [this Zendesk ticket](https://getstream.zendesk.com/agent/tickets/60144).

Mainly, this flag is meant to be used for a very niche' usecase of the LLC when it's used purely as a REST API on the backend. The issue boils down to a couple of things:

- For performance reasons, we often do caching of various heavy operations/entities within the LLC (such as channels, polls and similar) in order to be able to look already seen things quickly
- This works well in principle, however if used on the backend it can quickly increase the LLC's memory and likely crash any pod it's being run on

As an example - let us say that we have 10000 users and each user has 5 distinct channels (some group ones, maybe some direct messaging channels etc.). The first time one of these users runs `queryChannels` towards the custom backend (which in turn runs `client.queryChannels`), the entities will end up in both `client.activeChannels` and `client.configs` until the pod is restarted. 

On the client side app this works fine because you physically would not be able to query through so many channels given a `userId`. 

Provided below is a quick summary on why it's okay to do this in a REST API setting:

- The only place where `client.activeChannels` is being used in this setting is [here](https://github.com/GetStream/stream-chat-js/blob/8daedeb18da49f77a91f30ca5d98aa4a2d13b679/src/client.ts#L1989), the rest are either WS event handlers or state dispatchers (none of which we need given the constraints)
- The `client.configs` object is only really used to get a `channel` configuration in the client-side SDKs, in the case of a REST API they can use the `config` returned from the HTTP request
- The `pollCache` is something we do not need on the server-side, as again we have other ways to get a poll (and it also comes baked in its own messages) 

## Changelog

-


src/types.ts
```diff
@@ -1170,6 +1170,15 @@ export type StreamChatOptions = AxiosRequestConfig & {
   baseURL?: string;
   browser?: boolean;
   device?: BaseDeviceFields;
+  /**
+   * Disables the hydration of all caches within the JS Client. This includes this.activeChannels,
+   * this.polls.pollCache and this.config.
+   * It is mainly meant to be used for integrations where stream-chat is used as a BE REST API, not
```
MartinCupela, line 1176:
```suggestion
   * It is mainly meant to be used for integrations where stream-chat is used as a server-side service interacting with Stream's REST API, not
```

src/client.ts
```diff
@@ -1985,7 +1988,9 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       return channel;
     }
     const channel = new Channel<StreamChatGenerics>(this, channelType, channelID, custom);
-    this.activeChannels[channel.cid] = channel;
+    if (!this.options.disableCache) {
```
MartinCupela, line 1991: Could this be necessary also in `getChannelByMembers` method?

MartinCupela:
@isekovanic Is it possible that [due to `Channel._checkIsInitialized` ](https://github.com/GetStream/stream-chat-js/blob/8daedeb18da49f77a91f30ca5d98aa4a2d13b679/src/channel.ts#L1699) some methods interacting with the BE will not work? (channel.sendEvent, [channel.deleteReaction](https://github.com/GetStream/stream-chat-js/blob/8daedeb18da49f77a91f30ca5d98aa4a2d13b679/src/channel.ts#L366), ...)

src/client.ts
isekovanic: We do ! Thanks, missed that one. 

We're also missing it in `channel.query` for adding the config. Adding them in a sec.

isekovanic:
@MartinCupela I would say no, since the channel instances do exist and are returned adequately - it's just the cache that is not being populated - meaning the `channel` itself should be initialized as it should if they `watch()` it (which is typically not something done on the BE as far as I know). 

I think that might also be out of scope of the constraints we give to the flag, since this is really only meant to be used as a REST API.

What do you think ?

MartinCupela:
I realized that the check for being initialized is looking into whether the client is running server-side, so should be ok.

src/client.ts
```diff
@@ -318,6 +318,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       withCredentials: false, // making sure cookies are not sent
       warmUp: false,
       recoverStateOnReconnect: true,
+      disableCache: false,
```
szuperaz, line 322: Can we add a check inside the client to only enable this when `this._isUsingServerAuth`? We won't test this on the client-side at all, so this might still be buggy there. I can see integrators enabling this trying to solve a different problem altogether causing weird to debug issues for us.

szuperaz: I meant enable this if `disableCache  && this._isUsingServerAuth`

isekovanic: Will update the jsdocs as well

isekovanic:
I do not, no @szuperaz - would be happy to though

szuperaz:
@isekovanic did you check that we don't add references here: https://github.com/GetStream/stream-chat-js/blob/master/src/client_state.ts#L14-L16 - this is a new to me as well, so I might be missing something, but following source code it seems possible that we'll store data even if cache is disabled

test/unit/client.js
```diff
@@ -580,6 +580,20 @@ describe('Client WSFallback', () => {
 });
 
 describe('Channel.queryChannels', async () => {
+	it('should not hydrate activeChannels and channel configs when disableCache is true', async () => {
```
MartinCupela, line 583: @isekovanic could you, please, add test also for channel.query()?

src/poll_manager.ts
```diff
@@ -101,6 +101,9 @@ export class PollManager<SCG extends ExtendableGenerics = DefaultGenerics> {
   };
 
   private setOrOverwriteInCache = (pollResponse: PollResponse<SCG>, overwriteState?: boolean) => {
+    if (this.client.options.disableCache) {
```
MartinCupela, line 104: @isekovanic would this be necessary also for ThreadManager?

isekovanic: No, for `ThreadManager` you have to use the `queryThreads` directly from it rather than `client.queryThreads` - so there is an alternative. For polls there isn't much of a choice hence why I added it

szuperaz:
@isekovanic I went through all `activeChannels` references and checked if we read from them without null checks. I found two places (it's very much possible you can't reach these without WS connection, still, I think it's better to add null checks everywhere):
- https://github.com/GetStream/stream-chat-js/blob/master/src/client.ts#L1224 
- https://github.com/GetStream/stream-chat-js/blob/fix/add-cache-disabling-option/src/client.ts#L1986

test/unit/client.js
isekovanic: Yup

MartinCupela:
@isekovanic I think we should also make sure that [client.mutedChannels and client.mutedUsers](https://github.com/GetStream/stream-chat-js/blob/8daedeb18da49f77a91f30ca5d98aa4a2d13b679/src/client.ts#L253) are not populated server-side. WDYT?

src/client.ts
oliverlaz: Nitpick: Instead of `disableCache,` can we use `enableCache`? It would save us from using double negations like `!disableCache`.

`if (enableCache)` reads better than `if (!disableCache)`

isekovanic: It's already changed to that, changed it so that it includes the server side checks too

isekovanic: Scratch that, hadn't pushed - NOW it's there

isekovanic:
@szuperaz sounds good, will do that ! 

@MartinCupela saw those too, however they only every updated through WS events, which we consider out of scope for this. I could add the flag for brevity but would prefer to not touch too many of these places at once if you don't mind

szuperaz:
PR for chat QA test update: https://github.com/GetStream/chat/pull/7890

src/client.ts
```diff
@@ -296,7 +296,7 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
     // set the key
     this.key = key;
     this.listeners = {};
-    this.state = new ClientState<StreamChatGenerics>();
+    this.state = new ClientState<StreamChatGenerics>({ client: this });
```
MartinCupela, line 299: I am wandering whether it makes sense to even have `client.state` when the client is initialized with cache disabled. But that would probably be a breaking change right?

isekovanic: Not necessarily a breaking change, but I guess it's safer this way for now - I guess it's better to have it there in case the `client.state` is extended at some point with other things so that we don't have to revisit the stuff here.

MartinCupela: Why do we even have a attribute called `state` that keeps user cache only 🤦 .

isekovanic: No idea either :D Hadn't fiddled with it before sadly, hence why it took me a while to reverse engineer everything here

szuperaz (approved):

PR #1452, opened 2025-01-28
[CHA-403] Private messaging

totalimmersion:
This PR is adding private messaging for a select list of users in a channel. If the user who sends the message has the `CreateRestrictedVisibilityMessage` permission, then by adding `restricted_visibility: [ 'john', 'jane' ] would limit the visibility of that message to only john and jane. The other members of the channel can't see the message. The sender of the message can always see it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,8 +15,5 @@ jobs:
 
       - uses: ./.github/actions/setup-node
 
-      - name: Commit message lint
-        run: echo "${{ github.event.pull_request.title }}" | yarn commitlinter
-
       - name: Lint
         run: yarn lint
--- a/src/types.ts
+++ b/src/types.ts
@@ -2643,6 +2643,7 @@ export type MessageBase<
   pinned_at?: string | null;
   poll_id?: string;
   quoted_message_id?: string;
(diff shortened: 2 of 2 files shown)


vishalnarkhede (approved):

.github/workflows/lint.yml
```diff
@@ -15,8 +15,5 @@ jobs:
 
       - uses: ./.github/actions/setup-node
 
-      - name: Commit message lint
-        run: echo "${{ github.event.pull_request.title }}" | yarn commitlinter
```
MartinCupela, line 19: Instead of removing the linter we could use `FIXES CHA-403` and maintain the rules for commit description.

MartinCupela: Following those rules would help us in the future using automated release with tools like semantic-release

totalimmersion: of course we could, but what is the purpose? all these commits gets squashed anyways.

MartinCupela: I understand this lints the PR title that is later used when squash merging the PR. Or does it do sth different?

totalimmersion: it's linting individual commits as well within the branch/pr.

PR #1456, opened 2025-01-29
fix: improper poll state overwrite

isekovanic:
## CLA

- [x] Code changes are tested

This PR fixes a case of overwriting the poll cache when we should not do that. Essentially, hydrating all messages in `messageSet.messages` whenever `channel.query` is invoked does 2 things:

- Takes in all of the new messages and hydrates the cache with them (they do not exist in the cache yet)
- Also takes in the old messages, which were used as a basis to paginate on top of and rehydrates the cache with them (they already existed in the cache before)

The problem here is; since we do not update `message.poll` objects at all, but rather use them as an ID reference towards the cache - these polls are stale (since they're retrieved from the stale state and not as an HTTP response). The state is then overwritten with the stale data until we query all messages again.

The bug is particularly noticeable if we do some state changes to a poll (for example vote on it, add some answers or similar) and then paginate elsewhere within the messages (up or down). We can see that the poll will reset back to the state it had on the initial load.

## Changelog

-

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1211,7 +1211,7 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
       }),
     };
 
-    this.getClient().polls.hydratePollCache(messageSet.messages, true);
+    this.getClient().polls.hydratePollCache(state.messages, true);
 
     const areCapabilitiesChanged =
       [...(state.channel.own_capabilities || [])].sort().join() !==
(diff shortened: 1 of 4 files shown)


isekovanic:
Our `queryChannels` hydration logic also had the same bug (when paginating channels for example). It's fixed with https://github.com/GetStream/stream-chat-js/pull/1456/commits/299a3afc2dff315f8cfc8320808c351894563558.

Thanks @MartinCupela for pointing it out !

MartinCupela (approved):

PR #1458, opened 2025-01-30
fix!: replace StreamChatGenerics with module augmentation

arnautov-anton:
fix: `channel_role` has been removed from the type of the `members` parameter of the `Channel.inviteMembers` method
fix: properties `created_by` and `created_by_id` have been added to `ChannelData` and `ChannelQueryOptions` types

BREAKING CHANGES:
- dropped `StreamChatGenerics`, use `Custom<Entity>Data` to extend your types
- type `InviteOptions` has been renamed to `UpdateChannelOptions`
- type `UpdateChannelOptions` has been renamed to `UpdateChannelTypeRequest`
- type `ThreadResponseCustomData` has been renamed to `CustomThreadData`
- type `MarkAllReadOptions` has been deleted in favour of type `MarkChannelsReadOptions`
- type `QueryFilter` no longer supports `$ne` and `$nin` operators
- type `ChannelMembership` has been deleted in favour of type `ChannelMemberResponse`
- function `formatMessage` (`utils.ts`) no longer returns `__html` property in the formatted message output


src/types.ts
```diff
@@ -33,33 +38,19 @@ export type RequireOnlyOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<
...
-  pollOptionType: UR;
-  pollType: UR;
-  reactionType: UR;
-  userType: UR;
-};
+export type UnknownType = UR; // alias to avoid breaking change
+
+export interface CustomAttachmentType {}
+export interface CustomChannelType {}
+export interface CustomCommandType {}
```
arnautov-anton, line 45: Might have to figure out a better way for this to work. Currently using `keyof CustomCommandType` to make it work with interface merging.

src/client.ts
```diff
@@ -1244,36 +1225,40 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
 
         /** Updating only available properties in _user object. */
-        for (const key in event.user) {
-          if (_user && key in _user) {
-            _user[key] = event.user[key];
+        for (const key in _user) {
+          const updateKey = key as keyof typeof _user;
+
+          if (updateKey in event.user) {
+            // @ts-expect-error it has an issue with this, not sure why
```
arnautov-anton, line 1256: 😬

src/types.ts
arnautov-anton: This right here: https://github.com/GetStream/stream-chat-js/pull/1458/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR2124

Actual module augmentation would have to look something like:
```ts
// file.d.ts

import 'stream-chat';

declare module 'stream-chat' {
    interface CustomCommandType {
        customCommand: unknown;
        otherCustomCommand: unknown;
    }
}
```

src/poll.ts
```diff
@@ -83,36 +70,37 @@ export type PollOptionVotesQueryParams = {
 
 type OptionId = string;
 
-export type PollState<SCG extends ExtendableGenerics = DefaultGenerics> = SCG['pollType'] &
-  Omit<PollResponse<SCG>, 'own_votes' | 'id'> & {
+// FIXME: ideally this would work without the unknown record...
+export type PollState = (CustomPollType & Record<string, unknown>) &
```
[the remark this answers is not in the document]
myandrienko: Object type literals receive an implicit key signature by default. Interfaces don't. But I think it should be fine to just declare `CustomPollType` like this:

```ts
interface CustomPollType {
  [key: string]: unknown;
}
```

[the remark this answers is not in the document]
arnautov-anton: I did that prior to doing this nonsense but then - not sure what happened - chose current solution over this one. I believe I wanted to apply this as a default to all the other custom interfaces but that seemed to break the rest of the codebase (defined properties suddenly reporting `uknown` value when accessed). I'm not against it so I'll apply your suggestion.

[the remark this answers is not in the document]
arnautov-anton: Ah - maybe not - this new solution seems to filter some of the known properties for some reason:

With
```ts
interface CustomPollType {
  [key: string]: unknown;
}
```
we get this:
Screenshot 2025-02-04 at 4 59 24 PM

With existing solution we get all the properties as expected (including the merged custom one):
Screenshot 2025-02-04 at 5 01 51 PM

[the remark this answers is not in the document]
arnautov-anton: Nevermind, there's no need - the `PollResponse` type is already being extended [here](https://github.com/GetStream/stream-chat-js/pull/1458/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR2942-R2943) so this second extension is useless - I'll remove it.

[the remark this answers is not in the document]
arnautov-anton: https://github.com/GetStream/stream-chat-js/pull/1458/commits/9ce44f75cca89f1249db7eab29b6e2d3a992aaff

src/types.ts
isekovanic: Looks good ! We're already doing similar things for other parts of the RN SDK (extending functionality and overloads) so this will come natural.

The entire change is breaking I assume, right ? What's the strategy to go about merging this when the time comes ? I suppose all SDKs would need to bump their major versions if they wish to upgrade their version of `stream-chat`. Although kind of large in size, the change seems easy to do (if generics were being used in the first place).

arnautov-anton: Yes, in case of React SDK we'd be releasing v13.0.0 which will bump its `stream-chat` peer to v9.0.0 - the peer of the SDK is currently bound to v8-something.

As for the migration - see this draft of a [migration guide here](https://github.com/GetStream/docs-content/pull/167), feel free to suggest enhancements. Not sure where it'll live yet, I assume both React and RN will be similar in terms of migration but I'm open to adding tabs to switch between RN and React.

src/types.ts
```diff
@@ -33,33 +38,19 @@ export type RequireOnlyOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<
...
+export interface CustomEventType {}
+export interface CustomMemberType {}
+export interface CustomMessageType {}
+export interface CustomPollOptionType {}
+export interface CustomPollType {}
+export interface CustomReactionType {}
+export interface CustomUserType {}
+export interface CustomThreadType {}
```
arnautov-anton, line 53: As for the naming - are we happy with this? I'd personally drop the `Type` suffix as it does not really fit (being an interface). It's breaking either way. Personally - I'd either replace `Type` suffix with `Data` suffix or just remove the suffix altogether.

src/types.ts
isekovanic: That's fair - in that case we should pick it up earlier rather than later since `stream-chat` is a direct dependency for `stream-chat-react-native` (and not a peer one).

Since we'll likely have to do a V7 to go along with it as well, maybe we can do the same ? it should get rid of our pesky TS errors when there's a version mismatch (despite the fact it's a direct dependency).

fyi @oliverlaz

oliverlaz: @isekovanic I agree! 

We should have a similar guide, and also we need to update the following chapters:
- https://getstream.io/chat/docs/sdk/react/guides/typescript_and_generics/
- https://getstream.io/chat/docs/sdk/react-native/customization/typescript/

src/types.ts
oliverlaz: I like the `CustomXyzData` more. That's the term we use across our documentation.
Also, I think it would be good if we moved these to a `customTypes.ts` file. It should make it more obvious which types are meant to be extended.

arnautov-anton: [Done](https://github.com/GetStream/stream-chat-js/pull/1458/commits/093493604020cd283f4ab5827f6480dedefcc0e7). :)

PR #1460, opened 2025-01-31
feat: reactive channel list

isekovanic:
## CLA

- [ ] Code changes are tested

This PR moves the channel list state from the SDKs down to the LLC. A better description will be available once I finish writing the tests for this, but wanted to open it as a draft until then.

TODOs:

- [x] Write tests
- [x] Address the open `TODO`s in the code with the team

RN SDK linked PR: https://github.com/GetStream/stream-chat-react-native/pull/2924

## Changelog

-


src/client.ts
```diff
@@ -1935,10 +1948,13 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
   getChannelByMembers = (channelType: string, custom: ChannelData<StreamChatGenerics>) => {
     // Check if the channel already exists.
     // Only allow 1 channel object per cid
-    const membersStr = [...(custom.members || [])].sort().join(',');
-    const tempCid = `${channelType}:!members-${membersStr}`;
+    const memberIds = (custom.members ?? []).map((member: string | NewMemberPayload<StreamChatGenerics>) =>
```
isekovanic, line 1951: This was a bug (that likely no one encountered because our docs suggest using `members` as a list of strings, but still possible).

src/utils.ts
```diff
@@ -747,3 +755,247 @@ export const messageSetPagination = <StreamChatGenerics extends ExtendableGeneri
     return messagePaginationLinear(params);
   }
 };
+
+/**
+ * A utility object used to prevent duplicate invocation of channel.watch() to be triggered when
+ * 'notification.message_new' and 'notification.added_to_channel' events arrive at the same time.
+ */
+const WATCH_QUERY_IN_PROGRESS_FOR_CHANNEL: Record<string, Promise<QueryChannelAPIResponse> | undefined> = {};
```
isekovanic, line 763: All of the utilities below this are snagged from [here](https://github.com/GetStream/stream-chat-react/blob/master/src/utils/getChannel.ts) and (here)[https://github.com/GetStream/stream-chat-react/blob/master/src/components/ChannelList/utils.ts].

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+    const { channels, pagination } = this.state.getLatestValue();
+    const { sort } = pagination ?? {};
+    if (channels && event.channel_type && event.channel_id) {
+      const channel = await getAndWatchChannel({
+        client: this.client,
+        id: event.channel_id,
+        type: event.channel_type,
+      });
+
+      this.setChannels(
```
isekovanic, line 412: The fix for this is already [ported to develop](https://github.com/GetStream/stream-chat-react-native/pull/2925/files).

src/types.ts
```diff
@@ -3796,3 +3797,14 @@ export type VelocityFilterConfig = {
   rules: VelocityFilterConfigRule[];
   async?: boolean;
 };
+
+export type MoveChannelUpwardsParams<SCG extends ExtendableGenerics = DefaultGenerics> = {
```
MartinCupela, line 3801: Will this always always be upwards?

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+  moveChannelUpwards,
+  shouldConsiderArchivedChannels,
+  shouldConsiderPinnedChannels,
+} from './utils';
+
+export type ChannelManagerPagination<SCG extends ExtendableGenerics = DefaultGenerics> = {
+  filters: ChannelFilters<SCG>;
+  hasNext: boolean;
+  isLoading: boolean;
+  isLoadingNext: boolean;
```
MartinCupela, line 28: Why do we need 2 loading flags?

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+  isChannelArchived,
+  isChannelPinned,
+  moveChannelUpwards,
+  shouldConsiderArchivedChannels,
+  shouldConsiderPinnedChannels,
+} from './utils';
+
+export type ChannelManagerPagination<SCG extends ExtendableGenerics = DefaultGenerics> = {
+  filters: ChannelFilters<SCG>;
+  hasNext: boolean;
```
MartinCupela, line 26: Do we need hasNext flag if we have a cursor?

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+  isLoading: boolean;
+  isLoadingNext: boolean;
+  options: ChannelOptions;
+  sort: ChannelSort<SCG>;
+  stateOptions: ChannelStateOptions;
```
MartinCupela, line 31: I am wondering, if one day we would introduce configuration parameters that would not be related to state neither to HTTP request. In that case we would probably need another XOptions field?

Another thing I am thinking - do the `skipInitialization` and `offlineMode` need to me reactive?

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+};
+
+export type ChannelManagerState<SCG extends ExtendableGenerics = DefaultGenerics> = {
+  channels: Channel<SCG>[];
+  pagination: ChannelManagerPagination<SCG>;
+  ready: boolean;
```
MartinCupela, line 37: When is `ready` flipped to `true`?

src/channel_manager.ts
```diff
@@ -0,0 +1,517 @@
...
+  'message.new': 'newMessageHandler',
+  'member.updated': 'memberUpdatedHandler',
+  'notification.added_to_channel': 'notificationAddedToChannelHandler',
+  'notification.message_new': 'notificationNewMessageHandler',
+  'notification.removed_from_channel': 'notificationRemovedFromChannelHandler',
+};
+
+export type ChannelManagerOptions = {
+  allowNewMessagesFromUnfilteredChannels?: boolean;
+  lockChannelOrder?: boolean;
```
MartinCupela, line 98: Could `skipInitialization` and `offlineMode` go here? Would it make sense?

isekovanic: I actually thought a lot about this as well, decided to move them into the reactive state in the end because they seemed `pagination` related. But getting into it further, it does indeed make sense to move 'em out of here as they're mostly related to different initialization behaviour of the state if provided. 

I'll do that in a bit, thanks !

src/channel_manager.ts
isekovanic: It's done [here](https://github.com/GetStream/stream-chat-js/blob/93dec4b37d6e1e49b17e853378c6476ec5505906/src/channel_manager.ts#L223)

src/channel_manager.ts
isekovanic: Answered [here](https://github.com/GetStream/stream-chat-js/pull/1460/files#r1939379422), will do that - thanks !

src/channel_manager.ts
isekovanic: One is for when the entire channel list is loading (i.e `queryChannels`) whereas the other is only when the next page is loading (i.e when `loadNext` is invoked); we need to distinguish between the 2 in my opinion as they technically mean two different things - what do you think ?

src/channel_manager.ts
isekovanic: Technically we do not, we can check if the `cursor` is `null` at some point and handle it that way; this seemed useful as some extra state that would help us know whether we can paginate further or not (for example, if a pagination response has less elements than our `limit`)

src/types.ts
isekovanic: I believe so, yes - we do not handle other cases with this util specifically; I'm up for renaming it to something more generic if you prefer though

src/channel_manager.ts
khushal87: @isekovanic, do we have a better name for the `ready` var? This doesn't seem to define its actual usage and is confusing.

src/utils.ts
```diff
@@ -823,3 +831,247 @@ export const messageSetPagination = <StreamChatGenerics extends ExtendableGeneri
...
+ * Calls channel.watch() if it was not already recently called. Waits for watch promise to resolve even if it was invoked previously.
+ * If the channel is not passed as a property, it will get it either by its channel.cid or by its members list and do the same.
+ * @param client
+ * @param members
+ * @param options
+ * @param type
+ * @param id
+ * @param channel
+ */
+export const getAndWatchChannel = async <StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>({
```
MartinCupela, line 859: Does this come from React SDK? (https://github.com/GetStream/stream-chat-react/blob/master/src/utils/getChannel.ts)

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+      Object.entries({
+        channelDeletedHandler: this.channelDeletedHandler,
+        channelHiddenHandler: this.channelHiddenHandler,
+        channelVisibleHandler: this.channelVisibleHandler,
+        memberUpdatedHandler: this.memberUpdatedHandler,
+        newMessageHandler: this.newMessageHandler,
+        notificationAddedToChannelHandler: this.notificationAddedToChannelHandler,
+        notificationNewMessageHandler: this.notificationNewMessageHandler,
+        notificationRemovedFromChannelHandler: this.notificationRemovedFromChannelHandler,
+      }) as [string, EventHandlerType<SCG>][],
```
arnautov-anton, line 149:
```suggestion
      Object.entries<EventHandlerType<SCG>>({
        channelDeletedHandler: this.channelDeletedHandler,
        channelHiddenHandler: this.channelHiddenHandler,
        channelVisibleHandler: this.channelVisibleHandler,
        memberUpdatedHandler: this.memberUpdatedHandler,
        newMessageHandler: this.newMessageHandler,
        notificationAddedToChannelHandler: this.notificationAddedToChannelHandler,
        notificationNewMessageHandler: this.notificationNewMessageHandler,
        notificationRemovedFromChannelHandler: this.notificationRemovedFromChannelHandler,
      }),
```
This then reveals that `EventTypeHandler` is not properly taking into account all the possible handler return combinations (in this case `Promise<void>`) - maybe we should just do `any` - it's not like we really care about the outputs, right?

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+    );
+  }
+
+  public setChannels = (valueOrFactory: ChannelSetterParameterType<SCG>) => {
+    this.state.next((current) => {
+      const { channels: prevChannels } = current;
+      let newValue = valueOrFactory;
+      if (newValue && typeof newValue === 'function') {
+        newValue = newValue(prevChannels);
+      }
```
arnautov-anton, line 159: Let's reuse `StateStore.isPatch` for this part (needs to be exported from `store.ts`), also let's change the naming as it's not a previous state yet but state before it gets commited to handlers (which it does not have to).

```suggestion
      const { channels: currentChannels } = current;
      const newChannels = isPatch(valueOrFactory) ? valueOrFactor(currentChannels) : valueOrFactory;
```

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+      Partial<ChannelManagerEventHandlerOverrides<SCG>>
+    >((acc, [key, value]) => {
+      if (value) {
+        acc[key as keyof ChannelManagerEventHandlerOverrides<SCG>] = value;
+      }
+      return acc;
+    }, {});
+    this.eventHandlerOverrides = new Map(
+      Object.entries(truthyEventHandlerOverrides) as [string, EventHandlerOverrideType<SCG>][],
+    );
```
arnautov-anton, line 181:
```suggestion
    this.eventHandlerOverrides = new Map(
      Object.entries<EventHandlerOverrideType<SCG>>(truthyEventHandlerOverrides),
    );
```

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+          options: newOptions,
+        },
+        ready: true,
+      });
+    } catch (error) {
+      this.client.logger('error', (error as Error).message);
+      this.state.next((currentState) => ({
+        ...currentState,
+        pagination: { ...currentState.pagination, isLoading: false },
+      }));
```
arnautov-anton, line 237: This part should probably live in `finally` block.

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+    const { pagination, channels } = this.state.getLatestValue();
+    if (!channels) {
+      return;
+    }
+    const { filters, sort } = pagination ?? {};
+
+    const channelType = event.channel_type;
+    const channelId = event.channel_id;
+
+    if (channelType && channelId) {
```
MartinCupela, line 340: Can we do this to prevent nesting?:

```suggestion
  if (!channelType || !channelId) return;
```

src/utils.ts
```diff
@@ -823,3 +831,247 @@ export const messageSetPagination = <StreamChatGenerics extends ExtendableGeneri
...
+  options,
+  type,
+}: GetChannelParams<StreamChatGenerics>) => {
+  if (!channel && !type) {
+    throw new Error('Channel or channel type have to be provided to query a channel.');
+  }
+
+  // unfortunately typescript is not able to infer that if (!channel && !type) === false, then channel or type has to be truthy
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const channelToWatch = channel || client.channel(type!, id, { members });
```
oliverlaz, line 873: TS can't infer because I feel you have a bug in the condition above. Shouldn't it be:

```ts
  if (!channel || !type) {
    throw new Error('Channel or channel type have to be provided to query a channel.');
  }
```

src/utils.ts
```diff
@@ -823,3 +831,247 @@ export const messageSetPagination = <StreamChatGenerics extends ExtendableGeneri
...
+
+  // need to keep as with call to channel.watch the id can be changed from undefined to an actual ID generated server-side
+  const originalCid = channelToWatch?.id
```
oliverlaz, line 876: `channelToWatch` can't be undefined at this point. Let's drop the `?.`

arnautov-anton: This is a problem of the [`Channel` definition](https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L75), unfortunately.

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+      }));
+      throw error;
+    }
+  };
+
+  public loadNext = async () => {
+    const { pagination, channels, ready } = this.state.getLatestValue();
+    const { filters, sort, options, isLoadingNext, hasNext } = pagination;
+
+    if (!ready || isLoadingNext || !hasNext) {
```
MartinCupela, line 246: If we had only `isLoading` instead of `isLoading + isLoadingNext` and we took into consideration `offset` as `undefined` thus could remove `read`, `loadNext` and `queryChanels` could be merged into single method.

src/utils.ts
```diff
@@ -823,3 +831,247 @@ export const messageSetPagination = <StreamChatGenerics extends ExtendableGeneri
...
+      lastPinnedChannelIndex++;
+    } else {
+      lastPinnedChannelIndex = 0;
+    }
+  }
+
+  return lastPinnedChannelIndex;
+}
+
+export const moveChannelUpwards = <StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>({
```
MartinCupela, line 1036: I would change the name as what the function does it moves the channel to determined index, but it does not need to be upwards as that is UI conotation.

```suggestion
export const changeChannelPosition = <StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>({
```

src/channel_manager.ts
isekovanic: Sure ! We can do that. While it is true that we don't particularly care about what it returns (it should generally be void, though) I thought it'd make sense to cover what's expected - hence the typing. The general idea was to avoid the expectation that this also returns a value.

src/channel_manager.ts
isekovanic: Can do, thanks !

src/channel_manager.ts
isekovanic: Not necessarily, due to 2 reasons: 

- It won't mean that `ready` will get set if the request failed
- `channels` will also not be set if the request failed

If we wanted to put it in a `finally` block, it would mean that we need to do 2 state updates whenever the query succeeds: 

- One to actually set the state (`ready`, `channels` etc)
- One to set `isLoading`

which I believe we want to avoid

src/utils.ts
isekovanic: I don't think so, because if a `channel` is passed and not a `type` then the `type` can be inferred from `channel.type`. It should be able to handle the situation of not passing a specific `channel` but passing its `type` so that later we can try to create a `cid` (either through `members` or `channel_id`).

The suggestion above would fail for mostly any invocation of this that does not have a `channel` passed to it.

src/utils.ts
isekovanic: Nope, @oliverlaz 's right - we don't need a `null` check here; thanks !

src/channel_manager.ts
```diff
@@ -0,0 +1,527 @@
...
+    eventHandlerOverrides?: ChannelManagerEventHandlerOverrides<SCG>;
+    options?: ChannelManagerOptions;
+  }) {
+    this.client = client;
+    this.state = new StateStore<ChannelManagerState<SCG>>({
+      channels: [],
+      pagination: {
+        isLoading: false,
+        isLoadingNext: false,
+        hasNext: false,
```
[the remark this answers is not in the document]
isekovanic: While this is very true and would totally work, I would not rely on properties whose purpose is something else to figure this out - it may lead to pesky bugs in the future (for example, if we have some part of the code that relies on something like `if (hasNext) { ... }` the `undefined` value would act as a falsy value and all of a sudden we'd need to add very specific checks everywhere). Then again this is exactly the behaviour we'd expect as you suggested, so perhaps I'm overthinking it a little bit :D 

Do you see an issue with keeping `ready` as part of the state (we can of course rename this to something much better and descriptive as @khushal87 suggested) rather than trying to infer it from other properties ?

src/utils.ts
isekovanic: Yep - this method does not have the capability to move a `channel` any other direction than upwards (or at the very least make it stay in place). I do agree with what @MartinCupela is saying though, perhaps we need something better here as indeed `upwards` is a bit vague as a term to use for a list specifically.

Maybe something like `shiftChannelForward`, to be inline with JS lingo ?

src/channel_manager.ts
isekovanic: Yep ! Exactly. Having `channels` be `undefined` essentially means that wherever we actually use `channels` in the future we need to take this into consideration; this way, we can use `channels` directly everywhere and use `ready` (or whatever we decide to name it) wherever applicable (for example, loading state on the UI).

src/channel_manager.ts
isekovanic: While it is true that we can (on the SDK's side) distinguish between these 2, if an integrator were to use their own custom `ChannelManager` this information would be useful for them since they wouldn't need to rely on additional logic to figure out the loading state, but rather only on the 2 loading flags. 

`isLoading` basically means a full reload whereas `isLoadingNext` refers to paginating. Also, this opens up the possibility to perhaps depend on one of these specifically in the future if we need to do so, without introducing a breaking change

src/channel_manager.ts
isekovanic: Yep, thanks !

src/channel_manager.ts
isekovanic: Essentially the same deal as what I mentioned [here](https://github.com/GetStream/stream-chat-js/pull/1460#discussion_r1948670443). While it is true that we can easily merge these 2, do we want to do that ? Making potential changes towards one of them would make things difficult in the future (for example, if cursor pagination is ever introduced to `queryChannels` as well).

Also, having them separate means that integrators could easily use their own `ChannelManager` given a set of rules and not worry about pagination

isekovanic: What do you think ?

src/utils.ts
MartinCupela: Proposal:

- `raiseChannelPriority`
- `promoteChannel`

src/channel_manager.ts
isekovanic: As discussed with Martin on a huddle, we'll keep this behaviour as it can be useful for integrators as well as us

src/utils.ts
arnautov-anton: `promoteChannel` does actually sound pretty good, +1 from me. :)

isekovanic: Done ! Thanks guys !

MartinCupela (approved):

PR #1461, opened 2025-01-31
feat(StateStore): state modifiers and store merging

arnautov-anton:


src/store.ts
```diff
@@ -69,4 +97,236 @@ export class StateStore<T extends Record<string, unknown>> {
...
+    // pull latest and compare against cached - if they differ, cache latest and produce new combined
+    if (!this.handlers.size) {
+      const originalValue = this.original.getLatestValue();
+      const mergedValue = this.merged.getLatestValue();
+
+      if (
+        originalValue !== this.cachedOriginalValue ||
+        mergedValue !== this.cachedMergedValue
+      ) {
+        this.value = {
```
MartinCupela, line 267: Do we want to give a try to utils/mergeWith.ts? It could be handy for nested objects like message composer config is

MartinCupela (approved):

PR #1466, opened 2025-02-12
fix: premature channels state consumption in ChannelManager

isekovanic:
## CLA

- [ ] Code changes are tested

There was an unfortunate bug with the pagination logic in the new `ChannelManager` that somehow evaded me in the last PR. The offset was not being calculated properly, particularly in cases where:

- We might have `state.channels` receive updates from something other than pagination (WS events for instance) and it will then contain either broken pagination or duplicate values when actually paginating (for example if a WS event arrives as the query is being executed)
- The state might also change since there's nothing from preventing it (`setChannels` is `public`)
- Any other case really between 2 `loadNext()` invocations

For this purpose, when running the queries we always calculate the offset through `state.channels.length` and we make sure to run `state.getLatestValue()` as late as possible, particularly when there are HTTP calls involved (we can ignore the other instances of this happening).

In addition to fixing `loadNext()`, the relevant WS event handlers have also been fixed (ones that run asynchronously) to make sure the most up to date state is always there. 

## Changelog

-

### DIFF
--- a/src/channel_manager.ts
+++ b/src/channel_manager.ts
@@ -290,7 +290,7 @@ export class ChannelManager<SCG extends ExtendableGenerics = DefaultGenerics> {
       });
       const nextChannels = await this.client.queryChannels(filters, sort, options, this.stateOptions);
       const { channels } = this.state.getLatestValue();
-      const newChannels = uniqBy([...(channels || []), ...nextChannels], 'cid');
+      const newChannels = uniqBy<Channel<SCG>>([...(channels || []), ...nextChannels], 'cid');
       const newOptions = { ...options, offset: newChannels.length };
 
       this.state.partialNext({
--- a/src/channel_manager.ts
+++ b/src/channel_manager.ts
@@ -19,6 +19,7 @@ import {
   promoteChannel,
   shouldConsiderArchivedChannels,
   shouldConsiderPinnedChannels,
+  uniqBy,
 } from './utils';
 
 export type ChannelManagerPagination<SCG extends ExtendableGenerics = DefaultGenerics> = {
@@ -275,7 +276,7 @@ export class ChannelManager<SCG extends ExtendableGenerics = DefaultGenerics> {
   };
 
   public loadNext = async () => {
-    const { pagination, channels, initialized } = this.state.getLatestValue();
+    const { pagination, initialized } = this.state.getLatestValue();
     const { filters, sort, options, isLoadingNext, hasNext } = pagination;
 
     if (!initialized || isLoadingNext || !hasNext) {
@@ -283,16 +284,17 @@ export class ChannelManager<SCG extends ExtendableGenerics = DefaultGenerics> {
     }
 
     try {
-      const { offset, limit } = { ...DEFAULT_CHANNEL_MANAGER_PAGINATION_OPTIONS, ...options };
+      const { limit } = { ...DEFAULT_CHANNEL_MANAGER_PAGINATION_OPTIONS, ...options };
       this.state.partialNext({
         pagination: { ...pagination, isLoading: false, isLoadingNext: true },
       });
       const nextChannels = await this.client.queryChannels(filters, sort, options, this.stateOptions);
-      const newOffset = offset + (nextChannels?.length ?? 0);
-      const newOptions = { ...options, offset: newOffset };
+      const { channels } = this.state.getLatestValue();
+      const newChannels = uniqBy([...(channels || []), ...nextChannels], 'cid');
+      const newOptions = { ...options, offset: newChannels.length };
 
       this.state.partialNext({
-        channels: [...(channels || []), ...nextChannels],
+        channels: newChannels,
         pagination: {
           ...pagination,
           hasNext: (nextChannels?.length ?? 0) >= limit,
--- a/src/channel_manager.ts
+++ b/src/channel_manager.ts
@@ -284,17 +284,17 @@ export class ChannelManager<SCG extends ExtendableGenerics = DefaultGenerics> {
     }
 
     try {
-      const { limit } = { ...DEFAULT_CHANNEL_MANAGER_PAGINATION_OPTIONS, ...options };
+      const { offset, limit } = { ...DEFAULT_CHANNEL_MANAGER_PAGINATION_OPTIONS, ...options };
       this.state.partialNext({
         pagination: { ...pagination, isLoading: false, isLoadingNext: true },
       });
       const nextChannels = await this.client.queryChannels(filters, sort, options, this.stateOptions);
       const { channels } = this.state.getLatestValue();
-      const newChannels = uniqBy<Channel<SCG>>([...(channels || []), ...nextChannels], 'cid');
-      const newOptions = { ...options, offset: newChannels.length };
+      const newOffset = offset + (nextChannels?.length ?? 0);
+      const newOptions = { ...options, offset: newOffset };
 
       this.state.partialNext({
-        channels: newChannels,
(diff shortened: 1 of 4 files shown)


isekovanic:
This also makes it evident that `offset` is pretty much never going to be correct as `state.channels` gets populated from various WS events. In my eyes, we have 2 options:

- We ignore this and only really use it as a reference pointer to what the last offset was since we anyway calculate the correct one before each query
- We can add an update to `offset` on every `setChannels` invocation (a simple size calculation) and then have it be a lot more reliable

Thoughts ?

src/utils.ts
```diff
@@ -602,6 +602,27 @@ export const throttle = <T extends (...args: unknown[]) => unknown>(
...
+const get = <T>(obj: T, path: string): unknown =>
+  path.split('.').reduce<unknown>((acc, key) => {
+    if (acc && typeof acc === 'object' && key in acc) {
+      return (acc as Record<string, unknown>)[key];
+    }
+    return undefined;
+  }, obj);
+
+// works exactly the same as lodash.uniqBy
+export const uniqBy = <T>(array: T[] | unknown, iteratee: ((item: T) => unknown) | keyof T): T[] => {
```
isekovanic, line 614: Not the same implementation as in `lodash` obviously and it does not use their fast `SetCache`, but close enough for such small lists. Still ~ `O(n)`.

src/channel_manager.ts
```diff
@@ -275,24 +276,25 @@ export class ChannelManager<SCG extends ExtendableGenerics = DefaultGenerics> {
...
+      const { limit } = { ...DEFAULT_CHANNEL_MANAGER_PAGINATION_OPTIONS, ...options };
       this.state.partialNext({
         pagination: { ...pagination, isLoading: false, isLoadingNext: true },
       });
       const nextChannels = await this.client.queryChannels(filters, sort, options, this.stateOptions);
-      const newOffset = offset + (nextChannels?.length ?? 0);
-      const newOptions = { ...options, offset: newOffset };
+      const { channels } = this.state.getLatestValue();
+      const newChannels = uniqBy<Channel<SCG>>([...(channels || []), ...nextChannels], 'cid');
+      const newOptions = { ...options, offset: newChannels.length };
```
szuperaz, line 294: I think offset should be `prevOffset + nextChannels.length` set before deduplication, example:
- We load first page - offset is 10
- We receive 5 channels via WS events
- We load second page - offset is 25 -> but offset should be 20 at this point otherwise we might skip channels with offset 20-25

This approach minimizes missing channels because of gaps in `offset`

src/utils.ts
```diff
@@ -602,6 +602,27 @@ export const throttle = <T extends (...args: unknown[]) => unknown>(
...
+  if (!Array.isArray(array)) return [];
+
+  const seen = new Map<unknown, boolean>();
```
oliverlaz, line 617: nit-picking, but I think you can go with `new Set<unknown>()` here.

src/channel_manager.ts
MartinCupela: @isekovanic Do we really skip channels that were originally on positions 21-25 or have they moved down to positions 26-30 on the backend due to the changes reflected in the WS events?

szuperaz: We had a discussion about this previously: https://getstream.slack.com/archives/CE5N802GP/p1699387959700559?thread_ts=1698762949.051059&cid=CE5N802GP

src/utils.ts
isekovanic: Totally ! I think I started off with something else in mind and then just never changed it. Thanks !

src/channel_manager.ts
szuperaz: @MartinCupela we don't know how adding a new channel changes pagination on the backend, that depends on the sort params integrators use, `last_updated, last_message_at,updated_at,created_at,member_count, unread_count or has_unread (unread status). Direction can be ascending (1) or descending (-1)` -> your logic probably works in some cases, but not if someone uses something like `member_count` or `unread_count` or `has_unread`.

I think the safest option is to be conservative about increasing offset and filter duplicated channels.

isekovanic: So, I gave this some thought yesterday and came to the following conclusion.

The way I see it, there are 2 possible scenarios with this:

1. We receive WS events for channels that were not initially part of any pagination page. This means that if we've loaded 10 channels and we receive 5, the point from which pagination should proceed is 15 since 10 is no longer the index of the last channel and we would be loading less items with the next page (because the ones 10 - 15 would be deduplicated, so we'd only get 5 whereas we should be getting 10). So basically the new `channels.length` would be the correct `offset`
2. We receive WS events for channels that are part of one of the next pages (for the sake of brevity, let's say it's the very next one). Then, if we load 10 and then 5 from the next page move upwards, we end up having again the pagination proceed from 25, which is the correct index as the next channel that we're interested in starts there since deduplication kicks in. Starting from index 20 means that the state will still be correct (because of deduplication) but it also means that we won't receive 10 channels in a single page but rather receive 5.

Or am I missing something deeply here ? As far as I know BE should reflect the new offsets for each channel since if a WS event arrives the DB had already been updated, and that's how the offset is used

isekovanic: Ah, @szuperaz comment arrived late for me - let me read through it one sec

szuperaz: @isekovanic yes, your logic is correct, but that's better than missing out on channels

isekovanic: Also does that mean that we can pretty much never do this reliably ? Since in that scenario it would be wrong to update the `offset` and it would pretty much always be wrong.

isekovanic: So just to confirm, if a WS event arrives for a channel and it indeed moves up, this does not necessarily mean that it would reflect the actual backend `offset`, right ?

In the general scenario it should, but considering we're allowing overrides for various things it is entirely possible that it simply doesn't reflect on this.

szuperaz: > Yep yep, for sure - I'm just saying if channels from page 2 have all been deduplicated we would get a pagination indicator -> no update -> we paginate again -> only then do we get an update.

We can also solve this by not setting `isLoading` to `false` until the end of the second query channels call. I think in that case the user should notice anything.

isekovanic: Yeah definitely, this is all solvable on the UI itself, but for an empty response only - if we receive 1 instead of 0 then we shouldn't run run `loadNext()` again. But these are all very very odd edge cases, integrators mostly paginate with the highest number of channels possible (which is 30)

isekovanic: In other words, this will not really be noticeable by users at all

MartinCupela: > Yeah definitely, this is all solvable on the UI itself, but for an empty response only - if we receive 1 instead of 0

@isekovanic  You should not run loadNext if the limit parameter is > returnedChannels.length.

isekovanic: This is already accounted for [here](https://github.com/GetStream/stream-chat-js/blob/e6b2b815d5d6e2d808b5260ea9f609748db9789f/src/channel_manager.ts#L281) thought, right ?

isekovanic: `hasNext` is calculated from the actual (non-deduplicated) response

MartinCupela: @isekovanic I did not understand what you meant by:

> if we receive 1 instead of 0 then we shouldn't run run loadNext() again

szuperaz (approved):

MartinCupela (approved):

PR #1468, opened 2025-02-14
chore: update build system

arnautov-anton:


scripts/bundle-cjs.mjs
```diff
@@ -0,0 +1,71 @@
...
+const browserPlugin = {
+  name: 'ignoreBrowser',
+  setup: (build) => {
+    const options = build.initialOptions;
+
+    if (options.platform !== 'browser') return;
+
+    build.onResolve({ filter: /(node:.+)|(jsonwebtoken)/ }, async ({ path }) => ({ path, external: true }));
+  },
+};
```
arnautov-anton, line 50: I'm not sure if this is the right approach, previous builds seemed to keep the imports (of `crypto`, `https` and `jsonwebtoken`) even though they shouldn't be there in my opinion.

arnautov-anton: Nevermind, [this](https://github.com/GetStream/stream-chat-js/pull/1468/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R22-R26) seems to work just as well. :)

package.json
```diff
@@ -29,15 +41,9 @@
     "getstream.io"
   ],
   "files": [
-    "/dist",
-    "/src",
-    "readme.md",
-    "license"
+    "/dist"
```
arnautov-anton, line 44: From [npm documentation](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files): 

Certain files are always included, regardless of settings:

- package.json
- README
- LICENSE / LICENCE
- The file in the "main" field
- The file(s) in the "bin" field

Not sure why we were releasing sourcefiles as well though.

package.json
```diff
@@ -46,46 +52,38 @@
...
-    "@babel/preset-env": "^7.16.4",
-    "@babel/preset-typescript": "^7.16.0",
-    "@babel/register": "^7.16.0",
+    "@babel/plugin-proposal-class-properties": "^7.18.6",
+    "@babel/plugin-proposal-object-rest-spread": "^7.20.7",
+    "@babel/plugin-transform-object-assign": "^7.25.9",
+    "@babel/plugin-transform-runtime": "^7.26.9",
+    "@babel/preset-env": "^7.26.9",
+    "@babel/preset-typescript": "^7.26.0",
+    "@babel/register": "^7.25.9",
```
arnautov-anton, line 61: Unfortunately these are still needed for the Mocha to run properly, using different loader generated a TON of other issues I wasn't comfortable dealing with within this PR, updates should suffice for now.

tsconfig.json
```diff
@@ -1,21 +1,30 @@
...
     "noImplicitAny": true,
-    "preserveConstEnums": true,
     "strict": true,
     "strictNullChecks": true,
     "declaration": true,
-    "declarationMap": true,
-    "declarationDir": "./dist/types",
-    "module": "commonjs",
-    "target": "ES5"
+    "module": "ES2020",
```
arnautov-anton, line 25: Note - important change here, the default output is now ESM.

tsconfig.json
```diff
@@ -1,21 +1,30 @@
...
+    "target": "ES6",
```
arnautov-anton, line 26: We should probably review the target too.

.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
oliverlaz, line 1: Let's stick with actual versions here - like `v22`

.github/actions/setup-node/action.yml
```diff
@@ -3,11 +3,12 @@ description: Sets up Node and Build SDK
 
 inputs:
   node-version:
+    description: "Specify Node version"
     required: false
-    default: '16'
+    default: "lts/*"
```
oliverlaz, line 8: Every October this may surprise us :)
Let's use a fixed version (v22 for example).

package.json
oliverlaz: Sources are good convenience... I personally like it when they are around, as I can easily poke around with my IDE instead of using GitHub or clone the repo and make sure I'm on the correct tag.

package.json
oliverlaz: What issues did you face? We managed to easily solve this in GetStream/chat QA.

scripts/bundle-cjs.mjs
```diff
@@ -0,0 +1,55 @@
...
+  ...packageJson.peerDependencies,
+});
+const external = deps.filter((dep) => !bundledDeps.includes(dep));
+
+/** @type esbuild.BuildOptions */
+const cjsBundleConfig = {
+  entryPoints: [mainEntrypoint],
+  bundle: true,
+  format: 'cjs',
+  target: "ES6",
```
oliverlaz, line 33: Can we use `ES2020` as a target everywhere?

tsconfig.json
oliverlaz: Yeah, let's use ES2020 for the time being. It aligns with the other JS SDKs.

package.json
arnautov-anton: After installing `ts-node` and replacing the loader (through the `--require` Mocha option) you're immediately after running greeted with `Exception during run: Error: Cannot require() ES Module` - if you rename `.js` to `.mjs` you get something else and after that it has issue with our tests importing from `src` at which point I stopped pursuing.

arnautov-anton: The QA seems to have all the tests in TS, the tests here are a combination of both - at some point I was able to run tests (with some different loader, can't remember now) but the TS was failing on completely bogus errors which are not detected normally.

arnautov-anton: Let me check one last time with separate `tsconfig` for testing only.

arnautov-anton: [baa9758](https://github.com/GetStream/stream-chat-js/pull/1468/commits/baa97582fb75a72eded9090bdfd2af841e2048d3)

package.json
arnautov-anton: I rarely check the `dist`'s for sources but I'll return them if you think it's convenience. 👍

scripts/bundle.mjs
```diff
@@ -26,30 +20,45 @@ const deps = Object.keys({
...
+    ...commonBuildOptions,
+    entryPoints: [resolve(__dirname, '../src/**/*')],
+    format: 'esm',
+    outdir: resolve(__dirname, '../dist/esm'),
+    entryNames: `[dir]/[name]`,
+    outExtension: { '.js': '.mjs' },
+    platform: 'browser',
+    define: {
+      'process.env.PKG_VERSION': JSON.stringify(packageJson.version),
+    },
```
myandrienko: We can move it to `commonBuildOptions`.

scripts/bundle.mjs
```diff
@@ -26,30 +20,45 @@ const deps = Object.keys({
...
     platform,
     define: {
       'process.env.PKG_VERSION': JSON.stringify(packageJson.version),
     },
-  };
+  })),
+  // ESM (browser only)
+  {
+    ...commonBuildOptions,
+    entryPoints: [resolve(__dirname, '../src/**/*')],
```
myandrienko, line 51: Could you make sure this is not too broad, e.g. tests shouldn't be included.

tsconfig.json
myandrienko: I _think_ we can do `"module": "preserve"` (`"moduleResolution": "bundler"` is implied), since we're indeed using a bundler. This should introduce less uncertainty.

myandrienko (approved):

tsconfig.json
arnautov-anton: Adjusted in [c35c23d](https://github.com/GetStream/stream-chat-js/pull/1468/commits/c35c23d164caab04630ad3fae5ffac8bddeb5f78) and after another round of manual testing everything else seems to work fine too. 👍

scripts/bundle.mjs
arnautov-anton: Done. :)

scripts/bundle.mjs
arnautov-anton: Also done.

PR #1477, opened 2025-02-28
fix: add extra user agent fields resolvable by react native

isekovanic:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2944,7 +2944,11 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
       ['device_model', model],
       // reports which bundle is being picked from the exports
       ['client_bundle', clientBundle],
-    ] as const).reduce((acc, [key, value]) => (value ? acc.concat(`|${key}=${value ?? ''}`) : acc), userAgentString);
+    ] as const).reduce(
+      (withArguments, [key, value]) =>
+        value && value.length > 0 ? withArguments.concat(`|${key}=${value}`) : withArguments,
+      userAgentString,
+    );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


szuperaz (approved):

isekovanic:
@arnautov-anton Added your V9 changes [here](https://github.com/GetStream/stream-chat-js/pull/1477/commits/8c35f25807c8caee516cc200131f25e6549a99ed), thanks a lot for pointing it out !

src/client.ts
```diff
@@ -2922,12 +2924,27 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+    const { os, model } = this.deviceIdentifier ?? {};
+
+    return ([
+      // reports the device OS, if provided
+      ['os', os],
+      // reports the device model, if provided
+      ['device_model', model],
+      // reports which bundle is being picked from the exports
+      ['client_bundle', clientBundle],
+    ] as const).reduce((acc, [key, value]) => (value ? acc.concat(`|${key}=${value ?? ''}`) : acc), userAgentString);
```
arnautov-anton, line 2947:
```suggestion
    ] as const).reduce((acc, [key, value]) => (value ? acc.concat(`|${key}=${value}`) : acc), userAgentString);
```
Value should be defined here so we can skip this I think.

isekovanic: Ah, of course; thanks !

arnautov-anton, line 2947: nit:
```suggestion
    ] as const).reduce((withArguments, [key, value]) => (value ? withArguments.concat(`|${key}=${value ?? ''}`) : withArguments), userAgentString);
```

PR #1496, opened 2025-03-13
feat: add reminders

nijeesh-stream:
## Goal

For the reminders the following is added:
- Reminder types
- API request methods
- reactive state managed by RemindersManager
- timers for reminders
- reminder pagination with reactive state

### DIFF
--- a/src/pagination/BasePaginator.ts
+++ b/src/pagination/BasePaginator.ts
@@ -26,29 +26,6 @@ export type PaginatorState<T = any> = {
   offset?: number;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export interface IPaginator<T = any> {
-  cancelScheduledQuery(): void;
-  canExecuteQuery(direction: PaginationDirection): boolean;
-  next(): Promise<void>;
-  nextDebounced(): void;
-  prev(): Promise<void>;
-  prevDebounced(): void;
-  query(params: PaginationQueryParams): Promise<PaginationQueryReturnValue<T>>;
-  readonly hasNext: boolean;
-  readonly hasPrev: boolean;
-  readonly hasResults: boolean;
-  readonly initialState: PaginatorState<T>;
-  readonly isLoading: boolean;
-  readonly items: T[] | undefined;
-  readonly lastQueryError: Error | undefined;
-  readonly cursor: Cursor | undefined;
-  readonly offset: number | undefined;
-  resetState(): void;
-  setDebounceOptions(options: PaginatorDebounceOptions): void;
-  readonly state: StateStore<PaginatorState<T>>;
-}
-
 export type PaginatorOptions = {
   /** The number of milliseconds to debounce the search query. The default interval is 300ms. */
   debounceMs?: number;
@@ -59,7 +36,7 @@ export const DEFAULT_PAGINATION_OPTIONS: Required<PaginatorOptions> = {
   pageSize: 10,
 } as const;
 
-export abstract class BasePaginator<T> implements IPaginator<T> {
+export abstract class BasePaginator<T> {
   state: StateStore<PaginatorState<T>>;
   protected pageSize: number;
   protected _executeQueryDebounced!: DebouncedExecQueryFunction;


totalimmersion (approved):

src/client.ts
```diff
@@ -4097,4 +4102,64 @@ export class StreamChat<StreamChatGenerics extends ExtendableGenerics = DefaultG
...
+  }
+
+  /**
+   * updateReminder - Updates an existing reminder for a message
+   *
+   * @param {string} messageId The ID of the message whose reminder to update
+   * @param {UpdateReminderOptions} options The options for updating the reminder
+   * @returns {Promise<ReminderResponse>}
+   */
+  async updateReminder<StreamChatGenerics extends ExtendableGenerics = DefaultGenerics>(
```
MartinCupela, line 4130: @nijeesh-stream please remove the generics as they are not supported by the SDK anymore. Thank you

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+    overwrite?: boolean;
+  }) => {
+    if (!this.client._cacheEnabled()) {
+      return;
+    }
+    const cachedReminder = this.getFromState(data.message_id);
+    if (!cachedReminder) {
+      const reminder = new Reminder({ data });
+      this.state.partialNext({
+        reminders: new Map(this.reminders.set(data.message_id, reminder)),
```
isekovanic, line 118: Why are we creating a new `Map` every time ? If we're following the way `polls` work, the manager should simply update the existing reminder within the map (or alternatively insert it if it doesn't exist); this will cause the entire state to get recreated every time we `upsertToState`

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+    return cachedReminder;
+  };
+
+  removeFromState = (messageId: string) => {
+    const cachedReminder = this.getFromState(messageId);
+    if (!cachedReminder) return;
+    this.timers.remove(cachedReminder.id);
+    const reminders = this.reminders;
+    reminders.delete(messageId);
+    this.state.partialNext({ reminders: new Map(reminders) });
```
isekovanic, line 132: Same deal here

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+  };
+
+  hydrateState = (messages: MessageResponse[]) => {
```
isekovanic: Would it be possible for this to be `LocalMessage[]` ? It would give us more flexibility in the future in case hydration needs to happen from something other than a `MessageResponse` (the `message.reminder` value should probably stay the same. All good if it's not viable though, just an afterthought

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+    this.timers.clearAll();
+  };
+  // Timers API END //
+
+  // WS event handling START //
+  static isReminderWsEventPayload = (event: Event): event is ReminderEvent =>
+    !!event.reminder &&
+    (event.type.startsWith('reminder.') || event.type === 'notification.reminder_due');
+
+  public registerSubscriptions = () => {
```
isekovanic, line 159: Perhaps we could add something like:

```
if (this.hasSubscriptions) {
      return undefined;
    }
```
in order to guard against multiple invocations of this.

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+        if (!items) return;
+        for (const reminder of items) {
+          this.upsertToState({ data: reminder });
+        }
+      },
+    );
+  // WS event handling END //
+
+  // API calls START //
+  createOrUpdateReminder = async (options: CreateReminderOptions) => {
```
isekovanic:
```suggestion
  upsertReminder = async (options: CreateReminderOptions) => {
```
maybe sounds better ?

src/reminders/ReminderTimerManager.ts
```diff
@@ -0,0 +1,121 @@
...
+    return refreshInterval;
+  };
+
+  add = (reminder: Reminder) => {
+    const existingTimer = this.timers.get(reminder.id);
+    const timeoutLength = ReminderTimerManager.getRefreshIntervalLength(reminder);
+    if (timeoutLength === null || !reminder.remindAt) return;
+    const boundaryTimestamp =
+      reminder.remindAt?.getTime() + this.config.stopRefreshBoundaryMs;
+    const timeLeftToBoundary = boundaryTimestamp - Date.now();
```
isekovanic: Just curious, how do `date.getTime()` and similar APIs work with `ms` precision here ? Is it possible that we remove something prematurely because of how dates are stored on the BE (`ns` precision) ? It probably doesn't matter at all but just a thought.

src/reminders/ReminderTimerManager.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (timeLeftToBoundary < 0) {
```
isekovanic: *nitpick*
```suggestion
    if (timeLeftToBoundary <= 0) {
```

src/reminders/ReminderManager.ts
```diff
@@ -0,0 +1,264 @@
...
+    const cachedReminder = this.getFromState(data.message_id);
+    if (!cachedReminder) {
+      const reminder = new Reminder({ data });
+      this.state.partialNext({
+        reminders: new Map(this.reminders.set(data.message_id, reminder)),
+      });
+      this.timers.addAll([reminder]);
+    } else if (overwrite) {
+      cachedReminder.setState(data);
+      this.timers.addAll([cachedReminder]);
```
isekovanic: *nitpick*
```suggestion
      this.timers.add(cachedReminder);
```

src/reminders/ReminderTimerManager.ts
```diff
@@ -0,0 +1,121 @@
...
+  minute: { lower: oneMinute, upper: oneHour },
+  hour: { lower: oneHour, upper: oneDay },
+  day: { lower: oneDay, upper: oneWeek },
+} as const;
+
+export const DEFAULT_REMINDER_TIMER_MANAGER_CONFIG: ReminderTimerManagerConfig = {
+  stopRefreshBoundaryMs: 2 * oneWeek,
+};
+
+export class ReminderTimerManager {
```
isekovanic: Why are the timers not part of actual `reminder` state or something similar ? I believe that way you wouldn't need to keep 2 separate maps that are managed but rather have the `ReminderTimerManager` manage a single timer, while the `Reminder` class itself would be responsible for using this manager to do stuff.

src/reminders/ReminderManager.ts
MartinCupela: If you add a new reminder, you need reactivity. If we just added the reminder to the existing map, the already rendered message would not react to that.

MartinCupela: Maybe I could split the upsertToState method into update (non-reactive addition to existing map) and insert (creates a new map)

src/reminders/ReminderManager.ts
MartinCupela: 👍 will adjust also above - this is a remnant of a refactor.

src/reminders/ReminderManager.ts
MartinCupela: If we do not create a new map, the already rendered message will not be able to tell, that the reminder has been removed from the map.

src/reminders/ReminderManager.ts
MartinCupela: Good idea. I would put `MessageResponse[] | LocalMessage[]`

src/reminders/ReminderManager.ts
MartinCupela: Good idea

src/reminders/ReminderManager.ts
MartinCupela: ok

src/reminders/ReminderTimerManager.ts
MartinCupela: BE stores in microseconds, JS cuts to milliseconds. But should not be observable by users.

src/reminders/ReminderTimerManager.ts
MartinCupela: 👍

src/reminders/ReminderTimerManager.ts
MartinCupela: Yes, good idea.

src/reminders/ReminderManager.ts
MartinCupela: Nope, I will not be splitting this. Actually the method is creating new map only when adding the Reminder to the map and that is correct. As already said, without doing that, we would not have reactivity when adding new remiders. On the already existing reminders we are just setting their state.

isekovanic: Ah, I see - yeah that's fair. My concern here is the fact that technically the map's values would also change (referentially) whenever this happens; meaning - whenever a reminder is added all other reminders would also update by reference needlessly. I unfortunately don't have a better solution for this myself, due to how `React`/`RN` work in general and we have to focus on copy-on-write functionality. 

I'm curious though, why do we need to react to the entire `Map` changing ? I'm not 100% familiar with how reminders are supposed to be displayed on the UI

src/reminders/ReminderTimerManager.ts
MartinCupela: done: https://github.com/GetStream/stream-chat-js/pull/1496/commits/ab572dae862b2b4326e9e7447f25599babb5f08e

src/reminders/ReminderManager.ts
MartinCupela: > I'm curious though, why do we need to react to the entire Map changing ? 

I would respond with a question: What other reactive state can a Message component observe so that it re-renders and at the same time we do not have 5-6 WS event subscribers for each reminder?

We could initiate a Reminder instance in a ReminderNotification component. But every of those components would have to create those subscriptions. The manager decreases the number of subscriptions to the minimum.

isekovanic: I meant it more in a sense that the map should function as a lookup table for each reminder, and not directly change as a reminder is added/removed - but I think what you mean here is that you generally need some structure to hold the reminders as well as cache them (similar to how we have a list for channels for example). 

I get it now ! Resolving this issue.

isekovanic (approved):
Looks great to me ! Thanks !

PR #1505, opened 2025-03-31
fix: remove asynchronously moderated message

arnautov-anton:

A customer reported that after posting a message with content that gets moderated right after (blocked/removed), it stays in their message list with an error icon - to mitigate this issue the customer decided to hide the message conditionally (do not render a message of type `error`) - which is wrong as that also disables the ability to retry sending messages which failed due to poor network conditions (or some other error). The other issue was that the list would keep a date separator rendered since the message was still part of the list.

To solve the problem I decided to remove the message from the state through `message.updated` event handler if it's final state is of type `error`. Other messages (retryable) wouldn't be affected - both because they're currently not part of the client's channel state (at least for React SDK) and because the `error` type is not being set by the `message.updated` event handler.

Now - I'm pretty sure that "Shadow Block" wouldn't be affected by this change (as `message.updated` should not be delivered to the offender) but feel free to correct me, @vishalnarkhede.


szuperaz:
@arnautov-anton, @vishalnarkhede, how can I test this sequence? I tried to reproduce async moderation, but I only found a way to delete a message - in which case a `message.deleted` event arrived and the message was removed. Is this a different flow?

Screenshot 2025-04-01 at 9 44 27

vishalnarkhede:
@szuperaz you need to upload some harmful image to test. You can choose some image from here: https://dashboard.getstream.io/app/1320873/moderation/media

szuperaz:
@arnautov-anton, @vishalnarkhede thank you for your help, I was able to test this, but something is not clear to me:
- If I reload the page, the async moderated message is gone -> so I assume the message is deleted -> but then why don't I get a `message.deleted` event?
- Is it to give a chance to SDKs to implement a UI where we don't remove the message just display "this message was removed due to moderation policies"? -> In that case we shouldn't remove the message from the list in stream-chat-js. But we can't display this UI because the `message.updated` event doesn't have a moderation payload, so we don't know what's wrong with the message

arnautov-anton:
Closing in favor of [different behavior](https://getstream.slack.com/archives/C07EV8DFMRN/p1743597792603759) (according to spec).

PR #1510, opened 2025-04-03
fix: [REACT-218] adjust filterErrorMessages to only filter current user's messages

arnautov-anton:

Ref: https://github.com/GetStream/stream-chat-react/pull/2675

### DIFF
--- a/src/channel_state.ts
+++ b/src/channel_state.ts
@@ -581,7 +581,17 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
    *
    */
   filterErrorMessages() {
-    const filteredMessages = this.latestMessages.filter((message) => message.type !== 'error');
+    const userId = this._channel.getClient().userID;
+
+    if (!userId) return;
+
+    const filteredMessages = this.latestMessages.filter(
+      (message) =>
+        !(
+          message.type === 'error' &&
+          (message.user?.id === userId || message.user_id === userId)
+        ),
+    );
 
     this.latestMessages = filteredMessages;
   }


khushal87 (approved):
What was the purpose of this change? It looks fine anyways.

src/channel_state.ts
```diff
@@ -581,7 +581,17 @@ export class ChannelState<StreamChatGenerics extends ExtendableGenerics = Defaul
...
-    const filteredMessages = this.latestMessages.filter((message) => message.type !== 'error');
+    const userId = this._channel.getClient().userID;
+
+    if (!userId) return;
+
+    const filteredMessages = this.latestMessages.filter(
+      (message) =>
+        !(
+          message.type === 'error' &&
+          (message.user?.id === userId || message.user_id === userId)
```
szuperaz, line 592: @arnautov-anton can we add this as a param like `onlyOwnMessages` or similar? To avoid changing the method's behavior. Btw this is a bit of a weird behavior because this means that async moderated messages will be filtered for the sender, but not for receivers. But I understand the reason for this.

szuperaz:
@arnautov-anton how does all this work for thread messages?

arnautov-anton:
> What was the purpose of this change? It looks fine anyways.

My apologies, @khushal87, forgot to reference the relevant PR at first - it's connected to the [`BlockedMessage` feature](https://github.com/GetStream/stream-chat-react/pull/2675), see also [this Slack message](https://getstream.slack.com/archives/C07EV8DFMRN/p1743597792603759).

> @arnautov-anton how does all this work for thread messages?

I'm not sure it does, @szuperaz. Either way I'm really not happy with this fix (or the existing behavior), I've drafted this up based on our conversation but while I was testing it out it feels clunky at best - I'd rather we go with Angular's behavior where the error messages are kept for the duration of the application runtime. I'll release the `BlockedMessage` feature without this for now.

arnautov-anton:
Closing for now, I'll re-open once relevant again.

PR #1515, opened 2025-04-10
fix: make archived_at & pinned_at nullable

arnautov-anton:

Extend `ChannelMemberResponse`'s `archived_at` & `pinned_at` properties with `null`.


arnautov-anton:
> @arnautov-anton will it be `undefined` initially and `null` once someone removed an archive/pin?

Correct, when you unpin/unarchive `member.updated` returns a `member` object which has these properties set to `null`.

oliverlaz (approved):

PR #1516, opened 2025-04-14
fix: [REACT-353] unify pinned_at & archived_at nullish values

arnautov-anton:

Event handler for `member.updated` receives a `member` object which contains `archived_at` and `pinned_at` with `null` values if unarchive/unpin actions are called - to make matters for our integrators easier, the only nullish value for these would be `undefined`. I'll keep the types unchanged (#1515) as the `ChannelMemberResponse` used in the `Event` type is still correct.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1618,22 +1618,36 @@ export class Channel<StreamChatGenerics extends ExtendableGenerics = DefaultGene
         }
         break;
       case 'member.added':
-      case 'member.updated':
-        if (event.member?.user) {
+      case 'member.updated': {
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):

szuperaz:
@arnautov-anton won't we still receive `null` in `queryChannels` response?

arnautov-anton:
> @arnautov-anton won't we still receive `null` in `queryChannels` response?

No - it seems that this sort of data is only delivered through events.

`queryChannels` after pinning:
image
after unpinning:
image

szuperaz (approved):

PR #1522, opened 2025-05-04
refactor: general code cleanup (Threads/Polls/MessageComposer)

arnautov-anton:

Introduces private `WithSubscriptions` abstract class as a base for all of our new entities.

### DIFF
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -371,10 +371,6 @@ export class MessageComposer extends WithSubscriptions {
     this.addUnsubscribeFunction(this.subscribeCustomDataManagerStateChanged());
     this.addUnsubscribeFunction(this.subscribeMessageComposerStateChanged());
     this.addUnsubscribeFunction(this.subscribeMessageComposerConfigStateChanged());
-    if (this.config.drafts.enabled) {
-      this.addUnsubscribeFunction(this.subscribeDraftUpdated());
-      this.addUnsubscribeFunction(this.subscribeDraftDeleted());
-    }
 
     return this.unregisterSubscriptions;
   };
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -478,8 +478,7 @@ export class MessageComposer extends WithSubscriptions {
           return;
         }
 
-        const [previousText] = previousSelection;
-        if (!this.linkPreviewsManager.enabled || currentText === previousText) return;
+        if (!this.linkPreviewsManager.enabled) return;
 
         if (!currentText) {
           this.linkPreviewsManager.clearPreviews();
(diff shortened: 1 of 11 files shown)


src/messageComposer/messageComposer.ts
```diff
@@ -449,58 +448,73 @@ export class MessageComposer {
...
+        if (typeof previousSelection === 'undefined') return;
+
         this.logStateUpdateTimestamp();
+
         if (this.compositionIsEmpty) {
           this.deleteDraft();
           return;
         }
-      }
-      if (!this.linkPreviewsManager.enabled || nextValue.text === previousValue?.text)
```
arnautov-anton, line 471: Removed this part:

`nextValue.text === previousValue?.text`

as selected text is already being compared against previous value internally and unless it differs between state changes, the handler won't run.

src/messageComposer/messageComposer.ts
```diff
@@ -347,36 +352,30 @@ export class MessageComposer {
...
+    this.addUnsubscribeFunction(this.subscribeAttachmentManagerStateChanged());
+    this.addUnsubscribeFunction(this.subscribeLinkPreviewsManagerStateChanged());
+    this.addUnsubscribeFunction(this.subscribePollComposerStateChanged());
+    this.addUnsubscribeFunction(this.subscribeCustomDataManagerStateChanged());
+    this.addUnsubscribeFunction(this.subscribeMessageComposerStateChanged());
+    this.addUnsubscribeFunction(this.subscribeMessageComposerConfigStateChanged());
     if (this.config.drafts.enabled) {
-      this.unsubscribeFunctions.add(this.subscribeDraftUpdated());
-      this.unsubscribeFunctions.add(this.subscribeDraftDeleted());
+      this.addUnsubscribeFunction(this.subscribeDraftUpdated());
```
MartinCupela, line 372: Would be great to have instant access to a specific unsubscribe function. Here for example, what if 

1. the drafts are disabled, 
2. enabled, 
3. disabled back

There are two issues:
1. The composer is not enabling the draft WS subscriptions on config change (1. -> 2.)
2. The composer is not unsubscribing the WS subscriptions on config change (1. -> 3.

The first issue can be solved immediately in the config state change subscriber, but for the second we do not have the setup that would allows us to take the specific unsubscribe function reference and call it.

Could we introduce a possibility to associate unsubscribe functions with identifiers or store direct references to them?

MartinCupela: Added proposals here: https://github.com/GetStream/stream-chat-js/pull/1529

arnautov-anton: Copied suggestions over from #1529 in  [900178b](https://github.com/GetStream/stream-chat-js/pull/1522/commits/900178b8d1beabfe88f9f6a77527cb63c093cfaf).

MartinCupela (approved):

PR #1526, opened 2025-05-07
fix: add user to RESERVED_UPDATED_MESSAGE_FIELDS

arnautov-anton:

Update fails if the payload contains both `user` object and `user_id`property.

```
StreamChat error code 4: UpdateMessage failed with error: "cannot set both message.user and message.user_id."
```

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,31 +1,30 @@
-import type { ReservedUpdatedMessageFields } from './types';
-
 export const DEFAULT_QUERY_CHANNELS_MESSAGE_LIST_PAGE_SIZE = 25;
 export const DEFAULT_QUERY_CHANNEL_MESSAGE_LIST_PAGE_SIZE = 100;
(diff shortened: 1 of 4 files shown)


MartinCupela (approved):

src/constants.ts
```diff
@@ -1,31 +1,30 @@
...
-  'i18n',
-  'type',
-  'html',
-  '__html',
-] as const;
+  i18n: true,
+  type: true,
+  html: true,
+  __html: true,
+  user: true,
```
arnautov-anton, line 27: 📝

src/utils.ts
```diff
@@ -384,7 +384,6 @@ export const toUpdatedMessagePayload = (
     mentioned_users: message.mentioned_users?.map((user) =>
       typeof user === 'string' ? user : user.id,
     ),
-    user_id: message.user?.id ?? message.user_id,
```
arnautov-anton, line 387: 📝

PR #1529, opened 2025-05-09
feat: allow to register subscriptions by key

MartinCupela:
Suggestions regarding comments:

- https://github.com/GetStream/stream-chat-js/pull/1522/files#r2081230871

### DIFF
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -544,17 +544,24 @@ export class MessageComposer extends WithSubscriptions {
     });
 
   private subscribeMessageComposerConfigStateChanged = () =>
-    this.configState.subscribeWithSelector(
-      ({ text }) => [text] as const,
-      ([currentText]) => {
-        if (this.textComposer.text === '' && currentText.defaultValue) {
-          this.textComposer.insertText({
-            text: currentText.defaultValue,
-            selection: { start: 0, end: 0 },
-          });
-        }
-      },
-    );
+    this.configState.subscribe((nextValue, previousValue) => {
+      const { text } = nextValue;
+      if (this.textComposer.text === '' && text.defaultValue) {
+        this.textComposer.insertText({
+          text: text.defaultValue,
+          selection: { start: 0, end: 0 },
+        });
+      }
+
+      if (!previousValue?.drafts.enabled && nextValue.drafts.enabled) {
+        this.addUnsubscribeFunction(this.subscribeDraftUpdated(), 'draft.updated');
+        this.addUnsubscribeFunction(this.subscribeDraftDeleted(), 'draft.deleted');
+      }
+      if (previousValue?.drafts.enabled && !nextValue.drafts.enabled) {
+        this.unregisterSubscriptionsByKey('draft.updated');
+        this.unregisterSubscriptionsByKey('draft.deleted');
+      }
+    });
 
   setQuotedMessage = (quotedMessage: LocalMessage | null) => {
     this.state.partialNext({ quotedMessage });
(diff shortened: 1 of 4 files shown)


src/messageComposer/messageComposer.ts
```diff
@@ -544,17 +544,24 @@ export class MessageComposer extends WithSubscriptions {
...
+
+      if (!previousValue?.drafts.enabled && nextValue.drafts.enabled) {
+        this.addUnsubscribeFunction(this.subscribeDraftUpdated(), 'draft.updated');
+        this.addUnsubscribeFunction(this.subscribeDraftDeleted(), 'draft.deleted');
+      }
+      if (previousValue?.drafts.enabled && !nextValue.drafts.enabled) {
+        this.unregisterSubscriptionsByKey('draft.updated');
+        this.unregisterSubscriptionsByKey('draft.deleted');
+      }
+    });
```
arnautov-anton, line 564: I understand what you're trying to say but I'm not sure if I like this approach. State-dependent subscriptions can be resolved using subscribe method's scope (I've used similar approach [here](https://github.com/GetStream/stream-chat-js/pull/1423/files#diff-3cbf5743568cdbfb77280d8a0287d76a70f3584c67977c1a979eef698f1546b0R219-R244)):

```ts
private subscribeMessageComposerConfigStateChanged = () => {
  let draftUnsubscribeFunctions: Unsubscribe[] | null;

  const unsubscribe = this.configState.subscribeWithSelector(
    (currentValue) => ({
      textDefaultValue: currentValue.text.defaultValue,
      draftsEnabled: currentValue.drafts.enabled,
    }),
    ({ textDefaultValue, draftsEnabled }) => {
      if (this.textComposer.text === '' && textDefaultValue) {
        this.textComposer.insertText({
          text: textDefaultValue,
          selection: { start: 0, end: 0 },
        });
      }

      if (draftsEnabled && !draftUnsubscribeFunctions) {
        draftUnsubscribeFunctions = [
          this.subscribeDraftUpdated(),
          this.subscribeDraftDeleted(),
        ];
      } else if (!draftsEnabled && draftUnsubscribeFunctions) {
        draftUnsubscribeFunctions.forEach((fn) => fn());
        draftUnsubscribeFunctions = null;
      }
    },
  );

  return () => {
    draftUnsubscribeFunctions?.forEach((unsubscribe) => unsubscribe());
    unsubscribe();
  };
};
```

With key-based approach (at least in this form) we're limited to one unsubscribe function per key, if we decide to make it a `Map<string, Set<Unsubscribe>>` (to allow registering more of the same subscription) then we're introducing another unnecessary complexity IMO. I'd always hold onto the functions we get from registering subscriptions as true "keys" per-se and utilize function/method scopes as much as possible.

arnautov-anton: One extra - say if we had `subscribeWhatever(someInput: unknown)` and your `registerSubscriptions` would look something like this: 

```ts
public registerSubscriptions = () => {
  if (this.unsubscribeFunctions.size) {
    return noop;
  }
  this.unsubscribeFunctions.add(this.subscribeWhatever(1));
  this.unsubscribeFunctions.add(this.subscribeWhatever(2));
}
```

Then... if we went key-based route, you'd have to make sure this `someInput` (of whatever type) would have to be considered when you're constructing the key otherwise you could overwrite the unsubscribe function it holds (in its simple form).

MartinCupela: Understood, it all makes sense. Let's go the scope way. Do you plan to introduce it in your PR?

arnautov-anton: I'll cherry-pick your changes and adjust, thanks. 👍

PR #1530, opened 2025-05-09
fix: add optional raw queryChannels api response

isekovanic:
## CLA

- [ ] Code changes are tested

This PR allows `client.queryChannels` to optionally return the raw response and not run hydration at all (and subsequently not provide an array of `Channel` instances).

It is meant to be used generally server-side so that integrators can have access to other fields not available in a `Channel` instance. It should resolve [this Zendesk ticket](https://getstream.zendesk.com/agent/tickets/64585). 

We can probably keep it the same api if we use discriminated unions, but I feel like it'd complicate things further for no good reason. Keeping the apis separate made sense to me.

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1735,23 +1735,19 @@ export class StreamChat {
   }
 
   /**
-   * queryChannels - Query channels
+   * queryChannelsRequest - Queries channels and returns the raw response
    *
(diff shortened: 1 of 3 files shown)


arnautov-anton (approved):
🌮

PR #1531, opened 2025-05-12
fix: check for max votes allowed respecting a value from 2-10

khushal87:
Respect the value of `max_votes_allowed` to be in the range 2-10. Right now, if you type 1, the `error` didn't complain.

### DIFF
--- a/src/messageComposer/middleware/pollComposer/state.ts
+++ b/src/messageComposer/middleware/pollComposer/state.ts
@@ -31,7 +31,7 @@ export const pollStateChangeValidators: Partial<
   max_votes_allowed: ({ data, value }) => {
     if (data.enforce_unique_vote && value)
       return { max_votes_allowed: 'Enforce unique vote is enabled' };
-    if (value?.length > 1 && !value.match(VALID_MAX_VOTES_VALUE_REGEX))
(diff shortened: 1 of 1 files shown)


src/messageComposer/middleware/pollComposer/state.ts
```diff
@@ -35,7 +35,7 @@ export const pollStateChangeValidators: Partial<
     if (!numericMatch && value) {
       return { max_votes_allowed: 'Only numbers are allowed' };
     }
-    if (value?.length > 1 && !value.match(VALID_MAX_VOTES_VALUE_REGEX))
+    if (!value.match(VALID_MAX_VOTES_VALUE_REGEX))
```
MartinCupela, line 38: It should not be validation on change, but rather on blur. I think it is too early to validate when you start typing 10.

khushal87: Makes sense. We had this for `onChange` on react native, so I thought it would be a bug that way. But what you are saying makes sense. I will close it in that case. Thanks :)

MartinCupela: and on blur this validation happens

khushal87: Yes, it seems to happen on blur

PR #1532, opened 2025-05-12
fix: add new poll option only when all of the current options are filled

khushal87:
Currently the options are added when the last option has some text. But when the new option added is reordered, there is no way to add a new option. To improve this, the new behaviour is to add new option when all of the option fields are filled.

### DIFF
--- a/src/messageComposer/middleware/pollComposer/state.ts
+++ b/src/messageComposer/middleware/pollComposer/state.ts
@@ -128,6 +128,7 @@ export const pollCompositionStateProcessors: Partial<
       ...optionListTail,
     ];
 
+    // Add new option if all the options are filled
     if (!newOptions.some((option) => !option.text.trim())) {
       newOptions.push({ id: generateUUIDv4(), text: '' });
     }
--- a/src/messageComposer/middleware/pollComposer/state.ts
+++ b/src/messageComposer/middleware/pollComposer/state.ts
@@ -128,8 +128,11 @@ export const pollCompositionStateProcessors: Partial<
       ...optionListTail,
     ];
 
-    // Add new option if all the options are filled
-    if (!newOptions.some((option) => !option.text.trim())) {
+    const shouldAddNewOption =
+      prevOptions.length < MAX_POLL_OPTIONS &&
+      !newOptions.some((option) => !option.text.trim());
+
+    if (shouldAddNewOption) {
       newOptions.push({ id: generateUUIDv4(), text: '' });
     }
 
--- a/src/messageComposer/middleware/pollComposer/state.ts
+++ b/src/messageComposer/middleware/pollComposer/state.ts
@@ -116,23 +116,21 @@ export const pollCompositionStateProcessors: Partial<
     const { index, text } = value;
     const prevOptions = data.options || [];
 
-    const shouldAddEmptyOption =
-      prevOptions.length < MAX_POLL_OPTIONS &&
-      (!prevOptions || (prevOptions.slice(index + 1).length === 0 && !!text));
-
     const shouldRemoveOption =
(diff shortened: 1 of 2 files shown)


src/messageComposer/middleware/pollComposer/state.ts
```diff
@@ -116,24 +116,23 @@ export const pollCompositionStateProcessors: Partial<
...
+    const optionListTail = prevOptions.slice(index + 1);
 
     const newOptions = [
       ...optionListHead,
       ...(shouldRemoveOption ? [] : [{ ...prevOptions[index], text }]),
       ...optionListTail,
     ];
 
+    // Add new option if all the options are filled
+    if (!newOptions.some((option) => !option.text.trim())) {
```
MartinCupela, line 132: I would prefer to have a variable that says what the condition is checking. Like for example `shouldAddEmptyOption`. It helps understanding the code.

MartinCupela: Also, we should be checking whether we are not reaching `MAX_POLL_OPTIONS` limit - hard limit from BE (`prevOptions.length < MAX_POLL_OPTIONS`)

test/unit/MessageComposer/middleware/pollComposer/state.test.ts
```diff
@@ -182,24 +182,36 @@ describe('PollComposerStateMiddleware', () => {
       expect(result.status).toBeUndefined;
     });
 
-    it('should add a new empty option when the last option is filled', async () => {
+    it('should add a new empty option when the all the options are filled', async () => {
       const stateMiddleware = setup();
       const result = await stateMiddleware.handlers.handleFieldChange(
         setupHandlerParams({
           nextState: { ...getInitialState() },
```
MartinCupela, line 189: The next state has one empty option and the previous state has 3 options. That is inconsistent.

MartinCupela: Maybe you could keep this test as it was and add a new tests that:

1. simulates the scenario of moving an option to a different position
2. simulates the scenario of providing a value to a previously empty non-last option

src/messageComposer/middleware/pollComposer/state.ts
khushal87: I have pushed the suggestions. Please let me know what do you think. Thanks 😄

test/unit/MessageComposer/middleware/pollComposer/state.test.ts
khushal87: I have pushed the suggestions. Please let me know what you think. Thanks 😄

test/unit/MessageComposer/middleware/pollComposer/state.test.ts
```diff
@@ -199,7 +201,54 @@ describe('PollComposerStateMiddleware', () => {
 
       expect(result.state.nextState.data.options.length).toBe(2);
       expect(result.state.nextState.data.options[0].text).toBe('Option 1');
-      expect(result.state.nextState.data.options[1].text).toBe('');
+      expect(result.state.nextState.data.options[1].text).toEqual('');
+
+      expect(result.status).toBeUndefined;
+    });
+
+    it('should reorder options and add a new empty option when the all the options are filled', async () => {
```
MartinCupela, line 209:
```suggestion
    it('should reorder options and add a new empty option when all the options are filled', async () => {
```

MartinCupela (approved):

test/unit/MessageComposer/middleware/pollComposer/state.test.ts
khushal87: Added the typo fix

MartinCupela:
@khushal87 feel free to merge. I have approved

PR #1534, opened 2025-05-12
feat: add taking state snapshots to MessageComposer

MartinCupela:
## Goal

Allow to take state snapshots in order to be able to restore the state in the future. The logic of taking snapshot is encapsulated as we may want to change it in the future. The only important part of the contract is that the method returns an object with property `restore` that allows us to restore the state to the given snapshot. We do not care how the snapshot looks like.

### DIFF
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -33,6 +33,10 @@ export type EditingAuditState = {
   lastChange: LastComposerChange;
 };
 
+export type MessageComposerStateSnapshot = {
+  restore(): void;
+};
+
 export type LocalMessageWithLegacyThreadId = LocalMessage & { legacyThreadId?: string };
 export type CompositionContext = Channel | Thread | LocalMessageWithLegacyThreadId;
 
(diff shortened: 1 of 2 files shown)


src/messageComposer/messageComposer.ts
```diff
@@ -337,6 +341,26 @@ export class MessageComposer {
...
+
+    return {
+      restore: () => {
+        this.state.next(state);
+        this.textComposer.state.next(textComposerState);
+        this.attachmentManager.state.next(attachmentManagerState);
+        this.linkPreviewsManager.state.next(linkPreviewsManagerState);
+        this.pollComposer.state.next(pollComposerState);
+        this.customDataManager.state.next(customDataManagerState);
+      },
```
oliverlaz, line 360: What would happen if something modifies the state, but a transaction fails and it is rolled back?
Would we lose data in that case?

MartinCupela: What is you are concerned about?

MartinCupela: If you are concerned about state changes due to some effects we could add something line `messageComposer.lockState()` and that could be called before `messageComposer. takeStateSnapshot()`.

oliverlaz: Do we need this functionality in the state store? I'm not sure if introducing global locks is necessary.
Can we do this in the UI SDK only?

PR #1542, opened 2025-05-22
fix: improve attachment upload error notification data

MartinCupela:
## Goal

Closes REACT-399

Provide more contextual information regarding the reasons of upload failure or blockage.
Add code property to the Notification type to allow to identify notifications based on a type (kind of a grouping).


src/messageComposer/attachmentManager.ts
```diff
@@ -473,21 +482,7 @@ export class AttachmentManager {
...
-      } else if (typeof error === 'object') {
-        finalError = Object.assign(finalError, error);
-      }
-
-      this.client.notifications.addError({
-        message: finalError.message,
-        origin: { emitter: 'AttachmentManager', context: { attachment } },
-      });
-
+      const reason = (error as Error).message ?? 'unknown error';
```
arnautov-anton, line 485:
```suggestion
      const reason = error instanceOf Error ? error.message : 'unknown error';
```

src/messageComposer/attachmentManager.ts
```diff
@@ -453,8 +453,17 @@ export class AttachmentManager {
...
+        message: `The attachment upload was blocked`,
+        origin: {
+          emitter: 'AttachmentManager',
+          context: { attachment, blockedAttachment: localAttachment },
+        },
+        options: {
+          code: 'attachment.upload.blocked',
+          metadata: {
+            reason: localAttachment.localMetadata.uploadPermissionCheck?.reason,
+          },
```
arnautov-anton, line 465: Is it worth adding original error object at the same level the `reason` property is (if this is available in the UI)? `ErrorFromResponse` contains quite a bit of data which could be useful, like `code` and `status` numbers.

MartinCupela: But this not a error from a response. This is a reason calculated based on the back-end config.

src/messageComposer/attachmentManager.ts
arnautov-anton: Not inherently better, just "type-safer" - either of the options do not handle errors which come back as simple strings though. 

image

src/messageComposer/messageComposer.ts
```diff
@@ -668,12 +668,18 @@ export class MessageComposer extends WithSubscriptions {
...
+      this.client.notifications.addError({
         message: 'Failed to create the poll',
         origin: {
           emitter: 'MessageComposer',
           context: { composer: this },
         },
+        options: {
+          code: 'poll.creation.failed',
+          metadata: {
+            reason: (error as Error).message,
```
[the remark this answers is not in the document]
MartinCupela: I have added `originalError` to the notification object.

src/messageComposer/attachmentManager.ts
```diff
@@ -386,12 +386,11 @@ export class AttachmentManager {
       this.client.notifications.addError({
         message: 'File is required for upload attachment',
         origin: { emitter: 'AttachmentManager', context: { attachment } },
+        options: { code: 'validation:attachment:file:missing' },
```
isekovanic: *nitpick*

I'm assuming `options.code` is meant to be a unique identifier for the error, right ? Would it be possible to rename it to something like `identifier` or similar ?

`code` reminds me of an `error.code`, which this is not.

Also, it might be useful to extract the potential codes, (or identifiers) into a separate object that would be something like `Record<Symbol, string>` and then we tell integrators to use those as a reference point (that way if we decide to change some of these or extend them, we're not breaking anyone as they would still consume them from the symbolicated enum).

Nothing blocking though.

MartinCupela: What if we changed the name to `code` to `type`? Name `identifier` could lead to confusion too as each notification has `id`. On the other hand, the purpose of `code` is grouping - thus `type` could serve better here.

isekovanic: Yep, that also sounds good to me !

PR #1544, opened 2025-05-23
chore: fix MessageComposer.registerSubscriptions

arnautov-anton:

Newly introduced helper `WithSubscriptions` abstract class provides a pre-defined `unregisterSubscriptions` method which needs to be in this form (not arrow function) for devs to be able to use it with `super` keyword. This obviously means that `this` has to be bound to it when used as standalone function as `this` means something else entirely when it's not attached to an object (or attached to a different object). Rookie mistake, my apologies. Thank you @khushal87 for spotting this issue.

Marked as `chore` as to not pollute changelog with unreleased broken stuff.

### DIFF
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -372,7 +372,7 @@ export class MessageComposer extends WithSubscriptions {
     this.addUnsubscribeFunction(this.subscribeMessageComposerStateChanged());
     this.addUnsubscribeFunction(this.subscribeMessageComposerConfigStateChanged());
 
-    return this.unregisterSubscriptions;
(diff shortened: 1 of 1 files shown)


MartinCupela (approved):

khushal87 (approved):

isekovanic (approved):
Nice ! I suppose it's not needed in the other places as we don't return the subscriptions anywhere ?

arnautov-anton:
> Nice ! I suppose it's not needed in the other places as we don't return the subscriptions anywhere ?

Exactly, this is the only place where we return it. I briefly considered it in other places but that's a thing for another day. :)

PR #1545, opened 2025-05-23
fix: respect attachment upload config for files of type FileReference

khushal87:
The file upload config was not checked for `FileReference` type files, due to which it did not work on React Native. This is fixed in this PR.

### DIFF
--- a/test/unit/MessageComposer/attachmentManager.test.ts
+++ b/test/unit/MessageComposer/attachmentManager.test.ts
@@ -14,6 +14,43 @@ import {
 } from '../../../src';
 import { AppSettings, AttachmentManager } from '../../../src';
 
+/**
+ * Utility to generate a  file
+ */
+const generateFile = ({
+  name,
+  size = 0,
+  type,
+}: {
+  name: string;
+  size?: number;
+  type: string;
+}) => {
+  return new File([new ArrayBuffer(size)], name, { type });
+};
+
+/**
+ * Utility to generate a FileReference
+ */
+const generateFileReference = ({
+  name,
+  size = 0,
+  type,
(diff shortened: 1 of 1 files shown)


test/unit/MessageComposer/attachmentManager.test.ts
```diff
@@ -575,6 +576,35 @@ describe('AttachmentManager', () => {
   });
 
   describe('getUploadConfigCheck', () => {
+    it('should not check for file extension config if its a Blob type', async () => {
+      const blob = new Blob([''], { type: 'image/gif' });
+      const {
+        messageComposer: { attachmentManager },
+      } = setup();
+      const result = await attachmentManager.getUploadConfigCheck(blob);
+      expect(result).not.toEqual({
```
MartinCupela, line 585: reason should not be present if upload is not blocked. This assertion probably does not make sense

test/unit/MessageComposer/attachmentManager.test.ts
```diff
@@ -575,6 +576,35 @@ describe('AttachmentManager', () => {
...
+        uploadBlocked: false,
+        reason: 'allowed_file_extensions',
+      });
+    });
+
+    it('should check for file extension config if its a FileReference type', async () => {
```
MartinCupela, line 591: We should be probably checking the the same results we get for File and FileReference.

test/unit/MessageComposer/attachmentManager.test.ts
khushal87: You are right. I have removed it

MartinCupela (approved):

PR #1548, opened 2025-05-28
feat: handle command injection optionally

khushal87:
The goal of the PR is to handle command injection and apply the command settings to the text using middleware that the SDK or apps can plug in. The existing commands middleware will be responsible for setting the `command` state to the `textComposer`, which can be used by the other middleware.

SDKs can basically add something like this to support the commands UI on message input:
```
 messageComposer.compositionMiddlewareExecutor.insert({
    middleware: [createCommandInjectionMiddleware(messageComposer)],
    position: { after: 'stream-io/message-composer-middleware/attachments' },
  });

  messageComposer.draftCompositionMiddlewareExecutor.insert({
    middleware: [createDraftCommandInjectionMiddleware(messageComposer)],
    position: { after: 'stream-io/message-composer-middleware/draft-attachments' },
  });

  messageComposer.textComposer.middlewareExecutor.insert({
    middleware: [createActiveCommandGuardMiddleware() as TextComposerMiddleware],
    position: { before: 'stream-io/text-composer/commands-middleware' },
  });

  messageComposer.textComposer.middlewareExecutor.insert({
    middleware: [createCommandStringExtractionMiddleware() as TextComposerMiddleware],
    position: { after: 'stream-io/text-composer/commands-middleware' },
  });
```

### DIFF
--- a/src/messageComposer/middleware/textComposer/commandStringExtraction.ts
+++ b/src/messageComposer/middleware/textComposer/commandStringExtraction.ts
@@ -1,26 +1,27 @@
-import type { Middleware } from '../../../middleware';
 import type { TextComposerMiddlewareExecutorState } from './TextComposerMiddlewareExecutor';
-import { stripTextFromStartOfTheText } from './textMiddlewareUtils';
 import type { CommandSuggestion } from './types';
+import type { Middleware } from '../../../middleware';
+import { escapeRegExp } from './textMiddlewareUtils';
 
-export type ApplyCommandSettingsMiddleware = Middleware<
+export type CommandStringExtractionMiddleware = Middleware<
   TextComposerMiddlewareExecutorState<CommandSuggestion>,
   'onChange' | 'onSuggestionItemSelect'
 >;
 
+const stripCommandFromText = (text: string, commandName: string) =>
+  text.replace(new RegExp(`^${escapeRegExp(`/${commandName}`)}\\s*`), '');
+
 export const createCommandStringExtractionMiddleware =
-  (): ApplyCommandSettingsMiddleware => ({
+  (): CommandStringExtractionMiddleware => ({
     handlers: {
       onChange: ({ complete, forward, state }) => {
         const { command } = state;
 
-        if (!command) {
+        if (!command?.name) {
           return forward();
         }
 
-        const triggerWithCommand = `/${command.name} `;
-
-        const newText = stripTextFromStartOfTheText(triggerWithCommand, state.text);
+        const newText = stripCommandFromText(state.text, command.name);
 
         return complete({
           ...state,
--- a/src/messageComposer/middleware/textComposer/commands.ts
+++ b/src/messageComposer/middleware/textComposer/commands.ts
@@ -6,10 +6,9 @@ import type { CommandResponse } from '../../../types';
 import { mergeWith } from '../../../utils/mergeWith';
 import type { CommandSuggestion, TextComposerMiddlewareOptions } from './types';
 import {
-  getFirstWordFromText,
+  getCompleteCommandInString,
   getTriggerCharWithToken,
   insertItemWithTrigger,
-  startsWithTextAndSpace,
 } from './textMiddlewareUtils';
 import type { TextComposerMiddlewareExecutorState } from './TextComposerMiddlewareExecutor';
 
@@ -121,6 +120,17 @@ export const createCommandsMiddleware = (
       onChange: ({ state, next, complete, forward }) => {
         if (!state.selection) return forward();
         const finalText = state.text.slice(0, state.selection.end);
+        const commandName = getCompleteCommandInString(finalText);
+        if (commandName) {
+          const command = searchSource?.query(commandName).items[0];
+          if (command) {
+            return next({
+              ...state,
+              command,
+              suggestions: undefined,
+            });
+          }
+        }
 
         const triggerWithToken = getTriggerCharWithToken({
           trigger: finalOptions.trigger,
@@ -148,30 +158,11 @@ export const createCommandsMiddleware = (
           return next(newState);
         }
 
-        const query = triggerWithToken.slice(1);
-
-        const searchQuery = getFirstWordFromText(query);
-
-        const commands = searchSource?.query(searchQuery).items;
-
-        const [command] = commands;
-
-        const matchedCommand =
-          command && startsWithTextAndSpace(query, command.name) ? command : null;
-
-        if (matchedCommand) {
-          return next({
-            ...state,
-            command: matchedCommand,
-            suggestions: undefined,
-          });
-        }
-
         return complete({
           ...state,
           command: null,
           suggestions: {
-            query,
+            query: triggerWithToken.slice(1),
             searchSource,
             trigger: finalOptions.trigger,
           },
--- a/src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
+++ b/src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
@@ -16,14 +16,12 @@ export const getTriggerCharWithToken = ({
   isCommand?: boolean;
   acceptTrailingSpaces?: boolean;
 }) => {
-  // No trigger in between text
-  const notTrigger = `[^${trigger}]*`;
   const triggerNorWhitespace = `[^\\s${trigger}]*`;
 
   const match = text.match(
     new RegExp(
       isCommand
-        ? `^[${trigger}]${notTrigger}$`
+        ? `^[${trigger}]${triggerNorWhitespace}$`
         : acceptTrailingSpaces
           ? `(?!^|\\W)?[${trigger}]${triggerNorWhitespace}\\s?${triggerNorWhitespace}$`
           : `(?!^|\\W)?[${trigger}]${triggerNorWhitespace}$`,
@@ -31,8 +29,15 @@ export const getTriggerCharWithToken = ({
     ),
   );
 
-  const result = match && match[match.length - 1];
-  return isCommand ? result : result?.trim();
+  return match && match[match.length - 1].trim();
+};
+
+export const getCompleteCommandInString = (text: string) => {
+  // starts with "/" followed by 1+ non-whitespace chars followed by 1+ white-space chars
+  // the comand name is extracted into a separate group
+  const match = text.match(/^\/(\S+)\s+.*/);
+  const commandName = match && match[1];
+  return commandName;
 };
 
 export const insertItemWithTrigger = ({
@@ -143,44 +148,3 @@ export const getTokenizedSuggestionDisplayName = ({
       : [displayName],
   },
 });
-
-/**
- * Utility function to check if a given input text matches another text.
- * @param input
- * @param textToBeMatchedWith
- * @returns
- */
-export const startsWithTextAndSpace = (
-  input: string,
-  textToBeMatchedWith: string,
-): boolean => {
-  try {
-    const regex = new RegExp(
-      `^${escapeRegExp(textToBeMatchedWith.toLowerCase())}\\s+`,
-      'i',
-    );
-    return regex.test(input.toLowerCase());
-  } catch (error) {
-    console.error('Error in validating with the regex:', error);
-    return false;
-  }
-};
-
-/**
- * Extracts the first word from a given text.
- * @param text - The input text from which to extract the first word.
- * @returns The first word found in the text, or an empty string if no word is found.
- */
-export const getFirstWordFromText = (text: string): string => {
-  const match = text.match(/^\s*(\S+)/);
-  return match ? match[1] : '';
-};
-
-/**
- * Strips the textToStrip from the start of the text.
- * @param textToStrip - The string to be removed from the start of the text.
- * @param text - The input text from which to strip the textToStrip.
- * @returns The text with the textToStrip removed from the start.
- */
-export const stripTextFromStartOfTheText = (textToStrip: string, text: string) =>
-  text.replace(new RegExp(`^${escapeRegExp(textToStrip)}\\s*`), '');
--- a/src/messageComposer/textComposer.ts
+++ b/src/messageComposer/textComposer.ts
@@ -195,7 +195,13 @@ export class TextComposer {
     this.state.partialNext({ selection });
   };
 
-  insertText = ({ text, selection }: { text: string; selection?: TextSelection }) => {
+  insertText = async ({
+    text,
+    selection,
+  }: {
+    text: string;
+    selection?: TextSelection;
+  }) => {
     if (!this.enabled) return;
 
     const finalSelection: TextSelection = selection ?? this.selection;
@@ -217,7 +223,7 @@ export class TextComposer {
         ? finalText.length
         : currentText.slice(0, expectedCursorPosition).length;
 
-    this.state.partialNext({
+    await this.handleChange({
       text: finalText,
       selection: {
         start: cursorPosition,
--- a/src/pagination/BasePaginator.ts
+++ b/src/pagination/BasePaginator.ts
@@ -38,7 +38,7 @@ export const DEFAULT_PAGINATION_OPTIONS: Required<PaginatorOptions> = {
 
 export abstract class BasePaginator<T> {
   state: StateStore<PaginatorState<T>>;
-  protected pageSize: number;
+  pageSize: number;
   protected _executeQueryDebounced!: DebouncedExecQueryFunction;
   protected _isCursorPagination = false;
 
--- a/src/pagination/ReminderPaginator.ts
+++ b/src/pagination/ReminderPaginator.ts
@@ -9,8 +9,26 @@ import type { StreamChat } from '../client';
 
 export class ReminderPaginator extends BasePaginator<ReminderResponse> {
   private client: StreamChat;
-  filters: ReminderFilters | undefined;
-  sort: ReminderSort | undefined;
+  protected _filters: ReminderFilters | undefined;
+  protected _sort: ReminderSort | undefined;
+
+  get filters(): ReminderFilters | undefined {
+    return this._filters;
+  }
+
+  get sort(): ReminderSort | undefined {
+    return this._sort;
+  }
+
+  set filters(filters: ReminderFilters | undefined) {
+    this._filters = filters;
+    this.resetState();
+  }
+
+  set sort(sort: ReminderSort | undefined) {
+    this._sort = sort;
+    this.resetState();
+  }
 
   constructor(client: StreamChat, options?: PaginatorOptions) {
     super(options);
--- a/src/reminders/ReminderManager.ts
+++ b/src/reminders/ReminderManager.ts
@@ -1,4 +1,5 @@
 import { Reminder } from './Reminder';
+import { DEFAULT_STOP_REFRESH_BOUNDARY_MS } from './ReminderTimer';
 import { StateStore } from '../store';
 import { ReminderPaginator } from '../pagination';
 import { WithSubscriptions } from '../utils/WithSubscriptions';
@@ -26,6 +27,7 @@ export const DEFAULT_REMINDER_MANAGER_CONFIG: ReminderManagerConfig = {
     8 * oneHour,
     oneDay,
   ],
+  stopTimerRefreshBoundaryMs: DEFAULT_STOP_REFRESH_BOUNDARY_MS,
 };
 
 const isReminderExistsError = (error: Error) =>
@@ -51,12 +53,12 @@ export type ReminderManagerState = {
 
 export type ReminderManagerConfig = {
   scheduledOffsetsMs: number[];
-  stopTimerRefreshBoundaryMs?: number;
+  stopTimerRefreshBoundaryMs: number;
 };
 
 export type ReminderManagerOptions = {
   client: StreamChat;
-  config?: ReminderManagerConfig;
+  config?: Partial<ReminderManagerConfig>;
 };
 
 export class ReminderManager extends WithSubscriptions {
@@ -71,6 +73,9 @@ export class ReminderManager extends WithSubscriptions {
     this.configState = new StateStore({
       scheduledOffsetsMs:
         config?.scheduledOffsetsMs ?? DEFAULT_REMINDER_MANAGER_CONFIG.scheduledOffsetsMs,
+      stopTimerRefreshBoundaryMs:
+        config?.stopTimerRefreshBoundaryMs ??
+        DEFAULT_REMINDER_MANAGER_CONFIG.stopTimerRefreshBoundaryMs,
     });
     this.state = new StateStore({ reminders: new Map<MessageId, Reminder>() });
     this.paginator = new ReminderPaginator(client);
@@ -78,6 +83,15 @@ export class ReminderManager extends WithSubscriptions {
 
   // Config API START //
   updateConfig(config: Partial<ReminderManagerConfig>) {
+    if (
+      typeof config.stopTimerRefreshBoundaryMs === 'number' &&
+      config.stopTimerRefreshBoundaryMs !== this.stopTimerRefreshBoundaryMs
+    ) {
+      this.reminders.forEach((reminder) => {
+        reminder.timer.stopRefreshBoundaryMs =
+          config?.stopTimerRefreshBoundaryMs as number;
+      });
+    }
     this.configState.partialNext(config);
   }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -4036,7 +4036,7 @@ export type ReminderAPIResponse = APIResponse & {
 
 export type CreateReminderOptions = {
   messageId: string;
-  remind_at?: string;
+  remind_at?: string | null;
   user_id?: string;
 };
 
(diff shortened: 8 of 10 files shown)


src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -143,29 +148,49 @@ export const createCommandsMiddleware = (
           return next(newState);
         }
 
+        const inputText = triggerWithToken?.toLowerCase().slice(1);
+        const matchedCommand = commands.find((command) => {
```
MartinCupela: why are you not using the search source?

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -143,29 +148,49 @@ export const createCommandsMiddleware = (
...
+          if (!command.name || !inputText) return false;
+          return isTextMatched(inputText, command.name.toLowerCase());
```
MartinCupela: Isn't the command already identified with `triggerWithToken`?

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -143,29 +148,49 @@ export const createCommandsMiddleware = (
...
+        });
+
+        if (matchedCommand) {
```
MartinCupela: Is this condition necessary? The command can be passed to the complete below it.

MartinCupela: Resp. we could remove the complete call and just use next

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
```diff
@@ -134,3 +134,13 @@ export const getTokenizedSuggestionDisplayName = ({
       : [displayName],
   },
 });
+
+export const isTextMatched = (input: string, command: string): boolean => {
```
MartinCupela: Isn't this a duplicate logic already present in another util function in this file?

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
```diff
@@ -0,0 +1,55 @@
...
+        const { command } = state;
+
+        if (!command) {
+          return forward();
+        }
+
+        const trigger = `/${command?.name} `;
+
+        const newText = state.text.slice(trigger.length);
+        return complete({
```
MartinCupela: I am thinking whether if commands middleware is now doing next, whether this middleware should do complete

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
```diff
@@ -0,0 +1,55 @@
...
+    handlers: {
+      onChange: ({ complete, forward, state }) => {
+        const { command } = state;
+
+        if (!command) {
+          return forward();
+        }
+
+        const trigger = `/${command.name}`;
+        const newText = state.text.replace(new RegExp(`^${trigger}(\\s|$)`), '');
```
MartinCupela: this does not trim white space following the command.

MartinCupela: `'abc   efg'.replace(new RegExp(`^${'abc'}(\\s|$)`), '')` -> `'   efg'`

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
```diff
@@ -0,0 +1,55 @@
...
+  (): ApplyCommandSettingsMiddleware => ({
+    handlers: {
+      onChange: ({ complete, forward, state }) => {
+        const { command } = state;
+
+        if (!command) {
+          return forward();
+        }
+
+        const trigger = `/${command.name}`;
```
MartinCupela: trigger is the symbol `/`. I would choose another name, e.g. `triggerWithCommand`

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
```diff
@@ -0,0 +1,55 @@
...
+        const trigger = `/${command?.name} `;
+
+        const newText = state.text.slice(trigger.length);
+        return complete({
+          ...state,
+          selection: {
+            end: state.selection.end - trigger.length,
+            start: state.selection.start - trigger.length,
+          },
+          suggestions: undefined,
```
MartinCupela: Why are the suggestions set to undefined here? I would expect this middleware to take care only of the text.

src/messageComposer/middleware/messageComposer/commandInjection.ts
```diff
@@ -0,0 +1,72 @@
...
+          text: enrichedText,
+        },
+        message: {
+          ...state.message,
+          text: enrichedText,
+        },
+      });
+    },
+  },
+  id: 'stream-io/message-composer-middleware/command-injection',
```
MartinCupela:
```suggestion
  id: 'stream-io/message-composer-middleware/command-string-injection',
```

src/messageComposer/middleware/messageComposer/commandInjection.ts
```diff
@@ -0,0 +1,72 @@
...
+        return forward();
+      }
+      const { text } = state.localMessage;
+
+      const injection = `/${command?.name}`;
+      const enrichedText = `${injection} ${text}`;
+
+      return complete({
+        ...state,
+        localMessage: {
```
MartinCupela, line 30: Where is the command value nullified?

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: In the condition inside the `matchedCommand` in the commands middleware it does `next` that is why the `complete` was added here.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: I will check if this can be used.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: Yes, 100%, but it just gives me text starting with `/`, let's say `/h`, `/he`, `/hel` or null if the text isn't starting with `/`. But I need to set the command to a known command in the config for which I will have to do an exact match. Let me know if you think if there's a better way to know that.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: I used `next` here so that the `applyCommands` middleware will `complete` the chain if the command is matched. Else, we don't want to set anything to the command state.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: It was done because the suggestions were still defined post the trigger matched and the list was showed up but I think this can be handled on the commands middleware if the command got matched. Nice catch

src/messageComposer/middleware/messageComposer/commandInjection.ts
khushal87: It is done in the initState, but do you think we should clear it here as well?

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
khushal87: Which utility are you talking about here?

src/messageComposer/middleware/textComposer/commands.ts
khushal87: Question:

Is it fine to make the middleware async and get commands from `const commands = await searchSource.query(inputText);`?

src/messageComposer/middleware/messageComposer/commandInjection.ts
MartinCupela: I think we could unset the command here if we are returning it back to the text here.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
MartinCupela: I think that command middleware for handle selection already sets suggestions to undefined

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
MartinCupela: I meant that if the command middleware allows the chain to continue, why should this middleware stop it. If there are possible more than 1 command related middleware handlers, then we open the possibility to more than 2 command related handlers.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: Yep. I have removed it for `onChange` as well and handled that part on the `commands` middleware itself for matched command

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: If you test it, for example, even a simple `/ban` would trim the entire text and limit it to ''. I am not sure if whitespaces checks are needed here. But I am open to modify if you think this is incorrect. PFA a video here.

https://github.com/user-attachments/assets/a0ac6a42-fab5-47f7-b61d-609929f8801e

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: I have absolutely nothing against having `next` here if we wanna give that flexibility to our integrators.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
```diff
@@ -0,0 +1,55 @@
+import type { Middleware } from '../../../middleware';
+import type { TextComposerMiddlewareExecutorState } from './TextComposerMiddlewareExecutor';
+import type { CommandSuggestion } from './types';
+
+export type ApplyCommandSettingsMiddleware = Middleware<
+  TextComposerMiddlewareExecutorState<CommandSuggestion>,
+  'onChange' | 'onSuggestionItemSelect'
+>;
+
+export const createApplyCommandSettingsMiddleware =
```
MartinCupela: I would probably be looking for another name. It is not clear to me, what command settings mean in the name `createApplyCommandSettingsMiddleware`. The middleware is just removing string that corresponds to the trigger and command name.

MartinCupela:
@khushal87 the changes to command middleware you introduced lead to removal of suggestions prematurely. See the video with the app running on this branch:

[no_suggestions.webm](https://github.com/user-attachments/assets/ea535e70-e4db-4bf4-a0a8-63cc2c7c8925)

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
MartinCupela: If you copy-past "/ban   ", then the whitespaces are not trimmed. I think we should remove them here.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
MartinCupela: I can still see it : )

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: Let me check

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: Ah, I missed pushing it.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: Added `next`

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: Pushed now

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: In the `getTriggerCharWithToken`, it seems we are not allowing that anyway. I get a `null` when I paste `/ban  ` out as output. If we want to support it, we will have to use state.text as the input. Will that be okay?

khushal87:
Hey @MartinCupela, I have refined the implementation a bit to rely on the `state.text` instead of `triggerWithToken` as the `triggerWithToken` is pretty much limited to only a pattern of `/${command}` and it doesn't allow handling whitespace or so.  Relying on `state.text` makes it easier to handle the whitespace and also ensures that the `triggerWithToken` is untouched as the purpose of it is to only searching the command.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: This one will now work.

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: How about `removeCommandTriggerFromTextMiddleware`?

src/messageComposer/middleware/messageComposer/commandInjection.ts
```diff
@@ -0,0 +1,76 @@
...
+      const command = composer.textComposer.command;
+      if (!command) {
+        return forward();
+      }
+      const { text } = state.localMessage;
+
+      const injection = `/${command?.name}`;
+      const enrichedText = `${injection} ${text}`;
+
+      composer.textComposer.setCommand(null);
```
MartinCupela: The command should be nullified by setting the null value in the `return complete()` call. The whole middleware thing is about preparing a state value that is then committed to the state after the chain successfully finishes

src/messageComposer/middleware/messageComposer/commandInjection.ts
```diff
@@ -0,0 +1,76 @@
...
+      const command = composer.textComposer.command;
+      if (!command) {
+        return forward();
+      }
+      const { text } = state.draft;
+
+      const injection = `/${command?.name}`;
+      const enrichedText = `${injection} ${text}`;
+
+      composer.textComposer.setCommand(null);
```
MartinCupela: > The command should be nullified by setting the null value in the return complete() call. The whole middleware thing is about preparing a state value that is then committed to the state after the chain successfully finishes

Also applies to this

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
MartinCupela: If we have command `createCommandInjectionMiddleware` then here I would go with `createCommandExtractionMiddleware` or `createCommandStringExtractionMiddleware`

src/messageComposer/middleware/messageComposer/commandInjection.ts
MartinCupela: My mistake I did not realize that this is inside the message composition flow.

khushal87: Resolving it in that case

MartinCupela: As agreed, we should rely on state being cleared after the message is sent. There it is already nullified.

src/messageComposer/middleware/messageComposer/commandInjection.ts
khushal87: Discussed as above

src/messageComposer/middleware/textComposer/applyCommandSettings.ts
khushal87: I will go with `createCommandStringExtractionMiddleware` in that case

khushal87: added new name

MartinCupela:
I think this PR should also solve the scenario, when a user wants to cancel the command immediately after its selection - for example by pressing backspace - when there is no change event, but rather keyDown event.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: Using it now.

src/messageComposer/middleware/textComposer/commandStringExtraction.ts
```diff
@@ -0,0 +1,55 @@
...
+          ...state,
+          selection: {
+            end: state.selection.end - triggerWithCommand.length,
+            start: state.selection.start - triggerWithCommand.length,
+          },
+          text: newText,
+        });
+      },
+    },
+    id: 'stream-io/text-composer/apply-command-settings',
```
MartinCupela:
```suggestion
    id: 'stream-io/text-composer/command-string-extraction',
```

src/messageComposer/middleware/textComposer/commandStringExtraction.ts
```diff
@@ -0,0 +1,55 @@
...
+  'onChange' | 'onSuggestionItemSelect'
+>;
+
+export const createCommandStringExtractionMiddleware =
+  (): ApplyCommandSettingsMiddleware => ({
+    handlers: {
+      onChange: ({ complete, forward, state }) => {
+        const { command } = state;
+
+        if (!command) {
```
MartinCupela: We should also forward if the command in the composer state equals this command.  That way we can also avoid setting incorrect selection:

Here the text is the actual argument to the command (for example `@`), so we cannot subtract the command length from a text that has 1 char for example.
```
        end: state.selection.end - triggerWithCommand.length,
         start: state.selection.start - triggerWithCommand.length,
```

khushal87: About the 2nd thing that you mentioned. The `state.selection` values are {end: 7, start: 7} when we have a command matched as `/unban`, the `triggerWithCommand` length will also be 7. And now, since we trimmed the command out of the text, the selection was subtracted. When you type on `@` the new calculation on the trimmed text with follow so we should be gtg.

There was a bug with the `triggerWithCommand` not accounting a space for the `onChange` that i have fixed in the recent commit.

simulator_screenshot_4CED68B4-0F26-492E-8143-2E5E2BFA757C

khushal87: > We should also forward if the command in the composer state equals this command. That way we can also avoid setting incorrect selection:

The command here in this middleware is already the one from the state so I am not sure what do you mean here. Can you explain more?

khushal87: Update for the 2nd suggestion: I have changed it to `newText.length` cause I realised that for inputs like `/ban    user3`, if pasted, it would be a problem, so it should equate the length of the new string after the extraction. Now it should be gtg.

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -142,29 +148,49 @@ export const createCommandsMiddleware = (
           return next(newState);
         }
 
+        const query = triggerWithToken.slice(1);
+
+        const searchQuery = getFirstWordFromText(query);
+
+        const commands = searchSource?.query(searchQuery).items;
+
+        const matchedCommand = commands?.find((command) =>
```
MartinCupela: @khushal87 does it make sense to perform query(searchQuery) and then again commands.find()? Isn't it doing the same with the difference that commands.find in this case would just be commands[0] (as find returns on the first matched and the first matched is the at 0 index before query returns all the matches)?

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -142,29 +148,49 @@ export const createCommandsMiddleware = (
           return next(newState);
         }
 
+        const query = triggerWithToken.slice(1);
+
+        const searchQuery = getFirstWordFromText(query);
```
MartinCupela: Why is it necessary to get first word from query if the query itself is just a trigger followed by chars that are not white space?

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
```diff
@@ -11,19 +16,23 @@ export const getTriggerCharWithToken = ({
...
 }) => {
+  // No trigger in between text
+  const notTrigger = `[^${trigger}]*`;
   const triggerNorWhitespace = `[^\\s${trigger}]*`;
+
   const match = text.match(
     new RegExp(
       isCommand
-        ? `^[${trigger}]${triggerNorWhitespace}$`
+        ? `^[${trigger}]${notTrigger}$`
```
MartinCupela: why do we allow white space?

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
```diff
@@ -134,3 +143,44 @@ export const getTokenizedSuggestionDisplayName = ({
...
   },
 });
+
+/**
+ * Utility function to check if a given input text matches another text.
+ * @param input
+ * @param textToBeMatchedWith
+ * @returns
+ */
+export const startsWithTextAndSpace = (
```
MartinCupela: I think we may not need this function. It is used in a single place and it actually removes white space that was kept previously due to allowing white space in getTriggerWithToken

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
```diff
@@ -134,3 +143,44 @@ export const getTokenizedSuggestionDisplayName = ({
...
+    return false;
+  }
+};
+
+/**
+ * Extracts the first word from a given text.
+ * @param text - The input text from which to extract the first word.
+ * @returns The first word found in the text, or an empty string if no word is found.
+ */
+export const getFirstWordFromText = (text: string): string => {
```
MartinCupela: Also this function exists only because we introduced this change:
https://github.com/GetStream/stream-chat-js/pull/1548/files#diff-3f1a5dd2e123452ae2ff28a98798ca79d3cf5fde3c8f1c82d3e406c95d7967e5R26

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
```diff
@@ -134,3 +143,44 @@ export const getTokenizedSuggestionDisplayName = ({
...
+  return match ? match[1] : '';
+};
+
+/**
+ * Strips the trigger from the text.
+ * @param text - The input text from which to strip the trigger.
+ * @param trigger - The trigger string to be removed from the start of the text.
+ * @returns The text with the trigger removed from the start.
+ */
+export const stripTriggerFromText = (text: string, trigger: string) =>
```
MartinCupela: This does not strip trigger. In the single place where it is used it strips a trigger with command name. Trigger is for example `"/"` but not `"/ban"`.
The function itself could also be renamed to `stripTextFromText`

src/messageComposer/textComposer.ts
```diff
@@ -169,6 +178,11 @@ export class TextComposer {
     this.state.partialNext({ mentionedUsers });
   };
 
+  setCommand = (command: CommandResponse | null) => {
+    if (command === this.command) return;
```
MartinCupela: This may not work if the objects have not the same memory reference, but their content refer to the same command. Maybe something like `command.name === this.command.name` would be more appropriate or to have a function that evaluates that two command objects are equal.

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
khushal87: How about `stripTextFromStartOfTheText` as it strips the text from the start of the text?

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
khushal87: Well, this function just checks if the `query` starts with the command name or not, and we are obviously allowing at least one or more spaces so that only then it would return true for a match, i.e. `/unban ` will be a matched command and `/unban` would not for the onChange.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: I will simplify it. Makes sense

khushal87: Done

src/messageComposer/middleware/textComposer/commands.ts
khushal87: The problem is the triggerWithToken returned null for any text except `/command`, i.e, it would also return null for text having a space and then we cannot search anything as it early exits.

src/messageComposer/middleware/textComposer/textMiddlewareUtils.ts
khushal87: The problem is the function returns null for any text except the pattern `/command` which, we cannot rely on if we want to use it as an input for search later in the middleware. That is why I allowed it.

src/messageComposer/middleware/textComposer/commands.ts
khushal87: Now, if you allow spaces, the user can paste anything let's say `/giphy good morning,` in that case,e the `query` will be `giphy good morning` but the search source cannot return anything as the pattern didn't match. That is why we get the first word from the query just for the command search.

src/messageComposer/textComposer.ts
khushal87: Fixed using the 1st way

PR #1553, opened 2025-06-03
fix: attach original token provider error as cause to loadToken rejection

myandrienko:

This is a backport of https://github.com/GetStream/stream-video-js/pull/1812 - see there for details.

## Changelog

- Errors (rejections) thrown from `client.connectUser()` will have `cause` attached in some cases with an original error that caused token to fail to load.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,7 @@
     "emitDeclarationOnly": true,
     "outDir": "./dist/types",
 
-    "lib": ["ES2020", "DOM"],
+    "lib": ["ES2020", "DOM", "ES2022.Error"],
     "moduleResolution": "bundler",
     "module": "Preserve",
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -22,7 +22,7 @@
     "emitDeclarationOnly": true,
     "outDir": "./dist/types",
 
-    "lib": ["ES2020", "DOM"],
+    "lib": ["ES2020", "DOM", "ES2022.Error"],
```
myandrienko, line 25: This is not a big deal, since even if the browser doesn't support `Error.prototype.cause`, the additional argument will just be ignored by the Error constructor.

It's not one of those things that require polyfills etc.

arnautov-anton (approved):

PR #1558, opened 2025-06-09
feat: introduce sync search

khushal87:
The goal of the PR is to introduce sync search by extending the current API. The previously `SearchSource` is now changed to `ISearchSource`, and the `SearchSource` is the async variant of `ISearchSource`, and the `SearchSourceSync` is the sync variant of the `ISearchSource`. 

The previously named `BaseSearchSource` is still intact and handles the async search, and the `BaseSearchSourceSync` is introduced to provide sync search capabilities. Both extend the `BaseSearchSourceBase` that handles the common utilities required for search.

This change was made to ensure that the classes that previously extended the `BaseSearchSource` do not complain about the new abstract methods if they were introduced, because you will never have the same search source having async and sync variants in the same class. Moreover, this change is done in a non-breaking way.

NOTE: The names used right now are a bit lame, and I am open to changing them to a suitable convention. Any suggestions would be highly appreciated.

### DIFF
--- a/src/messageComposer/middleware/textComposer/commands.ts
+++ b/src/messageComposer/middleware/textComposer/commands.ts
@@ -1,14 +1,14 @@
 import type { Channel } from '../../../channel';
 import type { Middleware } from '../../../middleware';
 import type { SearchSourceOptions } from '../../../search';
-import { BaseSearchSource } from '../../../search';
+import { BaseSearchSourceSync } from '../../../search';
 import type { CommandResponse } from '../../../types';
 import { mergeWith } from '../../../utils/mergeWith';
 import type { CommandSuggestion, TextComposerMiddlewareOptions } from './types';
 import { getTriggerCharWithToken, insertItemWithTrigger } from './textMiddlewareUtils';
 import type { TextComposerMiddlewareExecutorState } from './TextComposerMiddlewareExecutor';
 
-export class CommandSearchSource extends BaseSearchSource<CommandSuggestion> {
+export class CommandSearchSource extends BaseSearchSourceSync<CommandSuggestion> {
   readonly type = 'commands';
   private channel: Channel;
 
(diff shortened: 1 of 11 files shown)


src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -31,7 +31,7 @@ export class CommandSearchSource extends BaseSearchSource<CommandSuggestion> {
     };
   }
 
-  query(searchQuery: string) {
+  searchCommands = (searchQuery: string) => {
```
MartinCupela: Why do we need searchCommands method? Could we just keep `query` method?

src/search/BaseSearchSource.ts
```diff
@@ -30,48 +32,40 @@ export interface SearchSource<T = any> {
...
-  offset?: number;
-};
-export type SearchSourceOptions = {
-  /** The number of milliseconds to debounce the search query. The default interval is 300ms. */
-  debounceMs?: number;
-  pageSize?: number;
-};
+export interface SearchSource<T = any> extends ISearchSource<T> {
+  cancelScheduledQuery(): void;
+  setDebounceOptions(options: DebounceOptions): void;
```
MartinCupela, line 44: Could `cancelScheduledQuery` and `setDebounceOptions` be part of `ISearchSource` if they have the same signature for both async and sync class?

src/search/BaseSearchSource.ts
```diff
@@ -30,48 +32,40 @@ export interface SearchSource<T = any> {
...
   protected pageSize: number;
   abstract readonly type: SearchSourceType;
-  protected searchDebounced!: DebouncedExecQueryFunction;
 
   protected constructor(options?: SearchSourceOptions) {
-    const { debounceMs, pageSize } = { ...DEFAULT_SEARCH_SOURCE_OPTIONS, ...options };
+    const { pageSize } = { ...DEFAULT_SEARCH_SOURCE_OPTIONS, ...options };
     this.pageSize = pageSize;
     this.state = new StateStore<SearchSourceState<T>>(this.initialState);
-    this.setDebounceOptions({ debounceMs });
```
MartinCupela, line 74: This could really stay here as both classes do the same.

src/search/BaseSearchSource.ts
MartinCupela: I see that setDebounceOptions depends on executeQuery

src/messageComposer/middleware/textComposer/commands.ts
khushal87: Yes, we can do that

khushal87: Done

src/search/BaseSearchSource.ts
khushal87: Yes that is the reason behind segregating it. Is it fine or do you want me to change something here? @MartinCupela

MartinCupela (approved):

PR #1559, opened 2025-06-09
feat: handle offline support for drafts

khushal87:
## CLA

- [x] Code changes are tested

The goal of the PR is to handle offline support for draft messages and also allow composing a message when the composition is not empty when the offline DB is enabled. Thanks to @isekovanic for the help with the offline support explanation here.

## Changelog

-

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1638,6 +1638,8 @@ export class Channel {
         },
       );
     }
+
+    return response;
   }
 
   /**
@@ -1668,6 +1670,8 @@ export class Channel {
         },
       );
     }
+
+    return response;
   }
 
   /**
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1626,19 +1626,7 @@ export class Channel {
         message,
       },
     );
-
-    if (response.draft) {
-      this.getClient().offlineDb?.executeQuerySafely(
-        (db) =>
-          db.upsertDraft?.({
-            draft: response.draft,
-          }),
-        {
-          method: 'upsertDraft',
-        },
-      );
-    }
-
+    console.log(response);
     return response;
   }
 
@@ -1651,27 +1639,9 @@ export class Channel {
    * @return {Promise<APIResponse>} API response
    */
   async deleteDraft({ parent_id }: { parent_id?: string } = {}) {
-    const response = await this.getClient().delete<APIResponse>(
-      this._channelURL() + '/draft',
-      {
-        parent_id,
-      },
-    );
-
-    if (response) {
-      this.getClient().offlineDb?.executeQuerySafely(
-        (db) =>
-          db.deleteDraft?.({
-            cid: this.cid,
-            parent_id,
-          }),
-        {
-          method: 'deleteDraft',
-        },
-      );
-    }
-
-    return response;
+    return await this.getClient().delete<APIResponse>(this._channelURL() + '/draft', {
+      parent_id,
+    });
   }
 
   /**
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1620,12 +1620,24 @@ export class Channel {
    * @return {Promise<CreateDraftResponse>} Response containing the created draft
    */
   async createDraft(message: DraftMessagePayload) {
-    return await this.getClient().post<CreateDraftResponse>(
+    const response = await this.getClient().post<CreateDraftResponse>(
       this._channelURL() + '/draft',
       {
         message,
       },
     );
+
+    if (response.draft) {
+      this.getClient().offlineDb?.executeQuerySafely(
+        (db) =>
+          db.upsertDraft?.({
+            draft: response.draft,
+          }),
+        {
+          method: 'upsertDraft',
+        },
+      );
+    }
   }
 
   /**
@@ -1637,9 +1649,25 @@ export class Channel {
    * @return {Promise<APIResponse>} API response
    */
   async deleteDraft({ parent_id }: { parent_id?: string } = {}) {
-    return await this.getClient().delete<APIResponse>(this._channelURL() + '/draft', {
-      parent_id,
-    });
+    const response = await this.getClient().delete<APIResponse>(
+      this._channelURL() + '/draft',
+      {
+        parent_id,
+      },
+    );
+
+    if (response) {
+      this.getClient().offlineDb?.executeQuerySafely(
+        (db) =>
+          db.deleteDraft?.({
+            cid: this.cid,
+            parent_id,
+          }),
+        {
+          method: 'deleteDraft',
+        },
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/channel.ts
```diff
@@ -1637,9 +1651,27 @@ export class Channel {
...
-      parent_id,
-    });
+    const response = await this.getClient().delete<APIResponse>(
+      this._channelURL() + '/draft',
+      {
+        parent_id,
+      },
+    );
+
+    if (response) {
```
isekovanic: Nothing guarantees that this response will not be an errored one with this check. You should be able to reproduce it if you just throw some random error on `client.delete`.

test/unit/MessageComposer/messageComposer.test.ts
```diff
@@ -376,6 +377,66 @@ describe('MessageComposer', () => {
...
+        compositionContext: compositionContext || mockChannel,
+        composition,
+        config,
+      });
+
+      return { mockClient, mockChannel, messageComposer };
+    };
+
+    it('hasSendableData should return false if the composition is empty', () => {
+      const { messageComposer } = offlineModeMessageComposerSetup();
```
isekovanic, line 416: There is no reason to do this on every test. Please switch to a `beforeEach` clause and an `afterEach` for cleanup.

src/channel.ts
```diff
@@ -1620,12 +1620,26 @@ export class Channel {
...
-    return await this.getClient().post<CreateDraftResponse>(
+    const response = await this.getClient().post<CreateDraftResponse>(
       this._channelURL() + '/draft',
       {
         message,
       },
     );
+
+    if (response.draft) {
+      this.getClient().offlineDb?.executeQuerySafely(
```
isekovanic: This query execution is missing tests.

src/channel.ts
isekovanic: Same as [here](https://github.com/GetStream/stream-chat-js/pull/1559/files#r2137810249).

test/unit/MessageComposer/messageComposer.test.ts
```diff
@@ -376,6 +377,66 @@ describe('MessageComposer', () => {
...
+
+      const spyCompositionIsEmpty = vi
+        .spyOn(messageComposer, 'compositionIsEmpty', 'get')
+        .mockReturnValue(true);
+
+      expect(messageComposer.client.offlineDb).toBeDefined();
```
isekovanic: This `expect` is redundant - whether `setOfflineDBApi` extends the `client` properly is tested elsewhere already.

src/offline-support/offline_support_api.ts
```diff
@@ -944,6 +973,25 @@ export abstract class AbstractOfflineDB implements OfflineDBApi {
...
+      if (event.draft) {
+        return await this.upsertDraft({ draft: event.draft, execute });
+      }
+    }
+
+    if (type === 'draft.deleted') {
+      if (!event.cid) return;
+
+      if (event.draft && event.draft.parent_id) {
+        return await this.deleteDraft({
```
isekovanic: Does this really need to branch out like this ? The `deleteDraft` API should be able to handle missing properties and decide internally what it's going to do (when `parent_id` is `undefined` for example).

src/offline-support/offline_support_api.ts
```diff
@@ -944,6 +973,25 @@ export abstract class AbstractOfflineDB implements OfflineDBApi {
       return await this.handleChannelVisibilityEvent({ event, execute });
     }
 
+    if (type === 'draft.updated') {
```
isekovanic: None of the `draft` specifics are tested here. For example, there is no test that guarantees that if `event.draft` is missing (or `event.cid`) we'd actually do the correct thing.

I would highly recommend encapsulating all of this into a `handleDraftEvent` abstraction that would then be easy to test (and all we need to check is whether that one is called with the correct event from `handleEvent`.

Please check how the other more complicated events are handled.

isekovanic: Additionally, that would make it much easier to later on add either new draft events or handle them in a different way if needed.

test/unit/offline-support/offline_support_api.test.ts
```diff
@@ -1511,6 +1520,18 @@ describe('OfflineSupportApi', () => {
               queryInput.unreadMessages = 0;
             }
 
+            if (['draft.updated'].includes(type)) {
```
isekovanic:
```suggestion
            if (type === draft.updated') {
```

test/unit/offline-support/offline_support_api.test.ts
```diff
@@ -1511,6 +1520,18 @@ describe('OfflineSupportApi', () => {
...
+              queryInput = { draft: event.draft, execute: true };
+            }
+
+            if (['draft.deleted'].includes(type)) {
```
isekovanic:
```suggestion
            if (type === 'draft.deleted') {
```

test/unit/MessageComposer/messageComposer.test.ts
khushal87: Well, it's a one-liner, ideally, so it didn't make sense to do it in a beforerEach to complicate it. If you still think I can do it and for afterEach we have nothing.

test/unit/MessageComposer/messageComposer.test.ts
khushal87: I added it for a sanity check but I can remove if you think its useless

src/channel.ts
khushal87: I lately realised that we have these DB operations handled on the `handleEvent` so we might not need them here at all. With every creation/updation, we receive `draft.updated` event and for deletion, we receive `draft.deleted` event. With these operations here, we might be doing repetitive work.

khushal87: Let me know what do you think @isekovanic

src/offline-support/offline_support_api.ts
khushal87: Done

src/offline-support/offline_support_api.ts
khushal87: Improved it

src/channel.ts
khushal87: Decided to remove this and rely on events completely. Discussed offline.

test/unit/MessageComposer/messageComposer.test.ts
isekovanic: Well, if new tests are added in the future someone has to again do this rather than relying on already existing utilities from testing libraries. I don't care too much about it but it's better. You're literally emulating what `beforeEach` would anyway do - and also you're creating a headache for yourself if tomorrow we want to reset mocks/spies since you'd need to always do it at the end of a test as well.

Leaving it up to you though, I don't mind either way for now.

PR #1560, opened 2025-06-10
fix: poll composer improvements

isekovanic:
## CLA

- [ ] Code changes are tested

This PR does a couple of things:

- Removes `messageComposer.pollComposer.initState()` from the `createPoll` api
  - This is being done because we both do not yet know if the composer should clear the state or not; especially if sending the message fails or the creation dialog gets closed with a delay (while the HTTP requests finish)
- Introduces a new middleware for injection of `user` data from the `client` for each `localMessage`
  - This serves as an enriched optimistic update for messages, since the `user` was missing for polls as the middleware chain quits earlier than that
- An optimistic update for poll creation
  - We now populate the `poll_manager`'s cache whenever we create a poll so that by the time the optimistic update for the poll message arrives we already have the data in the cache, ready to be used  

## Changelog

-

### DIFF
--- a/src/messageComposer/middleware/messageComposer/cleanData.ts
+++ b/src/messageComposer/middleware/messageComposer/cleanData.ts
@@ -29,7 +29,6 @@ export const createCompositionDataCleanupMiddleware = (
           ...composer.editedMessage,
           ...state.localMessage,
           ...common,
-          user: composer.client.user,
         }),
         message: {
           ...editedMessagePayloadToBeSent,
--- a/src/poll_manager.ts
+++ b/src/poll_manager.ts
@@ -49,7 +49,13 @@ export class PollManager extends WithSubscriptions {
   public createPoll = async (poll: CreatePollData) => {
     const { poll: createdPoll } = await this.client.createPoll(poll);
 
-    return new Poll({ client: this.client, poll: createdPoll });
+    if (!createdPoll.vote_counts_by_option) {
+      createdPoll.vote_counts_by_option = {};
+    }
+
+    this.setOrOverwriteInCache(createdPoll);
+
+    return this.fromState(createdPoll.id);
   };
 
   public getPoll = async (id: string) => {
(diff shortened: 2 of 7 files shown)


src/messageComposer/middleware/messageComposer/cleanData.ts
```diff
@@ -29,7 +29,6 @@ export const createCompositionDataCleanupMiddleware = (
           ...composer.editedMessage,
           ...state.localMessage,
           ...common,
-          user: composer.client.user,
```
isekovanic, line 32: This is handled earlier so it's no longer needed here

src/poll_manager.ts
```diff
@@ -49,7 +49,13 @@ export class PollManager extends WithSubscriptions {
   public createPoll = async (poll: CreatePollData) => {
     const { poll: createdPoll } = await this.client.createPoll(poll);
 
-    return new Poll({ client: this.client, poll: createdPoll });
+    if (!createdPoll.vote_counts_by_option) {
```
isekovanic, line 52: Unfortunately, `vote_counts_by_option` is `null` within the `createPoll` api response (despite the type being strict and not allowing `null` or `undefined`). 

This is a workaround to bypass potentially breaking someone who might not be taking this into account and relying on the type correctness (as they should).

src/poll_manager.ts
```diff
@@ -49,7 +49,13 @@ export class PollManager extends WithSubscriptions {
...
+      createdPoll.vote_counts_by_option = {};
+    }
+
+    this.setOrOverwriteInCache(createdPoll);
```
isekovanic, line 56: This is explicitly not `this.setOrOverwriteInCache(createdPoll, true);` as it should both never happen and is wrong if it does happen somehow.

src/poll_manager.ts
MartinCupela: Is this reported to the BE?

isekovanic: I have not yet, but I will !

MartinCupela (approved):

PR #1563, opened 2025-06-13
fix: use ref counting in WithSubscriptions abstract

arnautov-anton:

Stack-based cleanups don't work with components as components may dissapear from the view at any point in time, ref counting is a better approach to this problem.

Co-authored-by: Ivan <@isekovanic>

### DIFF
--- a/test/unit/MessageComposer/messageComposer.test.ts
+++ b/test/unit/MessageComposer/messageComposer.test.ts
@@ -447,6 +447,28 @@ describe('MessageComposer', () => {
       expect(unsubscribeFunctions.size).toBe(0);
     });
 
+    it('should unregister subscriptions only if refCount drops to 1', () => {
+      const { messageComposer } = setup();
+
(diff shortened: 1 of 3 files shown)


test/unit/MessageComposer/messageComposer.test.ts
```diff
@@ -447,6 +447,28 @@ describe('MessageComposer', () => {
...
+
+      const unsub1 = messageComposer.registerSubscriptions();
+      const unsub2 = messageComposer.registerSubscriptions();
+
+      expect(spy).toHaveBeenCalledTimes(2);
+      const lastResult = spy.mock.results.at(-1);
+
+      expect(lastResult?.value).toEqual(2);
+
+      unsub1();
```
arnautov-anton, line 463: Call `unsub1` first to show that order does not matter.

isekovanic (approved):
Awesome ! Thank you @arnautov-anton <3

PR #1564, opened 2025-06-13
feat: [CHA-926] add global file/image upload methods

kmitrovv:
## CLA

- [x] Code changes are tested

Add support for new global file/image upload endpoints, decoupled from channels.

This change introduces methods for uploading and deleting files/images at the app level, independent of any specific channel. The implementation reuses existing types and mirrors the structure of the channel-based methods, but without requiring a channel context.

Only method definitions are included in this change, as the core logic remains the same.

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4570,4 +4570,64 @@ export class StreamChat {
       ...rest,
     });
   }
+
+  /**
+   * uploadFile - Uploads a file to the configured storage (defaults to Stream CDN)
+   *
+   * @param {string|NodeJS.ReadableStream|Buffer|File} uri The file to upload
+   * @param {string} [name] The name of the file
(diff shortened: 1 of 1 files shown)


totalimmersion (approved):

MartinCupela:
@kmitrovv shall we add `@deprecation` warning to the `Channel.sendImage` and `Channel.sendFile` methods?

MartinCupela:
@kmitrovv should we migrate our existing code that relies on `Channel.sendImage` and `Channel.sendFile`?

kmitrovv:
@MartinCupela we're not deprecating the existing upload endpoints, these new ones are separate and will operate at the app level. They’re intended for different use cases, such as uploading user profile images (avatars), which isn’t currently possible due to channel-level permission restrictions.

PR #1566, opened 2025-06-18
fix: offline drafts issues and optimistic update implementation

isekovanic:
## CLA

- [ ] Code changes are tested

## Changelog

This PR addresses a couple of bugs with drafts, most notable of which:

1. Removed drafts are not reflected on the state upon an additional `queryChannels` (for example, going offline and regaining connection while having changed the drafts from another device)
2. A removed draft from a channel is not reflected on the state (same case as above)
3. Drafts are left as zombie objects within the offline DB in certain scenarios
4. Drafts not working at all while offline and trying to add a draft
5. Drafts in `ThreadList` that belong to threads from the same channel all update their state in unison (and also delete their state)

Additionally, it introduces:

- Optimistic updates in the offline DB for drafts (both creation and deletion)
- Draft updates/deletes while offline
- Pending `createDraft` tasks
- Pending `deleteDraft` tasks

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -406,6 +406,17 @@ export class Channel {
     );
   }
 
+  /**
+   * sendReaction - Sends a reaction to a message. If offline support is enabled, it will make sure
+   * that sending the reaction is queued up if it fails due to bad internet conditions and executed
+   * later.
+   *
+   * @param {string} messageID the message id
+   * @param {Reaction} reaction the reaction object for instance {type: 'love'}
+   * @param {{ enforce_unique?: boolean, skip_push?: boolean }} [options] Option object, {enforce_unique: true, skip_push: true} to override any existing reaction or skip sending push notifications
+   *
+   * @return {Promise<ReactionAPIResponse>} The Server Response
+   */
   async sendReaction(
     messageID: string,
     reaction: Reaction,
@@ -421,7 +432,7 @@ export class Channel {
     try {
       const offlineDb = this.getClient().offlineDb;
       if (offlineDb) {
-        return (await offlineDb.queueTask({
+        return await offlineDb.queueTask<ReactionAPIResponse>({
           task: {
             channelId: this.id as string,
             channelType: this.type,
@@ -429,7 +440,7 @@ export class Channel {
             payload: [messageID, reaction, options],
             type: 'send-reaction',
           },
-        })) as ReactionAPIResponse;
+        });
       }
     } catch (error) {
       this._client.logger('error', `offlineDb:send-reaction`, {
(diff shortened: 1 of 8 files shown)


src/channel.ts
```diff
@@ -1613,13 +1622,11 @@ export class Channel {
...
    * createDraft - Creates or updates a draft message in a channel
    *
-   * @param {string} channelType The channel type
-   * @param {string} channelID The channel ID
    * @param {DraftMessagePayload} message The draft message to create or update
    *
    * @return {Promise<CreateDraftResponse>} Response containing the created draft
    */
-  async createDraft(message: DraftMessagePayload) {
+  async _createDraft(message: DraftMessagePayload) {
```
MartinCupela, line 1629: What if an integrator wanted to perform pure back-end API call? Will this be `channel._createDraft()`?

src/messageComposer/messageComposer.ts
```diff
@@ -464,7 +484,7 @@ export class MessageComposer extends WithSubscriptions {
       const draft = event.draft as DraftResponse;
       if (
         !draft ||
-        !!draft.parent_id !== !!this.threadId ||
+        (draft.parent_id ?? null) !== (this.threadId ?? null) ||
```
MartinCupela, line 487: Why this change?

MartinCupela:
Isn't this more a feat kind of commit than a fix? This PR is probably not fixing the offline support but rather introducing it?

src/offline-support/types.ts
```diff
@@ -399,14 +399,17 @@ export type PendingTaskTypes = {
...
 };
 
 // TODO: Please rethink the definition of PendingTasks as it seems awkward
 export type PendingTask = {
   channelId?: string;
   channelType?: string;
-  messageId: string;
+  messageId?: string;
   id?: number;
+  parentId?: string;
```
MartinCupela: I am wondering whether in the future we could be working with parent of something else than a message and then `parentId` would not have enough specificity. I am thinking whether `threadId` would be less ambivalent.

src/utils.ts
```diff
@@ -330,6 +330,33 @@ export function formatMessage(
   } as LocalMessage;
 }
 
+/**
+ * Takes a LocalMessage, parses the dates back to strings,
+ * and converts the message back to a MessageResponse.
+ *
+ * @param {MessageResponse} message `MessageResponse` object
+ */
+export function unformatMessage(message: LocalMessage): MessageResponse {
```
MartinCupela, line 341: Is this really the opposite of `formatMessage` even though it keeps error, status, reaction_groups?

src/channel.ts
isekovanic: No, they can still use `createDraft` - it will even be better for them because queueing would happen automatically (if they have offline support enabled). `queueTask` immediately returns the response from the HTTP request if it succeeds and queues it up if it does not.

So, in a purely server-side scenario for example this would just work as it did before.

src/messageComposer/messageComposer.ts
isekovanic: This fixes point number 5. - in `ThreadList`, if you have 2 or more `thread`s that belong to the same channel and you update the draft of one of them all of them would update with the same value (because `!!draft.parent_id !== !!this.threadId` is `false` for something like `draft.parent_id === '123'` and `this.threadId === '321'`.

isekovanic:
> Isn't this more a feat kind of commit than a fix? This PR is probably not fixing the offline support but rather introducing it?

Perhaps, offline support has already been merged for drafts but this sort of extends it and fixes the issues of purely offline draft creation/handling. I can change it to `feat` though it should also be fine.

src/utils.ts
isekovanic: As far as I know `formatMessage` just modifies these in a way (or enriches them), it doesn't introduce them explicitly right ? I might be wrong here, but we can add this if needed.

src/offline-support/types.ts
isekovanic: Sure, this can make sense. I'll do this.

src/utils.ts
MartinCupela: 1. `reaction_groups` are introduced as null if the data does not exists or they are calculated
2. `status` is added if it does not exist
3. `error` is introduced as null if it does not exist

I have no problem with the logic, just trying to make sure that the function will not be used by integrators as an opposite to `formatMessage`.

isekovanic: Hmm, I think you're right. I'll have a look at these on how viable it is to reverse the formatting process like this. Otherwise I might name this differently to avoid confusion, thanks !

isekovanic: Looking at the implementation of the other defaults it appears that we "lose" some data (some potential nullish values from `MessageResponse`). Since we cannot revert these, would it maybe make more sense to mark this as `private` ? Either that or a rename if you prefer

src/offline-support/types.ts
isekovanic: Resolved [here](c52f328974b107ee9bf15a69d835b435a4e1c83c).

src/utils.ts
isekovanic: As decided on Slack, will keep this as `@private` for now and maybe revisit later if we want to make it fully backwards compatible with `formatMessage`.

MartinCupela (approved):

PR #1576, opened 2025-07-01
feat: user should be able to appeal against flag

varbhat:
User should be able to appeal against the flag by moderation system. Basically if a message from a user gets blocked by AI moderation, user should be able to appeal against that decision somehow. And such appeals should be visible in moderation dashboard on the review queue item


src/moderation.ts
```diff
@@ -186,6 +192,56 @@ export class Moderation {
...
 
+  /**
+   * Appeal against the moderation decision
+   * @param {string} text
+   * @param {string} entityID
+   * @param {string} entityType
+   * @param {Array<string>} attachments
+   * @param {string} channelCID
+   */
+  async appeal(
```
MartinCupela: @varbhat would it be possible to have a single object argument with these positional arguments as keys inside? It is a better DX to avoid so many positional params I believe.

PR #1581, opened 2025-07-11
fix: add duration field to the local attachment if its available

khushal87:


test/unit/MessageComposer/attachmentManager.test.ts
```diff
@@ -1339,4 +1339,65 @@ describe('AttachmentManager', () => {
       expect(result.localMetadata.id).toBe(originalId);
     });
   });
+
+  describe('fileToLocalUploadAttachment', () => {
```
MartinCupela, line 1344: @khushal87 could you create a test that will test the whole fileToLocalUploadAttachment function? Not only whether duration is returned. I think that way we would avoid having to create a separate test for every property.

khushal87: Done @MartinCupela. It wasn't possible to cover in a single test but I have added 2-3 to cover most of the conditions. Should be fine. :)

MartinCupela (approved):

PR #1585, opened 2025-07-29
fix: set loading state to false when query channel errors out

khushal87:
The PR focuses on the following:
1. Set the loading state to false if the query channel API errors out. This was essential as currently the channel loads infinitely.
2. When the sync API returns an response timedout error from the BE we should not clear/reset the DB. Doing it was resulting in empty channel list even if the data should be there in the DB. This happened because we just cleared the DB.

### DIFF
--- a/src/channel_manager.ts
+++ b/src/channel_manager.ts
@@ -317,7 +317,14 @@ export class ChannelManager extends WithSubscriptions {
           `Maximum number of retries reached in queryChannels. Last error message is: ${err}`,
         );
 
-        this.state.partialNext({ error: wrappedError });
+        this.state.partialNext({
+          error: wrappedError,
+          pagination: {
+            ...this.state.getLatestValue().pagination,
+            isLoading: false,
+            isLoadingNext: false,
+          },
+        });
         return;
       }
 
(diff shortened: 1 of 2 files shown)


isekovanic (changes requested):
Looks good to me ! Please write tests for this

test/unit/offline-support/offline_support_api.test.ts
```diff
@@ -2325,6 +2325,42 @@ describe('OfflineDBSyncManager', () => {
         expect(upsertUserSyncStatusSpy).toHaveBeenCalled();
       });
 
+      it('do not resets DB if sync API throws an AxiosError with request timeout', async () => {
```
isekovanic: *totally not important nitpick*
```suggestion
      it('do not reset the DB if the sync API throws an AxiosError with request timeout', async () => {
```

test/unit/offline-support/offline_support_api.test.ts
```diff
@@ -2325,6 +2325,42 @@ describe('OfflineDBSyncManager', () => {
...
+        } as AxiosError<APIErrorResponse>;
+
+        syncApiSpy.mockRejectedValueOnce(axiosError);
+
+        await (syncManager as any).sync();
+
+        expect(resetDBSpy).not.toHaveBeenCalled();
+        expect(upsertUserSyncStatusSpy).not.toHaveBeenCalled();
+      });
+
```
isekovanic, line 2363: In the test just below this one, please add an additional check for the `channelManager.state.isLoading` value (it should now be `false` with the recent fixes)

PR #1602, opened 2025-08-15
feat: added ChannelMemberResponse to MessageResponse

javierdfm:
## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -692,7 +692,7 @@ export type MessageResponseBase = MessageBase & {
     language: TranslationLanguages;
   };
   latest_reactions?: ReactionResponse[];
-  member?: ChannelMemberResponse | null;
+  member?: ChannelMemberResponse;
   mentioned_users?: UserResponse[];
   message_text_updated_at?: string;
   moderation?: ModerationResponse; // present only with Moderation v2


javierdfm:
> @javierdfm what is the purpose of this change please?

hey @MartinCupela this is regarding this feature: https://linear.app/stream/issue/CHA-1102/include-channel-member-roles-in-message-payload 
the idea is to include the `ChannelMember` object in the message response with the ChannelRole available. As of now we only include the information about the User that created it, but not what the role of that user in the channel is.

src/types.ts
```diff
@@ -692,6 +692,7 @@ export type MessageResponseBase = MessageBase & {
     language: TranslationLanguages;
   };
   latest_reactions?: ReactionResponse[];
+  member?: ChannelMemberResponse | null;
```
MartinCupela: @javierdfm what would value null mean?

javierdfm: Actually, I need to leave it as `member?: ChannelMemberResponse;` `null` it won't mean anything in this case.

MartinCupela: 👍

MartinCupela:
@javierdfm one more thing that this change is missing is to update the client state. In this case it means update `Channel.state.members` mapping. So just changing the type declaration seems just like a partial change. The JS SDK team will have to add some code to extract members from messages to the Channel state.

MartinCupela (approved):

PR #1604, opened 2025-08-19
feat: add deleted_for_me field in message response

daksh-r:
## Goal

handle delete-for-me scenario in messages:

1. hard delete and delete-for-me are mutually excluding
2. introduce options object as a second parameter to `client.deleteMessage()` method
3. enrich the delete message response message object with `type` `'deleted'` and `deleted_for_me` boolean flag

### DIFF
--- a/test/unit/client.test.js
+++ b/test/unit/client.test.js
@@ -934,7 +934,7 @@ describe('message deletion', () => {
 		await client.offlineDb.init(client.userID);
 
 		loggerSpy = vi.spyOn(client, 'logger').mockImplementation(vi.fn());
-		clientDeleteSpy = vi.spyOn(client, 'delete').mockResolvedValue({});
+		clientDeleteSpy = vi.spyOn(client, 'delete').mockResolvedValue({ message: {} });
 		queueTaskSpy = vi.spyOn(client.offlineDb, 'queueTask').mockResolvedValue({});
 	});
 


MartinCupela:
@daksh-r if I understand it well, we still need to update `client.deleteMessage()` and `client._deleteMessage()` implementation. This will also have impact on offline support (right @isekovanic?).

for the user that deleted it for himself: 

What will be the `message.type` value? Will it be `deleted` or `regular`?
What will be the `message.text` value?

MartinCupela:
@daksh-r  why the WS event has the following format:

```
type: "message.deleted"
deleted_for_me: true
message: { … }
```

but the message object will contain `deleted_for_me` too (as declared in `MessageResponseBase`)?

Isn't that duplication? Is it necessary to have `deleted_for_me` at the root of the WS event object?

daksh-r:
> @daksh-r why the WS event has the following format:
> 
> ```
> type: "message.deleted"
> deleted_for_me: true
> message: { … }
> ```
> 
> but the message object will contain `deleted_for_me` too (as declared in `MessageResponseBase`)?
> 
> Isn't that duplication? Is it necessary to have `deleted_for_me` at the root of the WS event object?

in the ws event, we don't send `deleted_for_me` inside the message {...} , it only gets sent once at the root of the `message.deleted` event, when a client subsequently tries to fetch the message(via the queryChannels or getMessage or getManyMessages endpoints) that's when we populate the `deleted_for_me` field in there.

src/types.ts
```diff
@@ -1455,6 +1456,7 @@ export type Event = CustomEventData & {
   connection_id?: string;
   // event creation timestamp, format Date ISO string
   created_at?: string;
+  deleted_for_me?: boolean;
```
MartinCupela, line 1480: Is it necessary to add this property to the WS event payload root if it is already contained inside the `message` object? I suppose that this would be message.deleted event that carries the `message`.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+
+		expect(result.message.deleted_for_me).toBeFalsy();
+		expect(result.message.deleted_at).toBeUndefined();
+		expect(otherUserClient.get).toHaveBeenCalledWith(
+			`${otherUserClient.baseURL}/messages/${encodeURIComponent(messageId)}`,
+			{},
+		);
+	});
+
+	it('should preserve deleted_for_me field in message responses', () => {
```
MartinCupela: It seems to me that this test just checks what has been previously assigned to the variables `messageWithDeletedForMe` and `normalMessage`.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+		const result = await deletingUserClient.getMessage(messageId);
+
+		expect(result.message.deleted_for_me).toBe(true);
+		expect(deletingUserClient.get).toHaveBeenCalledWith(
+			`${deletingUserClient.baseURL}/messages/${encodeURIComponent(messageId)}`,
+			{},
+		);
+	});
+
+	it('should return normal message when other user fetches the message', async () => {
```
MartinCupela: It seems that this test just tests, whether the spy returns the value it was told to return.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+		};
+
+		deletingUserClient.dispatchEvent(eventForDeletingUser);
+
+		// Other user should not receive any event
+		expect(otherUserEventSpy).not.toHaveBeenCalled();
+		expect(deletingUserEventSpy).toHaveBeenCalledOnce();
+	});
+
+	it('should return deleted_for_me=true when deleting user fetches the message', async () => {
```
MartinCupela: It seems that this test just tests, whether the spy returns the value it was told to return.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+			message: deletedMessage,
+		};
+
+		deletingUserClient.dispatchEvent(eventForDeletingUser);
+
+		expect(deletingUserEventSpy).toHaveBeenCalledWith(eventForDeletingUser);
+		expect(otherUserEventSpy).not.toHaveBeenCalled();
+	});
+
+	it('should not send event to other users when message is deleted for me', () => {
```
MartinCupela: I believe this test is not testing what it declares. The test itself invokes WS event handlers for `deletingUserClient` but not for `otherUserClient`. It is normal that `otherUserEventSpy` will not be called if it is not registered with `deletingUserClient`. I think this test can be removed.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+	it('should send delete_for_me=true to API when deleteForMe is true', async () => {
+		await deletingUserClient.deleteMessage(messageId, false, true);
+
+		expect(deletingUserClient.delete).toHaveBeenCalledWith(
+			`${deletingUserClient.baseURL}/messages/${encodeURIComponent(messageId)}`,
+			{ delete_for_me: true },
+		);
+	});
+
+	it('should handle message.deleted event with deleted_for_me=true for deleting user', () => {
```
MartinCupela: I believe this test is not testing what it declares. The test itself invokes WS event handlers for `deletingUserClient` but not for `otherUserClient`. It is normal that `otherUserEventSpy` will not be called if it is not registered with `deletingUserClient`. I think this test can be removed.

test/unit/client.test.js
MartinCupela: I think this test can be removed.

test/unit/client.test.js
MartinCupela: I think this test can be removed.

test/unit/client.test.js
MartinCupela: I think this test can be removed.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
...
+		});
+	});
+
+	afterEach(() => {
+		vi.resetAllMocks();
+		deletingUserClient.off('message.deleted', deletingUserEventSpy);
+		otherUserClient.off('message.deleted', otherUserEventSpy);
+	});
+
+	it('should send delete_for_me=true to API when deleteForMe is true', async () => {
```
MartinCupela: This test can be removed as it tests what is already tested in the previous tests.

test/unit/client.test.js
```diff
@@ -1019,6 +1058,171 @@ describe('message deletion', () => {
 	});
 });
 
+describe('delete for me functionality', () => {
```
MartinCupela: The whole describe block can be removed.

src/types.ts
MartinCupela: I see that I already pointed this out and see that the `message` inside the WS event is not getting the property.

daksh-r: the property has been added recently, can you tell for which request you aren't receiving this ? 

you can find the latest sdk spec [here](https://www.notion.so/stream-wiki/Design-Doc-Delete-Message-for-Me-Feature-2376a5d7f9f68086af20c49f172a091d?source=copy_link#2546a5d7f9f6807ba5acc18487fd6248)

test/unit/client.test.js
MartinCupela: Done with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

test/unit/client.test.js
MartinCupela: Done with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

test/unit/client.test.js
MartinCupela: Removed with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

test/unit/client.test.js
MartinCupela: Removed with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

test/unit/client.test.js
MartinCupela: Removed with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

test/unit/client.test.js
MartinCupela: Removed with https://github.com/GetStream/stream-chat-js/pull/1604/commits/c7cc175ee867278978b32bf96fec7a7f9345a7c7

src/types.ts
MartinCupela: I was just wondering why the property is duplicated in the message WS event root, when it is available in the message object in the same WS event

image

MartinCupela: no big deal

src/client.ts
```diff
@@ -3103,9 +3104,27 @@ export class StreamChat {
...
+  ): Promise<APIResponse & { message: MessageResponse }> {
+    const { hardDelete } = (
+      typeof optionsOrHardDelete === 'boolean'
+        ? { hardDelete: optionsOrHardDelete }
+        : (optionsOrHardDelete ?? {})
+    ) as DeleteMessageOptions;
+
     try {
       if (this.offlineDb) {
+        // todo: write code for this.offlineDb.deleteForMe
```
MartinCupela: @isekovanic is this something that can be done later? Will there be need to actually reflect this somehow with the offline DB?

PR #1605, opened 2025-08-26
fix: formatMessage - reference the correct msg in toLocalMessageBase

ginnymin:
## CLA

- [x] Code changes are tested

When upgrading our Stream packages at Glue, we noticed that our quoted messages were displaying the same `created_at` date and time as the messages they were attached to after the upgrade. After some debugging we isolated the issue to the upgrade, and after looking further, noticed that the code that converts `MessageResponse` to `LocalMessage` - i.e. `formatMessage` in `src/utils` - was referencing the original message in `toLocalMessageBase` rather than the `msg` passed to it, which would cause the issue we are seeing.

## Changelog

- Updated `message` references in `toLocalMessageBase` to `msg`
- Updated tests to correspond with this change

### DIFF
--- a/test/unit/channel.test.js
+++ b/test/unit/channel.test.js
@@ -458,7 +458,7 @@ describe('Channel _handleChannelEvent', function () {
 		expect(
 			channel.state.messages.find((msg) => msg.id === quotingMessage.id).quoted_message
 				.deleted_at,
-		).to.be.null;
+		).to.be.ok;
 	});
 
 	describe('user.messages.deleted', () => {
@@ -577,6 +577,8 @@ describe('Channel _handleChannelEvent', function () {
 							...deletedMessage,
 							id: message.quoted_message.id,
 							user: message.quoted_message.user,
+							created_at: message.quoted_message.created_at,
+							updated_at: message.quoted_message.updated_at,
 						},
 					});
 				} else {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -310,16 +310,16 @@ export function formatMessage(
     if (!msg) return null;
     return {
       ...msg,
-      created_at: message.created_at ? new Date(message.created_at) : new Date(),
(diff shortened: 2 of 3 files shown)


test/unit/channel.test.js
```diff
@@ -458,7 +458,7 @@ describe('Channel _handleChannelEvent', function () {
 		expect(
 			channel.state.messages.find((msg) => msg.id === quotingMessage.id).quoted_message
 				.deleted_at,
-		).to.be.null;
+		).to.be.ok;
```
ginnymin, line 461: I changed this to what is was before the [message composer changes](https://github.com/GetStream/stream-chat-js/commit/0c07524f6551e9257b229b262b4d1e03ab44561b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) were made, which changed this to `null`.

I believe the expected behavior here is that the `deleted_at` property of a deleted quoted message does not remain `null`.

MartinCupela:
Hello @ginnymin , thank you for filing the PR and spotting this bug.

MartinCupela (approved):

test/unit/channel.test.js
```diff
@@ -577,6 +577,7 @@ describe('Channel _handleChannelEvent', function () {
 							...deletedMessage,
 							id: message.quoted_message.id,
 							user: message.quoted_message.user,
+							created_at: message.quoted_message.created_at,
```
MartinCupela, line 580: @ginnymin could you please add updated_at so that the tests can pass?
```suggestion
							created_at: message.quoted_message.created_at,
							updated_at: message.quoted_message.updated_at,
```

ginnymin: Updated!

MartinCupela (approved):

PR #1607, opened 2025-08-27
feat: storing participant count in ThreadState

kenjush:
## Linear

- https://linear.app/stream/issue/CHA-1051/expose-zero-value-for-participant-count

## Linked PR

- https://github.com/GetStream/chat/pull/9720

## CLA

Expose a Thread's participant count through storing it in the `ThreadState` 

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -47,8 +47,6 @@ export type ThreadState = {
   read: ThreadReadState;
   replies: Array<LocalMessage>;
   replyCount: number;
-  participantCount: number;
-  activeParticipantCount: number;
   title: string;
   updatedAt: Date | null;
 };
@@ -113,6 +111,7 @@ const constructCustomDataObject = <T extends ThreadResponse>(threadData: T) => {
 
 export class Thread extends WithSubscriptions {
   public readonly state: StateStore<ThreadState>;
+  public readonly threadData: ThreadResponse;
   public readonly id: string;
   public readonly messageComposer: MessageComposer;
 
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -511,12 +511,12 @@ export interface ThreadResponse extends CustomThreadData {
   parent_message_id: string;
   title: string;
   updated_at: string;
-  active_participant_count?: number;
+  active_participant_count: number;
```
kanat: I guess, this may potentially break if the SDK is used against an older backend version.

kenjush: That's correct. I guess once we update the SDK, any new install might run into this issue till we deploy the new backend. I'll update it. Thanks.

src/thread.ts
```diff
@@ -167,6 +169,8 @@ export class Thread extends WithSubscriptions {
       ),
       replies: threadData.latest_replies.map(formatMessage),
       replyCount: threadData.reply_count ?? 0,
+      participantCount: threadData.participant_count ?? 0,
```
isekovanic: Can `participantCount` and `activeParticipantCount` change with every websocket event ? If not, which are the ones that could contribute to these changing ? Off the top of my head, it would probably be something like:

- `message.new` (getting tagged in a `thread` or replying to one on your own)
- `message.deleted` (same rationale' as `message.new`)
- `message.updated`

etc.

Regardless, we would need to upsert the state with these values if we're putting them within the state itself, unless these are a one-time thing and can't be updated otherwise (which is likely not the case, since participants can change dynamically).

MartinCupela: Yes, we would need to get the count data in WS events if we do not want to show just the initial count on thread data load.

kanat: As far as I can see the thread object is included in the following events:
- thread.updated
- message.read
- notification.mark_read

PR #1617, opened 2025-09-09
feat: add message delivery receipts

MartinCupela:
## Goal

Provide:

1. **API for message delivery confirmation**
```
client.syncDeliveredCandidates(collections: Channel[]) // in the future (Channel | Thread)[]
await client.markChannelsDelivered(data?: MarkDeliveredOptions)
```

2. **service for buffering of delivery confirmation payload** (`client.deliveryReportCoordinator` is handling the buffering and throttling of the requests). See **Buffering service** section in the spec.

3. **WS event processing for read state updates** (`message.new`, `message.read`, `notification.mark_unread`, `notification.channel_delivered` - name may change)

Closes REACT-501

## Spec

See description of CHA-991

## Out of scope
The buffering service is prepared to handle delivery confirmation for threads too, but it is disabled for the moment.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -2129,6 +2129,11 @@ export class Channel {
         };
 
         channelState.unreadCount = unreadCount;
+        this.messageReceiptsTracker.onNotificationMarkUnread({
+          user: event.user,
+          lastReadAt: event.last_read_at,
+          lastReadMessageId: event.last_read_message_id,
+        });
         break;
       }
       case 'channel.updated':
--- a/src/messageDelivery/MessageReceiptsTracker.ts
+++ b/src/messageDelivery/MessageReceiptsTracker.ts
@@ -175,6 +175,40 @@ export class MessageReceiptsTracker {
     }
   }
 
+  /** message.delivered — user device confirmed delivery up to and including messageId. */
+  onMessageDelivered({
+    user,
+    deliveredAt,
+    lastDeliveredMessageId,
+  }: {
+    user: UserResponse;
+    deliveredAt: string;
+    lastDeliveredMessageId?: string;
+  }) {
+    const timestampMs = new Date(deliveredAt).getTime();
+    const msgRef = lastDeliveredMessageId
+      ? { timestampMs, msgId: lastDeliveredMessageId }
+      : this.locateMessage(new Date(deliveredAt).getTime());
+    if (!msgRef) return;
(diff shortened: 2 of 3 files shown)


src/channel.ts
```diff
@@ -1940,6 +1975,10 @@ export class Channel {
                   last_read: new Date(event.created_at as string),
                   user: event.user,
                   unread_messages: 0,
+                  last_delivered_at: event.last_delivered_at
```
MartinCupela: Does it make sense to have it here?

src/channel.ts
```diff
@@ -1872,13 +1883,36 @@ export class Channel {
             last_read_message_id: event.last_read_message_id,
             user: event.user,
             unread_messages: 0,
+            last_delivered_at: event.last_delivered_at
```
MartinCupela: Probably not necessary to update delivery data here if we have read data

src/client.ts
```diff
@@ -4692,4 +4696,25 @@ export class StreamChat {
...
+  /**
+   * Send the mark delivered event for this user, only works if the `delivery_receipts` setting is enabled
+   *
+   * @param {MarkDeliveredOptions} data
+   * @return {Promise<EventAPIResponse | void>} Description
+   */
+  async markChannelsDelivered(data?: MarkDeliveredOptions) {
+    const deliveryReceiptsEnabled =
+      this.user?.privacy_settings?.delivery_receipts?.enabled;
+    if (!deliveryReceiptsEnabled) return;
```
rafaelmf3: we could return some feedback to know that the privacy_setting is missing

src/channel.ts
rafaelmf3, line 1: we should add to _initializeState the last_delivered_message_id and last_delivered_at to the state read object

MartinCupela: Added here: https://github.com/GetStream/stream-chat-js/pull/1617#:~:text=delivery%20receipt%20data-,e4a140b,-refactor%3A%20rename%20event

src/channel.ts
MartinCupela: will keep the data in sync with the back-end

src/channel.ts
MartinCupela: Adding fallback to the current delivery state in case the event does not carry the info

src/client.ts
MartinCupela: I don't plan it in this PR.

PR #1635, opened 2025-10-09
feat: add an api to prune messages from the end

isekovanic:
## CLA

- [ ] Code changes are tested

This PR introduces a new API in `ChannelState` that can be used to prune messages from the back of our current `messageSet` and update all pagination related properties. 

This can be useful in situations where we might have a huge number of messages that we do not necessarily want to keep in memory (i.e a livestream) and want to gradually prune while still allowing normal pagination to work as intended.

## Changelog

-

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,10 +1,13 @@
 export const DEFAULT_QUERY_CHANNELS_MESSAGE_LIST_PAGE_SIZE = 25;
 export const DEFAULT_QUERY_CHANNEL_MESSAGE_LIST_PAGE_SIZE = 100;
-export const DEFAULT_MESSAGE_SET_PAGINATION = { hasNext: false, hasPrev: false };
+export const DEFAULT_MESSAGE_SET_PAGINATION = Object.freeze({
+  hasNext: false,
+  hasPrev: false,
+});
 export const DEFAULT_UPLOAD_SIZE_LIMIT_BYTES = 100 * 1024 * 1024; // 100 MB
 export const API_MAX_FILES_ALLOWED_PER_MESSAGE = 10;
 export const MAX_CHANNEL_MEMBER_COUNT_IN_CHANNEL_QUERY = 100;
-export const RESERVED_UPDATED_MESSAGE_FIELDS = {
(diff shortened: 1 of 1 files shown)


src/channel_state.ts
```diff
@@ -821,7 +830,7 @@ export class ChannelState {
         messages: [],
         isLatest: true,
         isCurrent: true,
-        pagination: DEFAULT_MESSAGE_SET_PAGINATION,
+        pagination: { ...DEFAULT_MESSAGE_SET_PAGINATION },
```
isekovanic, line 833: These changes are not specifically related to the PR, but are a bug I found by accident.

Namely, whenever we use `DEFAULT_MESSAGE_SET_PAGINATION` directly, modifying pagination as `messageSet.pagination.something = someValue` for example will cause `DEFAULT_MESSAGE_SET_PAGINATION` to be updated during runtime, since its reference is kept. 

This causes all sorts of issues as any message set we orchestrate from that point on is going to have faulty `pagination` options.

It's noticeable by tests 2 and 3 failing in this PR.

oliverlaz: To prevent accidental update of assumed constants like this, I suggest using `Object.freeze`
```ts
const MY_DEFAULT = Object.freeze({
  value: 42
});
```

Also, keep in mind this isn't a deep freeze :)

src/channel_state.ts
```diff
@@ -152,6 +152,15 @@ export class ChannelState {
     );
   }
 
+  pruneFromEnd(maxMessages: number) {
```
oliverlaz: `pruneFromEnd` might not be the best name, as it can give a wrong impression (end being the end of the channel (last message), or beginning of the channel).

Maybe `pruneOldest` or something similar?

isekovanic: Sure that works, will change it

isekovanic: Done [here](https://github.com/GetStream/stream-chat-js/pull/1635/commits/265720bcbdfbe06bd0d0cc2e12fa84b2a484b054), forgot to push it last Friday it seems.

PR #1638, opened 2025-10-14
fix: prevent search source pagination on non-retryable response errors

MartinCupela:
## Goal

Closes React-607

Prevent repeating queries with a search string that will always end up in API error response. By default errors are not considered reason to switch `hasNext` flag to `false`. Non-retryable errors should however lead to that.

Example:

Infinite scroll pagination repeating requests with the same search string that causes the error. Leads to infinite loop.

### DIFF
--- a/src/search/ChannelSearchSource.ts
+++ b/src/search/ChannelSearchSource.ts
@@ -3,12 +3,7 @@ import type { FilterBuilderOptions } from '../pagination';
 import { FilterBuilder } from '../pagination';
 import type { Channel } from '../channel';
 import type { StreamChat } from '../client';
-import type {
-  APIErrorResponse,
-  ChannelFilters,
-  ChannelOptions,
-  ChannelSort,
-} from '../types';
+import type { ChannelFilters, ChannelOptions, ChannelSort } from '../types';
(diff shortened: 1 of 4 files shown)


src/search/ChannelSearchSource.ts
```diff
@@ -70,4 +75,20 @@ export class ChannelSearchSource<
...
+      /StreamChat error code \d+: .*failed with error: "(.*)"/,
+    )?.[1];
+    if (!originalErrorString) return false;
+    try {
+      const originalError = JSON.parse(originalErrorString) as APIErrorResponse;
+      return /field is empty or contains invalid characters/.test(originalError.message);
+    } catch (e) {
+      return false;
+    }
+  };
```
oliverlaz: This implementation is quite heuristic. Maybe extend this list with fatal/non-fatal information?
- https://github.com/GetStream/stream-chat-js/blob/master/src/errors.ts#L4

MartinCupela: Fixed here: [a2a15b3](https://github.com/GetStream/stream-chat-js/pull/1638/commits/a2a15b385fa8d40b7053abb06957669831b204d7)

PR #1649, opened 2025-11-07
feat: add app_banned and has_unread properties to ChannelFilters

SubashPradhan:
## CLA

- [ YES] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [YES] Code changes are tested

 - The QueryChannels filter currently supports `app_banned` in the filter: https://getstream.io/chat/docs/javascript/query_channels/
 - But this is not included in the Filter types, causing TS error
Screenshot 2025-11-07 at 10 47 30
- Add `app_banned` type to the Channel Filter types

## Changelog
-

### DIFF
--- a/test/unit/utils.test.ts
+++ b/test/unit/utils.test.ts
@@ -690,35 +690,6 @@ describe('Channel pinning and archiving utils', () => {
       expect(shouldConsiderArchivedChannels(mockFilters)).to.be.true;
     });
   });
-
-  describe('app_banned filter', () => {
-    it('should accept "only" as a valid value', () => {
-      const mockFilters: ChannelFilters = { app_banned: 'only' };
-      expect(mockFilters.app_banned).to.equal('only');
-    });
-
-    it('should accept "excluded" as a valid value', () => {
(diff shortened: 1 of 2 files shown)


test/unit/utils.test.ts
```diff
@@ -690,6 +690,35 @@ describe('Channel pinning and archiving utils', () => {
       expect(shouldConsiderArchivedChannels(mockFilters)).to.be.true;
     });
   });
+
+  describe('app_banned filter', () => {
```
oliverlaz: Hey @SubashPradhan, besides testing the type-safety of the `ChannelFilters` object, these tests don't test any behavior. Instead, they assert that the value set a few lines above is persisted.

I suggest removing them and preventing further confusion.

MartinCupela (approved):

test/unit/utils.test.ts
SubashPradhan: @oliverlaz Thanks for the feedback, it makes sense to remove them.

PR #1656, opened 2025-11-18
feat:  batch channel updates

javierdfm:
## CLA

- [ ] Code changes are tested

https://linear.app/stream/issue/CHA-1608/js

## Changelog

-

### DIFF
--- a/src/channel_batch_updater.ts
+++ b/src/channel_batch_updater.ts
@@ -67,7 +67,7 @@ export class ChannelBatchUpdater {
     members: string[] | NewMemberPayload[],
   ): Promise<APIResponse & UpdateChannelsBatchResponse> {
     return await this.client.updateChannelsBatch({
-      operation: 'invites',
+      operation: 'inviteMembers',
       filter,
       members,
     });
--- a/src/types.ts
+++ b/src/types.ts
@@ -4499,7 +4499,7 @@ export type EventHook = {
 export type BatchUpdateOperation =
   | 'addMembers'
   | 'removeMembers'
-  | 'invites'
+  | 'inviteMembers'
   | 'assignRoles'
   | 'addModerators'
   | 'demoteModerators'


src/channel_batch_updater.ts
```diff
@@ -0,0 +1,247 @@
...
+   * @param {UpdateChannelsBatchFilters} filter Filter to select channels
+   * @param {string[] | NewMemberPayload[]} members Members to invite
+   * @return {Promise<APIResponse & UpdateChannelsBatchResponse>} The server response
+   */
+  async inviteMembers(
+    filter: UpdateChannelsBatchFilters,
+    members: string[] | NewMemberPayload[],
+  ): Promise<APIResponse & UpdateChannelsBatchResponse> {
+    return await this.client.updateChannelsBatch({
+      operation: 'invites',
```
MartinCupela: Not sure if possible, but to be consistent with removeMembers, the operation should be called `'inviteMembers'`?

src/channel_batch_updater.ts
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  /**
+   * inviteMembers - Invite members to channels matching the filter
+   *
+   * @param {UpdateChannelsBatchFilters} filter Filter to select channels
+   * @param {string[] | NewMemberPayload[]} members Members to invite
+   * @return {Promise<APIResponse & UpdateChannelsBatchResponse>} The server response
+   */
+  async inviteMembers(
```
MartinCupela, line 65: Just a question.  What happens under the hood, when we call inviteMembers? How does it differ from addMembers?

rafaelmf3: We have this two options, or just add the member "without permission" or invite the member, this second one needs the member agree with that, the user can disagree and will not be added as a member

src/channel_batch_updater.ts
rafaelmf3: @javierdfm can answer this more properly, but I think the idea is it should be something clear with the abstraction layer for example:

channel.inviteMembers() -> client.updateChannelsBatch()

the operation will be the "user level" layer and the string indicating the operation will be more internal

javierdfm: I followed the same naming as in the single channel update:
https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L850-L872

Please let me know if it is really an issue (we already have a different naming)

PR #1658, opened 2025-11-20
test: fix scheduled test

mogita:
## Goal

This PR attempts to fix the scheduled test by adding the build step.

**Changes:**

- Import from the built bundle instead of trying to access internal utils
- Created a simple UUID function for testing purposes

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -162,6 +162,11 @@ export type AppSettingsAPIResponse = APIResponse & {
     name?: string;
     organization?: string;
     permission_version?: string;
+    /**
+     * The placement of the app in the form of `${region}.${shard}`.
+     * Examples: "us-east.c1", "dublin.c3", "singapore.c2"
+     * Note: The backend may add/remove regions or shards occasionally.
+     */
     placement?: string;
     policies?: Record<string, Policy[]>;
     poll_enabled?: boolean;


src/types.ts
```diff
@@ -140,12 +146,23 @@ export type AppSettingsAPIResponse = APIResponse & {
...
+    moderation_dashboard_preferences?: Record<string, unknown> | null;
+    moderation_enabled?: boolean;
+    moderation_llm_configurability_enabled?: boolean;
+    moderation_multitenant_blocklist_enabled?: boolean;
+    moderation_webhook_url?: string;
     multi_tenant_enabled?: boolean;
     name?: string;
     organization?: string;
     permission_version?: string;
+    placement?: string;
```
MartinCupela, line 170: @mogita is this any string or a set of predefined strings?

mogita: It's a composite string in the form of `${region}.${shard}`. A thorough type def might be something like this:

```typescript
type Region = 'canada' | 'dublin' | 'frankfurt' | 'mumbai' | 'oregon' | 
              'singapore' | 'sydney' | 'us-east' | 'ohio' | 'london' | 
              'jakarta' | 'stockholm';

type Shard = 'c1' | 'c2' | 'c3' | 'c4' | 'c5' | 'c6' | 'c7' | 'c8' | 
             'c9' | 'c10' | 'c11' | 'c12' | 'c13' | 'c14' | 'c15' | 
             'c16' | 'c17';

type Placement = `${Region}.${Shard}`;
```

But I don't recommend we bake these into the JS library. The backend could add/remove region or shard at any moment.

MartinCupela: Could we add this as a doc string for the field, please?

mogita: Updated.

Issue #1661, opened 2025-12-09
Query Channels by `filter_tags` TypeScript type error


fgirardey:
The following code give the following TypeScript type error, am I missing something?

```ts
const channels = await this.streamChat.queryChannels({
	filter_tags: { $in: [`my_tag`] },
});
```

```
Type '{ $in: string[]; }' is not assignable to type 'RequireOnlyOne<{ $eq?: PrimitiveFilter<string[] | undefined>; $exists?: boolean | undefined; $in?: PrimitiveFilter<string[] | undefined>[] | undefined; }, "$eq" | ... 1 more ... | "$in"> | PrimitiveFilter<...>'.
  Types of property '$in' are incompatible.
    Type 'string[]' is not assignable to type 'PrimitiveFilter<string[] | undefined>[]'.
      Type 'string' is not assignable to type 'PrimitiveFilter<string[] | undefined>'.ts(2322)
types.d.ts(258, 5): The expected type comes from property 'filter_tags' which is declared here on type 'ChannelFilters'
```

fgirardey:
Thank you so much, you guys are always really helpful and reactive, I love that!

PR #1663, opened 2025-12-10
fix: improve type inference for $in filter in queryChannels

oliverlaz:
## Changelog

Improves the type inference for the `$in` operator in `queryChannels()` filters. 
Fixes: #1661.

Array types weren't properly unpacked, so TS types required nested arrays:
```ts
const channels = await this.streamChat.queryChannels({
   filter_tags: { $in: [`my_tag`] }, // errors
   filter_tags: { $in: [[`my_tag`]]` }, // doesn't error on TS but our API expects a flat array
});
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2022,7 +2022,7 @@ export type QueryFilter<ObjectType = string> =
         $exists?: boolean;
         $gt?: PrimitiveFilter<ObjectType>;
         $gte?: PrimitiveFilter<ObjectType>;
-        $in?: PrimitiveFilter<Unpacked<ObjectType>>[];
+        $in?: PrimitiveFilter<ObjectType>[];
         $lt?: PrimitiveFilter<ObjectType>;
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -2022,14 +2022,14 @@ export type QueryFilter<ObjectType = string> =
         $exists?: boolean;
         $gt?: PrimitiveFilter<ObjectType>;
         $gte?: PrimitiveFilter<ObjectType>;
-        $in?: PrimitiveFilter<ObjectType>[];
+        $in?: PrimitiveFilter<Unpacked<ObjectType>>[];
```
MartinCupela: Is Unpacked necessary here if ObjectType will be something assignable to string | number | boolean?

oliverlaz: hmm, probably not. Let me revert it.

oliverlaz: reverted.

PR #1673, opened 2026-01-06
feat: extend sort with field type

arnautov-anton:

This PR allows integrators to decide how the target field should be asserted during sorting.

https://getstream.slack.com/archives/C06CF5TKRGA/p1767637494321509?thread_ts=1767298663.322459&cid=C06CF5TKRGA

```ts
const sort: ChannelSort = {
  member_count: { direction: -1, type: 'number' },
  has_unread: -1,
};
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2168,32 +2168,37 @@ export type MemberFilters = QueryFilters<
 
 export type BannedUsersSort = BannedUsersSortBase | Array<BannedUsersSortBase>;
 
-export type BannedUsersSortBase = { created_at?: AscDesc };
+export type BannedUsersSortBase = Sort<'created_at'>;
 
 export type ReactionSort = ReactionSortBase | Array<ReactionSortBase>;
 
-export type ReactionSortBase = Sort<CustomReactionData> & {
-  created_at?: AscDesc;
-};
+export type ReactionSortBase = Sort<CustomReactionData> & Sort<'created_at'>;
 
 export type ChannelSort = ChannelSortBase | Array<ChannelSortBase>;
 
-export type ChannelSortBase = Sort<CustomChannelData> & {
-  created_at?: AscDesc;
-  has_unread?: AscDesc;
-  last_message_at?: AscDesc;
-  last_updated?: AscDesc;
-  member_count?: AscDesc;
-  pinned_at?: AscDesc;
-  unread_count?: AscDesc;
-  updated_at?: AscDesc;
-};
+export type ChannelSortBase = Sort<CustomChannelData> &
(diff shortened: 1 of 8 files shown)


src/types.ts
```diff
@@ -2168,32 +2168,37 @@ export type MemberFilters = QueryFilters<
...
-  [P in keyof T]?: AscDesc;
+export type PinnedMessagesSortBase = Sort<'pinned_at'>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Sort<T extends string | Record<string, any>> = {
+  [P in T extends string ? T : keyof T]?:
+    | AscDesc
+    | {
+        direction: AscDesc;
+        type?: 'string' | 'number';
```
arnautov-anton, line 2200: @kanat what types will be supported?

src/utils.ts
```diff
@@ -131,20 +132,26 @@ export function addFileToFormData(
...
-    if (entries.length > 1) {
-      console.warn(
-        "client._buildSort() - multiple fields in a single sort object detected. Object's field order is not guaranteed",
-      );
-    }
-    for (const [field, direction] of entries) {
-      sortFields.push({ field, direction });
+
+export function normalizeQuerySort<T extends Sort<string>>(sort: T | T[]) {
+  const sortFields: Array<{ direction: AscDesc; field: keyof T; type: string | null }> =
```
MartinCupela, line 137:
```suggestion
  const sortFields: Array<{ direction: AscDesc; field: keyof T; type: string | number | null }> =
```

Not sure, but basing it on this:

https://github.com/GetStream/stream-chat-js/pull/1673/changes#r2664806559

arnautov-anton: The type is just a string in this case as it's supposed to be sent to the BE, TS should not remove this during compile time. In this case it's just generalized to `string` so I don't have to do `"string" | "number" | "some-other-type-once-added"` since we don't really care about specificity at that point.

src/types.ts
```diff
@@ -2212,51 +2217,40 @@ export type MemberSort =
...
+  >;
 
 export type SearchMessageSort = SearchMessageSortBase | Array<SearchMessageSortBase>;
 
 export type QuerySort = BannedUsersSort | ChannelSort | SearchMessageSort | UserSort;
 
-export type DraftSortBase = {
-  created_at?: AscDesc;
-};
+export type DraftSortBase = Sort<'created_ats'>;
```
MartinCupela, line 2241:
```suggestion
export type DraftSortBase = Sort<'created_at'>;
```

PR #1675, opened 2026-01-12
feat: [CHA-1584] - Predefined filters support

totalimmersion:


src/types.ts
```diff
@@ -4544,3 +4559,52 @@ export type UpdateChannelsBatchFilters = QueryFilters<{
...
+  sort?: PredefinedFilterSortParam[];
+};
+
+export type UpdatePredefinedFilterOptions = Omit<CreatePredefinedFilterOptions, 'name'>;
+
+export type PredefinedFilterResponse = APIResponse & {
+  predefined_filter: PredefinedFilter;
+};
+
+export type PredefinedFiltersResponse = APIResponse & {
```
MartinCupela:
```suggestion
export type ListPredefinedFiltersResponse = APIResponse & {
```

I would adjust the name to make it more distinguishable from PredefinedFilterResponse for a reader & to keep it consistent with the options type name: ListPredefinedFiltersOptions

src/types.ts
```diff
@@ -4544,3 +4559,52 @@ export type UpdateChannelsBatchFilters = QueryFilters<{
...
+
+/**
+ * Predefined Filter Types
+ */
+
+export type PredefinedFilterOperation = 'QueryChannels';
+
+export type PredefinedFilterSortParam = {
+  field: string;
+  direction?: 1 | -1;
```
MartinCupela:
```suggestion
  direction?: AscDesc;
```

src/types.ts
```diff
@@ -1011,6 +1011,21 @@ export type ChannelOptions = {
...
+  /**
+   * Name of a predefined filter to use instead of filter_conditions.
+   * When provided, filter_conditions and sort parameters are ignored.
+   */
+  predefined_filter?: string;
+  /**
+   * Values to interpolate into the predefined filter template placeholders.
+   * Only used when predefined_filter is provided.
+   */
+  filter_values?: Record<string, unknown>;
```
MartinCupela, line 1023: Keys in this object will need to correspond to some specific entity field names or it can really be any string?

MartinCupela: @totalimmersion Could you give an example of interpolation, please?

src/types.ts
```diff
@@ -4544,3 +4559,52 @@ export type UpdateChannelsBatchFilters = QueryFilters<{
...
   result: Record<string, string>;
 } & Partial<TaskResponse>;
+
+/**
+ * Predefined Filter Types
+ */
+
+export type PredefinedFilterOperation = 'QueryChannels';
+
+export type PredefinedFilterSortParam = {
```
MartinCupela, line 4569: @totalimmersion there was a discussion btw @kanat and @arnautov-anton regarding `type` sort field. Is this change maybe related to the general change in sort fields? Would [this PR](https://github.com/GetStream/stream-chat-js/pull/1673), where `type` field is introduced be relevant for your PR?

src/client.ts
```diff
@@ -4817,4 +4832,82 @@ export class StreamChat {
...
+   * @return {Promise<PredefinedFiltersResponse>} The list of predefined filters
+   */
+  async listPredefinedFilters(options: ListPredefinedFiltersOptions = {}) {
+    this.validateServerSideAuth();
+    const { sort, ...paginationOptions } = options;
+    return await this.get<PredefinedFiltersResponse>(
+      `${this.baseURL}/predefined_filters`,
+      {
+        ...paginationOptions,
+        ...(sort ? { sort: JSON.stringify(sort) } : {}),
```
MartinCupela, line 4909: @totalimmersion Why do we convert the object into string? I see in types.ts the following declarations implying and usually objects are not stringified in other API calls in other methods:

```
export type PredefinedFilterSort = SortParam[];

export type ListPredefinedFiltersOptions = Pager & {
  sort?: PredefinedFilterSort;
};
```

src/types.ts
totalimmersion: can be any string

totalimmersion: examples can be found here: https://github.com/GetStream/chat/blob/6e1e70fe87168db2cf0279fad5ceb6a0fb7e772c/qa/tests/query_channels_predefined_filters.test.ts

src/types.ts
totalimmersion: not sure about that discussion, this pr follows current practices

src/client.ts
totalimmersion: because it's a GET request

MartinCupela (approved):

PR #1697, opened 2026-03-10
feat: add Channel and Thread display name and image generators

MartinCupela:
## Goal

Have unified logic for Channel and Thread display name and image generation for all the UI SDKs.

In case of no image generated, the Avatar component would decide how many channel member images are used in it.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -115,6 +115,8 @@ export class Channel {
   public readonly messageComposer: MessageComposer;
   public readonly messageReceiptsTracker: MessageReceiptsTracker;
   public readonly cooldownTimer: CooldownTimer;
+  public customDisplayNameGenerator?: (channel: Channel) => string | null | undefined;
+  public customDisplayImageGenerator?: (channel: Channel) => string | null | undefined;
 
   /**
    * constructor - Create a channel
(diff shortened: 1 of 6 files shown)


oliverlaz (approved):

src/channel.ts
```diff
@@ -196,6 +198,82 @@ export class Channel {
...
+    }
+
+    const members = Object.values(this.state.members);
+    const currentUserId = this._client.userID;
+    const otherMembers = members.filter((member) => member.user?.id !== currentUserId);
+
+    // 3. Fall back 1:1 channels: other member's name, then id, then "Direct Message"
+    if (members.length === 2 && otherMembers.length === 1) {
+      const otherUser = otherMembers[0].user;
+      return otherUser?.name || otherUser?.id || 'Direct Message';
```
isekovanic, line 235: Shouldn't the default if none of these exist stay in the SDK for translation purposes ? Talking about `Direct Message`

src/channel.ts
```diff
@@ -196,6 +198,82 @@ export class Channel {
...
+    }
+
+    // 4. Fall back group channels: comma-separated list of other members' names (or ids); ellipsis if >2
+    if (otherMembers.length > 0) {
+      const names = otherMembers
+        .map((member) => member.user?.name || member.user?.id)
+        .filter(Boolean);
+      if (names.length > 0) {
+        return names.length > 2 ? `${names.slice(0, 2).join(', ')}...` : names.join(', ');
```
isekovanic, line 244: Maybe we should let the various platforms handle `elipsis` on their own instead of enforcing it within the name

src/thread.ts
```diff
@@ -194,6 +197,36 @@ export class Thread extends WithSubscriptions {
...
+      if (custom) return custom;
+    }
+
+    if (currentState.title) return currentState.title;
+
+    const { participants } = currentState;
+    if (participants && participants.length > 0) {
+      const names = participants.map((p) => p.user?.name || p.user?.id).filter(Boolean);
+      if (names.length > 0) {
+        return names.length > 2 ? `${names.slice(0, 2).join(', ')}...` : names.join(', ');
```
isekovanic, line 223: same comment as above

src/channel.ts
```diff
@@ -196,6 +198,82 @@ export class Channel {
...
+
+  /**
+   * Returns the display image URL for this channel using the following fallback chain:
+   * 1. Result of `customDisplayImageGenerator` (if set on this instance)
+   * 2. The channel's `image` custom data property
+   * 3. `null` if none of the above produced a value
+   *
+   * @return {string | null}
+   */
+  getDisplayImage(): string | null {
```
isekovanic, line 259: I'm not 100% sure this should be in the LLC specifically, as it's a very UI thing (and we then have to add a `customDisplayImageGenerator` to be able to override 2 lines of code, ignoring linting). I'm not inherently against it, I'm just struggling to see its usefulness

src/channel.ts
MartinCupela: Yes, I was thinking how to solve this and probably better to keep it there.

src/channel.ts
MartinCupela: ok

src/thread.ts
MartinCupela: ok

src/channel.ts
MartinCupela: The purpose was to have this unified across the SDKs

PR #1715, opened 2026-04-01
feat: global upload manager to track file uploads outside of AttachmentManager

szuperaz:
## CLA

- [ ] Code changes are tested

RN PR for reference: https://github.com/GetStream/stream-chat-react-native/pull/3527

React Native allows hitting the send button before attachment uploads are completed. This means they need a way to track attachment upload status outside of `AttachmentManager`.

To solve this we have an `UploadManager` (can be accessed via `client.uploadManager`) that uses the `AttachmentManager` of a given channel to do attachment upload, and has a state to track `uploadProgress` for attachments.

Attachments are tracked by `id` in `UploadManager`.

Long-term solution: we talked with @isekovanic that `messageComposer` should have support for this feature, but that is a bigger task, not the scope of this PR.

## Changelog

-

### DIFF
--- a/src/uploadManager.ts
+++ b/src/uploadManager.ts
@@ -8,31 +8,26 @@ export type UploadRecord = {
 };
 
 export type UploadManagerState = {
-  uploads: UploadRecord[];
+  uploads: Record<string, UploadRecord>;
 };
 
-const initState = (): UploadManagerState => ({ uploads: [] });
+const initState = (): UploadManagerState => ({ uploads: {} });
 
-const upsertById = (uploads: UploadRecord[], record: UploadRecord): UploadRecord[] => {
-  const idx = uploads.findIndex((u) => u.id === record.id);
-  if (idx === -1) return [...uploads, record];
-  const current = uploads[idx];
-  if (current === record) return uploads;
-  const next = [...uploads];
-  next[idx] = { ...current, ...record };
-  return next;
-};
+const upsertById = (
+  uploads: Record<string, UploadRecord>,
+  record: UploadRecord,
+): Record<string, UploadRecord> => ({
+  ...uploads,
+  [record.id]: { ...uploads[record.id], ...record },
+});
 
 const updateById = (
-  uploads: UploadRecord[],
+  uploads: Record<string, UploadRecord>,
   record: UploadRecord,
-): UploadRecord[] | null => {
-  const idx = uploads.findIndex((u) => u.id === record.id);
-  if (idx === -1) return null;
-  const current = uploads[idx];
-  const next = [...uploads];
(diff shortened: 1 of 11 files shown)


src/messageComposer/middleware/messageComposer/attachments.ts
```diff
@@ -44,7 +44,13 @@ export const createAttachmentsCompositionMiddleware = (
       }
 
       const attachments = (state.message.attachments ?? []).concat(
-        attachmentManager.successfulUploads.map(localAttachmentToAttachment),
+        attachmentManager.attachments
```
MartinCupela: Why are we not using attachmentManager.successfulUploads?

src/messageComposer/middleware/messageComposer/attachments.ts
```diff
@@ -81,7 +87,13 @@ export const createDraftAttachmentsCompositionMiddleware = (
       const successfulUploads = attachmentManager.successfulUploads;
       const attachments = successfulUploads.length
         ? (state.draft.attachments ?? []).concat(
-            successfulUploads.map(localAttachmentToAttachment),
+            attachmentManager.attachments
```
MartinCupela: Why not using attachmentManager.successfulUploads?

src/messageComposer/middleware/messageComposer/attachments.ts
szuperaz: Leftover from a failed attempt, will revert

src/messageComposer/middleware/messageComposer/attachments.ts
szuperaz: Leftover from a failed attempt, will revert

src/uploadManager.ts
```diff
@@ -0,0 +1,181 @@
...
+  private upsertUpload = (record: UploadRecord) => {
+    this.state.next((current) => ({
+      ...current,
+      uploads: upsertById(current.uploads, record),
+    }));
+  };
+
+  private finalizeSuccess = (id: string, response: unknown) => {
+    // Mark finished then clear on the next state update.
+    this.upsertUpload({
```
isekovanic: So we don't forget:

As we discussed on the call yesterday, I think the `uploadManager` should be self recoverable when it comes to successful uploads (i.e only uploads that have errored out can stay post the upload finishing, since we need these anyway and they can be removed from the upstream).

Uploads that have succeeded should clean themselves up from the manager

szuperaz: As discussed, went with the route where both success and failed records are cleaned up

isekovanic (approved):

PR #1723, opened 2026-04-27
feat: advanced commands flow and composer effects

isekovanic:
## CLA

- [ ] Code changes are tested

This PR refactors command activation and disabled command handling so command availability is evaluated from live `MessageComposer` state instead of being stored on command search results.

Command disabled state depends on mutable composer state:

- whether the composer in in an `editing` state
- whether the composer is has a reply
- which command is being evaluated

Initially, I had implemented this with a `disabled` and `disabledReason` state extension for the search source related to commands. However, this makes it super awkward for the UI SDKs as search results are not reactive (and we can't rely on `suggestions` for things like the commands picker content). Because of that, `disabled`/`disabledReason` have been moved away from `CommandSearchSource` results (keeping it purely CRUD). Since those results can be reused by different UI surfaces and they may become stale if composer state changes after the query runs, we rely purely on runtime logic for this.

Instead, we expose 2 new APIs:

- `messageComposer.getCommandDisabledReason(command)`
- `messageComposer.isCommandDisabled(command)`

We check whether a command is disabled as early in the response as possible so that we prevent stale command objects from activating after composer state changes.

Additionally, `setQuotedMessage(...)` and `setEditedMessage(...)` clear the currently active command if the new composer state makes it invalid. This prevents the composer from remaining in a disallowed active command state after reply/edit state changes.

### Effects

For the purpose of achieving all of the above, we introduce `effects` as a concept to middleware execution. Effects are essentially deferred side-effects that need to happen in multiple different stores but triggered from one. For example, setting a command (i.e `textComposer`) needs to take care of clearing `text`, `mentions` and `attachments` - parts of the state which belong to `textComposer.state` and `attachmentManager.state`. The easy solution here would be to force `textComposer` to also mutate `attachmentManager` (which we can do), however that's inherently wrong as even though we have access to `composer` we should not introduce such tight coupling between the two. 

Command effects remain scoped to activation side effects only:

- snapshot current text/mentions/attachments
- clear command entry state
- restore previous state when clearing the command

They do not control command availability. Removing command activation effects still opts out of snapshot/clear behavior, but disabled-command guards are evaluated separately at runtime.

As a note, if in the future we want to introduce `effects` that are potentially concurrently unsafe - the proper way to solve this would be to introduce an `effect` registry that would then make sure that these are executed sequentially (or at the very least topologically, if it makes sense for the `effects`). However for this iteration it seemed like a bit of an overkill to do that and so I decided to keep it simple.

The definition of `effects` happens strictly within `middleware`s. Their execution however, is tied to `messageComposer` explicitly. Customizing `effects` has many approaches. As an example:

```
messageComposer.textComposer.middlewareExecutor.insert({
    middleware: {
      id: 'custom/remove-command-activation-effect',
      handlers: {
        onChange: ({ state, next }) => {
          if (!state.effects?.length) return next(state);

          return next({
            ...state,
            effects: state.effects.filter(
              (effect) => effect.type !== 'command.activate',
            ),
          });
        },
        onSuggestionItemSelect: ({ state, next }) => {
          if (!state.effects?.length) return next(state);

          return next({
            ...state,
            effects: state.effects.filter(
              (effect) => effect.type !== 'command.activate',
            ),
          });
        },
      },
    },
    position: { after: 'stream-io/text-composer/commands-middleware' },
  });
```

would remove all effects before they even reach execution.

The following:

```
messageComposer.registerEffectHandler<TextComposerCommandActivationEffect>(
    'command.activate',
    (effect, composer) => {
      // Keep Stream's command state update.
      composer.textComposer.state.partialNext({
        command: effect.command,
        suggestions: undefined,
      });

      // Custom behavior: do not clear attachments or mentions.
      composer.textComposer.state.partialNext({
        text: '',
        selection: { start: 0, end: 0 },
      });
    },
  );

messageComposer.registerEffectHandler<CustomComposerEffect>(
    'custom.effect',
    (effect, composer) => {
      console.log(effect.payload, composer.id);
    },
  );
```

overrides the default handling for `command.activate` and adds a `custom.effect` that does something arbitrary. 

## Changelog

-


src/messageComposer/messageComposer.ts
```diff
@@ -414,6 +474,74 @@ export class MessageComposer extends WithSubscriptions {
...
+    handler?.(effect, this);
+  };
+
+  private applyCommandActivationEffect = (
+    effect: TextComposerCommandActivationEffect,
+  ) => {
+    if (!this.preCommandStateSnapshot) {
+      const { mentionedUsers, selection, text } =
+        effect.stateToRestore ?? this.textComposer.state.getLatestValue();
+      this.preCommandStateSnapshot = {
```
MartinCupela: I think we need a more generic approach here. We should not be hard-coding specific keys but rather iterate over all the manager states and capture their snapshots (we have also CustomData manager)

MartinCupela: Probably something like `messageComposer.takeStatesSnapshot()` or `messageComposer.snapshotStates`. This method could internally call `manager.takeSnapshot()` on every manager.

src/messageComposer/messageComposer.ts
```diff
@@ -414,6 +474,74 @@ export class MessageComposer extends WithSubscriptions {
...
+        attachments: this.attachmentManager.attachments,
+        textComposer: { mentionedUsers, selection, text },
+      };
+    }
+
+    this.attachmentManager.clearAttachments();
```
MartinCupela: And here we should probably clear all the manager states, not only attachments and text.

MartinCupela:
My proposal is not to introduce a notion of effects as they seem an overkill at the moment. Signaling that a command was activated can be seen from the `TextComposer`'s state - it already has `command` key. I think the following would suffice:

#### Command activation
- `MessageComposer.activateCommand(command)` — takes a snapshot, then clears child state for command entry
- `MessageComposer.deactivateCommand()` — restores from snapshot, clears the active command


#### Snapshot API
- Each manager implements `takeSnapshot(): T and restoreSnapshot(snapshot: T): void`, owning what's snapshot-worthy in its own state
- `MessageComposer.takeSnapshot()` iterates managers, stores their snapshots
- `MessageComposer.restoreSnapshot()` iterates and restores, then clears
- `MessageComposer.clearManagerStates()` - iterates all the managers. Each manager already implements clear() knowing what "empty" means for its own state.

By each manager implementing specific API, we follow already established pattern of the methods of the same name across the managers.

And so `MessageComposer.activateCommand` can be as follows:

```
activateCommand = (command: CommandResponse) => {
  if (this.isCommandDisabled(command)) return;
  this.takeSnapshot();
  this.clearManagerStates();
  this.textComposer.setCommand(command); // already exists
};
```

```
deactivateCommand = () => {
  this.textComposer.clearCommand(); // already exists
  this.restoreSnapshot();
};
```

isekovanic:
As some general info, @MartinCupela and I decided offline how to proceed with this. Will update the description accordingly.

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -8,6 +9,54 @@ import type {
...
+};
+
+const notifyCommandDisabled = (composer: MessageComposer, command: CommandResponse) => {
+  const disabledReason = composer.getCommandDisabledReason(command);
+  if (!disabledReason) return;
+
+  composer.client.notifications.addWarning({
+    message:
+      disabledReason === 'editing'
+        ? 'Not available while editing'
```
MartinCupela:
```suggestion
        ? 'Command not available while editing'
```

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -8,6 +9,54 @@ import type {
...
+        : 'Not available while replying',
```
MartinCupela:
```suggestion
        : 'Command not available while replying',
```

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -8,6 +9,54 @@ import type {
 import type { MessageComposer } from '../../messageComposer';
 import type { MiddlewareHandlerParams } from '../../../middleware';
 
+const getRawCommandName = (text?: string) => text?.match(/^\/(\S+)(?:\s.*)?$/)?.[1];
```
MartinCupela: In 
1. `src/messageComposer/middleware/textComposer/commandStringExtraction` there is another helper `stripCommandFromText` that extracts command. 
2. src/messageComposer/middleware/textComposer/ textMiddlewareUtils there are some other extraction tools

Maybe we could unify the command extraction utils.

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -8,6 +9,54 @@ import type {
...
+
+const getCommandByName = (
+  composer: MessageComposer,
+  commandName?: string,
+): CommandResponse | undefined => {
+  if (!commandName) return;
+
+  return composer.channel
```
MartinCupela: Not sure whether we should not use `CommandSearchSource` to look for command to use unified approach to retrieve commands.

isekovanic: Thought about this too, but the problem with that is sort of that this will never really be reactive (as search results are single pass. And also it was implied by the design team that we should still show the commands that are disabled, only have them be not available in these instances and fire a notification in that instance

isekovanic: integrators can also do this on their own if they'd like, I don't think it's a complicated search source to implement

src/messageComposer/middleware/textComposer/commandEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+const createCommandActivationEffect = (
+  command: CommandResponse,
+): TextComposerCommandActivationEffect => ({
+  command,
+  stateToRestore: emptyCommandStateToRestore,
+  type: 'command.activate',
+});
+
+const isCommandResponse = (suggestion: unknown): suggestion is CommandSuggestion =>
+  typeof (suggestion as CommandSuggestion | undefined)?.name === 'string';
```
MartinCupela, line 29: Not sure if this is not fragile and could end up in false positive. Maybe we should address this by adding property `type` to `BaseSuggestion` type in the future when doing breaking changes

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -172,6 +178,28 @@ export const createCommandsMiddleware = (
         const { selectedSuggestion } = state.change ?? {};
         if (!selectedSuggestion || state.suggestions?.trigger !== finalOptions.trigger)
           return forward();
+        const composer = options?.composer;
+        const disabledReason = composer?.getCommandDisabledReason(selectedSuggestion);
+        if (composer && disabledReason) {
+          composer.client.notifications.addWarning({
```
MartinCupela: A function for this notification emission already exists in compositionValidation - maybe we could use that one?

isekovanic: oops, good catch - forgot to change this.

src/messageComposer/middleware/textComposer/commands.ts
```diff
@@ -172,6 +178,28 @@ export const createCommandsMiddleware = (
...
+            },
+            options: {
+              type: 'validation:command:disabled',
+              metadata: {
+                command: selectedSuggestion.name,
+                reason: disabledReason,
+              },
+            },
+          });
+          return next(state);
```
MartinCupela:
```suggestion
          return forward();
```

Would it make more sense?

src/messageComposer/middleware/textComposer/commandEffects.ts
isekovanic: Yeah, that's something we could do - as it is now I guess it should be alright (provided it's used specifically here). Can't think of anything better for now but open to suggestions

src/messageComposer/middleware/textComposer/commandEffects.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const createCommandEffectsMiddleware = (): CommandEffectsMiddleware => ({
+  handlers: {
+    onChange: ({ forward }) => forward(),
+    onSuggestionItemSelect: ({ state, next, forward }) => {
+      const { selectedSuggestion } = state.change ?? {};
```
MartinCupela, line 35: if effect middleware comes after commands middleware, could we just reach for `state.command` instead  of `selectedSuggestions`?

src/messageComposer/messageComposer.ts
```diff
@@ -259,6 +300,10 @@ export class MessageComposer extends WithSubscriptions {
 
   setEditedMessage = (editedMessage: LocalMessage | null | undefined) => {
     this.state.partialNext({ editedMessage: editedMessage ?? null });
+    const activeCommand = this.textComposer.command;
+    if (editedMessage && activeCommand && this.isCommandDisabled(activeCommand)) {
```
MartinCupela: Is it necessary to check for `activeCommand && this.isCommandDisabled(activeCommand)` when what this does it checks, whether we are quote replying / editing - at first look it seems redundant as we are entering into the edit mode by calling `setEditedMessage`

MartinCupela: Maybe in textComposer.clearCommand we could be checking whether there is an active command

isekovanic: Eh, I think you're right - the composer context would anyway change and the command would be wiped out regardless

isekovanic: I'll try it out to confirm, thanks

src/messageComposer/MessageComposerEffectHandlers.ts
```diff
@@ -0,0 +1,85 @@
...
+  applyEffects = <T extends { type: string }>(effects: T[] = []) => {
+    effects.forEach((effect) => this.applyEffect(effect));
+  };
+
+  private applyEffect = (effect: { type: string }) => {
+    const handler = this.handlers.get(effect.type);
+    handler?.(effect, this.options.composer);
+  };
+
+  private applyCommandActivationEffect: MessageComposerEffectHandler<TextComposerCommandActivationEffect> =
```
MartinCupela: Could this be a normal function instead of a method to avoid poluting the class API?

src/messageComposer/MessageComposerEffectHandlers.ts
```diff
@@ -0,0 +1,85 @@
...
+        text: '',
+      });
+      composer.attachmentManager.clearAttachments();
+      composer.linkPreviewsManager.clear();
+      composer.locationComposer.clear();
+      composer.pollComposer.initState();
+      composer.customDataManager.initState();
+    };
+
+  private applyCommandClearEffect: MessageComposerEffectHandler<TextComposerCommandClearEffect> =
```
MartinCupela: Could this be a normal function instead of a method to avoid poluting the class API?

src/messageComposer/MessageComposerEffectHandlers.ts
isekovanic: It can, but we'd have to drill `effect` and `composer` to it. The idea was that each applicator receives the appropriate `effect` with its metadata along with `composer` and then the implementation can decide what it does

isekovanic: I don't mind doing it that way if we prefer to though

src/messageComposer/textComposer.ts
```diff
@@ -144,15 +151,26 @@ export class TextComposer {
   }
 
   initState = ({ message }: { message?: DraftMessage | LocalMessage } = {}) => {
+    this.composer.clearSnapshots();
```
MartinCupela: Why only this manager clears all the manager snapshots in the initState?

isekovanic: Great catch, this is a leftover

src/messageComposer/middleware/messageComposer/compositionValidation.ts
MartinCupela: I am thinking that if we do not search for commands in unified way, there can be discrepancies btw the results in places where search source is used vs we approach our channel.config. In case somebody applies some custom logic in search source, then it would not be reflected here.

src/messageComposer/middleware/textComposer/commands.ts
isekovanic: It should, yes - let me give it a shot

src/messageComposer/MessageComposerEffectHandlers.ts
MartinCupela: I am asking because there is no access to `this.something` in the method, what points to the method being just a normal function

isekovanic: That's a good point. We can either stop passing composer and make sure it's injected during construction or just extract these for sure. Since they're anyway private, I think for now I'll stick with the latter (easier for integrators too if they want to register their own)

src/messageComposer/MessageComposerEffectHandlers.ts
```diff
@@ -0,0 +1,85 @@
...
+      // we manually clear because we want the command to still persist
+      composer.textComposer.state.next({
+        command: effect.command,
+        mentionedUsers: [],
+        suggestions: undefined,
+        selection: { start: 0, end: 0 },
+        text: '',
+      });
+      composer.attachmentManager.clearAttachments();
+      composer.linkPreviewsManager.clear();
```
MartinCupela: I am wondering whether `composer.linkPreviewsManager.clearPreviews()` would not be more appropriate here. If I was composing a message, I have dismissed some previews, then I decide to do command and then I return to the previous draft, I would probably not want to re-dismiss already dismissed previews.

src/messageComposer/middleware/messageComposer/compositionValidation.ts
isekovanic: Hmm that's true, it would need to be changed in 2 places I guess. But then again since `isCommandDisabled` is very specific for our base case, it would anyway not return the command as disabled and so I guess integrators can still choose what they do about it

src/messageComposer/MessageComposerEffectHandlers.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      // we manually clear because we want the command to still persist
+      composer.textComposer.state.next({
+        command: effect.command,
+        mentionedUsers: [],
+        suggestions: undefined,
+        selection: { start: 0, end: 0 },
+        text: '',
+      });
+      composer.attachmentManager.clearAttachments();
```
MartinCupela: Why don't we use `attachmentManager.initState()`

src/messageComposer/MessageComposerEffectHandlers.ts
```diff
@@ -0,0 +1,85 @@
...
+      composer.linkPreviewsManager.clear();
+      composer.locationComposer.clear();
```
MartinCupela: Why don't we use `composer.locationComposer.initState()`? For poll composer and custom data manager we are using `initState`

src/messageComposer/MessageComposerEffectHandlers.ts
isekovanic: The thing is that we're supposed to cancel ongoing enrichment as well, since we might not be back to it. From what I can see, `clearPreviews` does not do this. But, why would we need to re-dismiss already dismissed previews ? I might be missing some core logic here

src/messageComposer/MessageComposerEffectHandlers.ts
isekovanic: Because `clearAttachments` also takes care of deleting the upload records and cancelling attachments which are in progress still. if we want to do this also in `initState` then sure

src/messageComposer/MessageComposerEffectHandlers.ts
isekovanic: For this one I believe we actually can indeed, I'll move to that

src/messageComposer/middleware/messageComposer/commandNotification.ts
```diff
@@ -0,0 +1,30 @@
+import type { MessageComposer } from '../../messageComposer';
+import type { CommandResponse } from '../../../types';
+
+export const notifyCommandDisabled = (
```
MartinCupela: Can we move this to some utils file - e.g. commandUtils.ts? As this is not a middleware.

isekovanic: sure

MartinCupela (approved):

PR #1734, opened 2026-05-05
feat: add state invalidation to BasePaginator class

arnautov-anton:

Adds state invalidation logic to a `BasePaginator` class, adjusts types. Super-simplified `react-query` feature ([`invalidateQueries`](https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation)).

Ref: GetStream/stream-chat-react#3173

### DIFF
--- a/src/pagination/BasePaginator.ts
+++ b/src/pagination/BasePaginator.ts
@@ -1,9 +1,13 @@
 import { StateStore } from '../store';
 import { debounce, type DebouncedFunc } from '../utils';
 
-type PaginationDirection = 'next' | 'prev';
-type Cursor = { next: string | null; prev: string | null };
-export type PaginationQueryParams = { direction: PaginationDirection };
+type PaginationDirection = PaginationDirectionNext | PaginationDirectionPrev;
+type PaginationDirectionPrev = 'prev';
+type PaginationDirectionNext = 'next';
+type Cursor = { next?: string; prev?: string };
+export type PaginationQueryParams =
+  | { direction: PaginationDirectionNext; next?: Cursor['next']; offset?: number }
+  | { direction: PaginationDirectionPrev; prev?: Cursor['prev']; offset?: number };
 export type PaginationQueryReturnValue<T> = { items: T[] } & {
   next?: string;
   prev?: string;
@@ -24,6 +28,7 @@ export type PaginatorState<T = any> = {
   lastQueryError?: Error;
   cursor?: Cursor;
   offset?: number;
+  isStateValid: boolean;
 };
 
 export type PaginatorOptions = {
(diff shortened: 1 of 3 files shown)


src/pagination/BasePaginator.ts
```diff
@@ -124,30 +140,45 @@ export abstract class BasePaginator<T> {
...
-        : [...(this.items ?? []), ...(stateUpdate.items || [])],
+        : [...(current.items ?? []), ...(stateUpdate.items ?? [])],
     };
   }
 
   async executeQuery({ direction }: { direction: PaginationDirection }) {
     if (!this.canExecuteQuery(direction)) return;
-    const isFirstPage = typeof this.items === 'undefined';
-    if (isFirstPage) {
-      this.state.next(this.getStateBeforeFirstQuery());
```
MartinCupela, line 135: Removing this is potentially breaking

MartinCupela: As this is setting state before the query is executed. This is a hook that can be used by integrators to adjust the state before the first query.

src/pagination/BasePaginator.ts
```diff
@@ -124,30 +140,45 @@ export abstract class BasePaginator<T> {
...
+
       const { items, next, prev } = results;
+
       if (isFirstPage && (next || prev)) {
         this._isCursorPagination = true;
       }
 
       if (this._isCursorPagination) {
-        stateUpdate.cursor = { next: next || null, prev: prev || null };
+        stateUpdate.cursor = { next, prev };
```
MartinCupela, line 181: Value `null` has a meaning that there are no more pages to retrieve. Are we getting `null` from `results.next` / `results.prev` or if we arrived at the end it is `undefined`? `undefined` in paginator was dedicated for pagination not started yet.

arnautov-anton: Oh - there were no particular `null` checks anywhere and tests (both manual and written) passed without any hiccups. As for the BE API - if you finish paginating, no cursor - at least for this particular API (reactions) - seems to be returned.

src/pagination/BasePaginator.ts
arnautov-anton: State before the first query should always be `initialState` unless invalidated. When invalidated and re-fetched, the state is kept until there's something to replace it with to prevent the data -> no data -> new data UI blinking.

MartinCupela: Ok, then we should also remove getStateBeforeFirstQuery method, because in this PR it does not make anymore sense - it is called before the first query but not applied to the state before the first query.

src/pagination/BasePaginator.ts
MartinCupela: ok, then let's keep it as it is in this PR.

MartinCupela:
Entire scope unknown

PR #1735, opened 2026-05-07
feat(webhooks): verifyAndParse* API for compressed payloads (CHA-3071)

nijeesh-stream:
## Summary

Adds first-class support for **gzip-compressed webhook payloads** (HTTP webhooks, SQS, SNS) and exposes a stable `verifyAndParse*` API that mirrors the cross-SDK contract published in [Webhooks Overview](https://getstream.io/chat/docs/node/webhooks_overview/#handling-the-webhook).

### New public API (`src/signing.ts`)

Primitives:

- `gunzipPayload(body)` — gzip-magic-byte detection, no-op when not compressed
- `decodeSqsPayload(body)` — strict base64 decode then gunzip-if-magic
- `decodeSnsPayload(notificationBody)` — JSON-parse the SNS HTTP notification envelope, extract the inner `Message`, then run the SQS pipeline. Falls through to a pre-extracted `Message` string when the input is not a JSON envelope
- `verifySignature(body, signature, secret)` — HMAC-SHA256 over the **uncompressed** body, with a constant-time comparison (matters for the HTTP webhook path where the `x-signature` header is exposed publicly; SQS / SNS deliveries arrive over AWS-internal transports where timing-attack resistance is not strictly required)
- `parseEvent(payload)` — JSON → typed `Event<…>`

Composites (return a typed `Event<…>`):

- `verifyAndParseWebhook(body, signature, secret)`
- `verifyAndParseSqs(body, signature, secret)`
- `verifyAndParseSns(body, signature, secret)`

`StreamChat#verifyAndParseWebhook` / `verifyAndParseSqs` / `verifyAndParseSns` use the configured client secret automatically.

### Backwards compatibility

- `StreamChat#verifyWebhook` is preserved and now delegates to `verifySignature` (same constant-time HMAC-SHA256).
- `CheckSignature` is deprecated and aliases `verifySignature`.

### Unified error handling

Per cross-SDK coordination (mogita's review across the 6 sibling SDK PRs), every webhook failure path now terminates at a **single error class — `InvalidWebhookError`** so customers only need one `catch` arm. The message text identifies which failure mode fired so callers that want to differentiate (security logging, retry policy) can filter on substring:

| Failure mode      | Message                       |
| ----------------- | ----------------------------- |
| Signature mismatch | `signature mismatch`         |
| Base64 decode      | `invalid base64 encoding`    |
| Gzip decompression | `gzip decompression failed`  |
| JSON parse         | `invalid JSON payload`       |

`InvalidWebhookErrorMessages` is exported alongside the class for callers that prefer exact-match filtering over substring matches. The legacy `StreamChat#verifyWebhook` (bool return) is untouched.

### Tests

`test/unit/webhook-compression.test.ts` covers plain / gzip / base64 / base64+gzip payloads, signature mismatches, malformed bytes, and JSON parsing into `Event<…>`. Linked Linear ticket: [CHA-3071](https://linear.app/stream/issue/CHA-3071/compress-webhook-payloads).

### Golden test fixtures (Tommaso)

Added shared reference fixtures to the test suite so future SDKs can sanity-check decoders against the same payloads:

```
aGVsbG93b3JsZA==                          -> helloworld   (base64)
H4sIAGrYAWoAA8tIzcnJL88vykkBAK0g6/kKAAAA -> helloworld   (base64 + gzip)
```
## Test plan

- [x] `yarn vitest run test/unit/webhook-compression.test.ts` — 47 passed (was 42)
- [x] `yarn lint` — clean on touched files
- [x] `yarn prettier:check` — clean


src/signing.ts
```diff
@@ -84,19 +85,169 @@ export function DevToken(userId: string) {
...
+    throw new WebhookSignatureError('failed to base64-decode payload: malformed input');
+  }
+  return ungzipPayload(decoded);
+}
+
+/**
+ * Identical to {@link decodeSqsPayload}; exposed under both names so
+ * call sites read intent.
+ */
+export function decodeSnsPayload(message: string): Buffer {
```
mogita: **SNS envelope unwrapping vs the agreed contract**

This aliases the SQS pipeline (base64 + ungzip), but the spec is `decode_sns_payload(notification_body: string)` — the parameter name implies the full SNS HTTP notification body. Real SNS deliveries are JSON envelopes:

```json
{"Type":"Notification","MessageId":"...","Message":"<base64+gzip payload>","MessageAttributes":{...}}
```

Right now the SDK requires every caller to JSON-parse the envelope and pull out `Message` themselves before calling this. If a user passes the raw HTTP body in, signature verification runs against the envelope JSON, not the payload, and fails silently.

Same gap exists across all 6 SDKs in this rollout (JS/Go/Ruby/PHP/Java/.NET) — needs a contract decision before merge:
- **Option A (preferred):** parse the envelope here, extract `Message`, then run the base64+gzip pipeline. Matches the parameter name. Removes a caller footgun.
- **Option B:** rename to `decode_sns_message` and document explicitly that callers pre-extract.

Please align all SDKs on one path.

mogita (commented):
Cross-SDK review pass for CHA-3071. Two inline comments — see below.

test/unit/webhook-compression.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    });
+
+    it('surfaces malformed base64 as WebhookSignatureError', () => {
+      expect(() => client.verifyAndParseSqs('!!!not-base64!!!', 'sig')).toThrow(
+        WebhookSignatureError,
+      );
+    });
+  });
+
+  describe('verifyAndParseSns', () => {
```
mogita: **Missing real SNS envelope coverage**

All `verifyAndParseSns` tests here pass an already-extracted `Message` string, never a full SNS HTTP notification envelope. Whatever the contract decision is on the `decodeSnsPayload` envelope question (see the other comment), this test gap currently masks the behavior.

Recommend adding a shared fixture across all 6 SDKs using the same SNS envelope JSON, so behavior is identical and verifiable.

src/signing.ts
nijeesh-stream: Addressed in b2332ff4 with Option A: `decodeSnsPayload` now parses the SNS HTTP notification envelope (`{"Type":"Notification","Message":"..."}`) and extracts the inner `Message` field before running the SQS pipeline. When the input is not a JSON envelope it falls through to the pre-extracted `Message` path so existing callers keep working. All 7 SDKs aligned on this contract.

test/unit/webhook-compression.test.ts
nijeesh-stream: Added in b2332ff4 — `describe('decodeSnsPayload')` now covers `unwraps a full SNS HTTP notification envelope` and a whitespace-prefix variant; `describe('verifyAndParseSns')` adds a `parses a full SNS HTTP notification envelope` case plus a `rejects signature computed over the envelope JSON, not the payload` guard. Shared fixture shape (Type/MessageId/TopicArn/Message/MessageAttributes/X-Signature) is reused across all 7 SDKs.

mogita:
**Cross-SDK coordination: unifying webhook exception types**

After the review pass across all 6 SDKs in this rollout and team discussion, we're consolidating the new webhook exception strategy to a **single unified exception class** rather than the split (signature vs parse) being introduced in this PR.

The [Webhook Handling Spec on Notion (CHA-2961)](https://www.notion.so/stream-wiki/Server-Side-SDK-Webhook-Handling-Spec-34b6a5d7f9f681e78003c443f227493c) has been revised to reflect this — §5.2 / §5.3 / §7 now specify a single class.

**Why unified:** From a customer's perspective, all failure modes — signature mismatch, gzip decompression failure, base64 decode failure, SNS envelope failure, JSON parse failure, missing schema field — terminate at the same `catch` block in customer code. A signature/parse split adds structural complexity without changing customer behavior. Customers who want to filter security logs for signature mismatches specifically can do so via exception message text or cause-chain.

**Class name family:** `InvalidWebhookError` — "Invalid" covers all failure modes accurately (it's the standard stdlib term across all target languages: `InvalidArgumentException`, `InvalidOperationException`, `ValueError`, etc., as opposed to "Parse" which would undersell the signature-failure case).

Per-SDK naming across the rollout:

| SDK | Class name |
|-----|-----------|
| JS | `InvalidWebhookError` (extends `Error`) |
| Python | `InvalidWebhookError` |
| Go | sentinel `ErrInvalidWebhook` + struct `InvalidWebhookError` |
| Java | `InvalidWebhookException` (extends existing `StreamException`) |
| PHP | `InvalidWebhookException` (extends existing `StreamException`) |
| Ruby | `StreamChat::InvalidWebhookError` (extends `StandardError`) |
| .NET | `StreamInvalidWebhookException` (extends `StreamBaseException`) |

**Asks for this PR:**

1. Rename `WebhookSignatureError` → `InvalidWebhookError`
2. Wrap **all** failure paths into this single type — signature mismatch, gzip failure, base64 failure, SNS envelope failure, JSON parse failure, missing `type`/schema failure
3. Attach a human-readable message identifying which failure mode fired (e.g. `"signature mismatch"`, `"invalid base64"`, `"missing type field"`) so customers can filter on message content
4. Legacy `verifyWebhook` (returning `bool`) stays unchanged — back-compat preserved
5. Update unit tests to assert against the new exception name; for mode-specific tests, also assert on message-content substrings

This same comment is being posted on all 6 SDK PRs (JS / Go / Ruby / PHP / Java / .NET) for coordination. Happy to discuss naming or scope tradeoffs.

src/signing.ts
mogita: The [SQS & SNS Payload Compression spec](https://linear.app/stream/document/sqs-and-sns-payload-compression-spec-e95df9020e08) just landed on the Linear ticket. It reaffirms that SNS deliveries arrive as JSON envelopes `{"Type":"Notification","Message":"<base64+gzip>",...}` with the actual payload inside `Message`. SQS-via-SNS-subscribers forwards the same wire format.

The SNS envelope-unwrap question I raised earlier on this line is still the open blocker. Contract resolution needed before any of these PRs merge:

- **(A)** `decode_sns_payload` parses the envelope, extracts `Message`, then base64 + gunzip — hides the envelope shape from customers, single call site.
- **(B)** Rename helper to `decode_sns_message` and document that the caller pre-extracts `Message`.

Pick one and apply uniformly across all 6 SDKs. I'm leaning **(A)** to spare every customer from a JSON unwrap step.

nijeesh-stream: Option A is already live as of `b2332ff4` (yesterday). `decodeSnsPayload(notificationBody)` now JSON-parses the SNS HTTP notification envelope, extracts `Message`, then runs the SQS pipeline (base64 → gunzip-if-magic). Lenient fallback retained for callers who already extracted `Message` themselves. `verifyAndParseSns(notificationBody, signature, secret)` delegates to it. Tests in `test/unit/webhook-compression.test.ts` cover `unwraps a full SNS HTTP notification envelope`, a leading-whitespace variant, and rejection of a signature computed over the envelope JSON. Pure rename to `decodeSnsMessage` (Option B) was not applied — the unwrap behavior matches the spec and removes the customer footgun. Re-resolving.

nijeesh-stream:
Shipped in [`e906b9a`](https://github.com/GetStream/stream-chat-js/pull/1735/commits/e906b9af454b7821c91e71692adfb9de435a00a6) along with the SNS envelope fix.

- `WebhookSignatureError` → `InvalidWebhookError` (extends `Error`, exported from `src/signing.ts` and re-exported through `src/index.ts`)
- Every primitive now throws `InvalidWebhookError` with one of four message tokens — `signature mismatch`, `invalid base64 encoding`, `gzip decompression failed`, `invalid JSON payload` — wrapped with `cause` where a native error exists
- `InvalidWebhookErrorMessages` exported alongside the class for callers that prefer exact-match filtering over substring matches
- Legacy `StreamChat#verifyWebhook` (bool return) is untouched
- `test/unit/webhook-compression.test.ts` now 47 passed (was 42); added per-mode `toThrow(InvalidWebhookError)` + message-substring tests for all four failure modes plus `cause`-preservation tests
- `yarn lint`, `yarn types`, full `yarn vitest run` (2847 / 1 skipped) all clean

Docs side, [docs-content#1276](https://github.com/GetStream/docs-content/pull/1276) now carries the per-SDK error-class table on the Webhooks overview page.

mogita (approved):

szuperaz (approved):

PR #1737, opened 2026-05-07
fix: make predefined filters API generic and prevent dropping regular sort params if provided

MartinCupela:
## Goal

The original predefined filters implementation omitted DX, when it used `Record<string, unknown>` for filters definitions in predefined filters types. We have `ChannelsFilter` type that can be passed to predefined filters types to help users type the filters easier.

Also the `channelsQueryRequest` method was dropping regular `sort` parameter, if it detected `predefined_filters` in the methods payload. But this is incorrect in cases, when somebody wants to use predefined filter that has not sort param internally and wants to rely on dynamically provided regular `sort` object. If the server-side stored predefined filter definition contains its own sort, the regular sort param is ignored server-side.

I have added `fixme` comment so that `channelsQueryRequest` returns the whole response payload as it contains `predefined_filters` payload, that is currently being discarded and cannot be used client-side to follow sorting and filtering rules client-side. This change would be breaking and so it is not implemented with this PR.


isekovanic (approved):

src/client.ts
```diff
@@ -1929,6 +1931,9 @@ export class StreamChat {
       payload,
     );
 
+    // FIXME: In the next major release, return the full QueryChannelsAPIResponse
+    // instead of only `data.channels` so top-level metadata such as
+    // `predefined_filter` is not lost.
```
isekovanic, line 1936: That's a bummer. Good call !

PR #1742, opened 2026-05-13
chore: migrate to Yarn 4 and bump dev-deps to current majors

oliverlaz:
## CLA

- [ ] Code changes are tested

### What

- Migrate the package manager from Yarn Classic (v1) to **Yarn 4.15.0** (Berry), with the binary committed under `.yarn/releases/` so no Corepack bootstrap is required.
- Bump several dev-dep majors that had been deferred — TypeScript 5.8 → 6.0, Vitest 3.1 → 4.1 (+ `@vitest/coverage-v8`), `@commitlint/{cli,config-conventional}` 19 → 21, `conventional-changelog-conventionalcommits` 8 → 9, esbuild 0.25 → 0.28, globals 16 → 17, lint-staged 15 → 17 — with minor bumps to eslint, typescript-eslint, prettier, semantic-release, concurrently, eslint-plugin-import, and `@types/*`.
- Refresh runtime deps to latest patches within the current majors: `axios` 1.15 → 1.16, `ws` 8.18 → 8.20.
- Update CI to install via `yarn install --immutable` and cache `.yarn/cache + install-state.gz` instead of `node_modules`.
- Drop the dead `testwatch` script and the redundant `@types/base64-js` stub (base64-js ships its own types since 1.5.0).
- Add `CLAUDE.md` documenting toolchain, build pipeline, architecture, and contributor conventions for agents working in this repo.

### Why

- **Yarn 1 is EOL.** Yarn 4 brings hardened mode (HTTPS-only registries, install-script allowlisting via `dependenciesMeta`, configurable npm age gate), built-in provenance support, and the modern Berry resolver. The committed binary keeps installs deterministic across machines and CI without Corepack churn.
- The dev-dep majors had been piling up and starting to block each other. Pulling them in as one bundled bump (with the few code adjustments they require) keeps lockfile churn down and unblocks future upgrades.


.github/actions/setup-node/action.yml
```diff
@@ -20,12 +20,16 @@ runs:
...
+        path: |
+          .yarn/cache
+          .yarn/install-state.gz
+        key: ${{ runner.os }}-${{ env.NODE_VERSION }}-yarn-${{ hashFiles('yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ env.NODE_VERSION }}-yarn-
 
     - name: Install dependencies
-      run: yarn install --frozen-lockfile
+      run: yarn install --immutable
```
arnautov-anton:
```suggestion
      run: yarn install
```

We can get rid of this as [it's automatically enabled](https://yarnpkg.com/cli/install#:~:text=If%20the%20%2D%2Dimmutable%20option%20is%20set%20(defaults%20to%20true%20on%20CI)) if `CI=true` ([which it is](https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables)).

.github/actions/setup-node/action.yml
```diff
@@ -20,12 +20,16 @@ runs:
...
       uses: actions/cache@v5
       with:
-        path: ./node_modules
-        key: ${{ runner.os }}-${{ env.NODE_VERSION }}-modules-${{ hashFiles('./yarn.lock') }}
+        path: |
+          .yarn/cache
+          .yarn/install-state.gz
+        key: ${{ runner.os }}-${{ env.NODE_VERSION }}-yarn-${{ hashFiles('yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ env.NODE_VERSION }}-yarn-
```
arnautov-anton: Are we sure about this?

.github/actions/setup-node/action.yml
```diff
@@ -20,12 +20,16 @@ runs:
...
-    - name: Cache dependencies
+    - name: Cache Yarn
       uses: actions/cache@v5
       with:
-        path: ./node_modules
-        key: ${{ runner.os }}-${{ env.NODE_VERSION }}-modules-${{ hashFiles('./yarn.lock') }}
+        path: |
+          .yarn/cache
+          .yarn/install-state.gz
+        key: ${{ runner.os }}-${{ env.NODE_VERSION }}-yarn-${{ hashFiles('yarn.lock') }}
```
arnautov-anton: We might have to add a yarn version here too.

.github/actions/setup-node/action.yml
oliverlaz: perfect!

.github/actions/setup-node/action.yml
oliverlaz: yeah, I forgot to change to `setup/node{cache}` before pushing. Now it uses the default caching config coming from GH.

.github/actions/setup-node/action.yml
oliverlaz: It should be handled by the setup-node action now

MartinCupela (approved):

PR #1746, opened 2026-05-14
feat(MessageComposer): control command sendability

MartinCupela:
## Goal

Closes REACT-971

Add API that allows to verify, whether the message composer in command mode has state ready to be composed and sent to the server.

This are the rules:

- **ban** - requires mention and additional text in order the `MessageComposer.isCommandSendable` returns `true`
- **unban** - requires mention in order the `MessageComposer.isCommandSendable` returns `true`
- **mute** - requires mention in order the `MessageComposer.isCommandSendable` returns `true`
- **unmute** - requires mention in order the `MessageComposer.isCommandSendable` returns `true`

It is possible to provide custom command sendability validation function(s) in case the integrators have custom commands and want to apply custom validation rules.

**Mentioned users state kept up-to-date on text change**
Make sure that the mentioned users array in text composer state is not stale. Until now, this was verified only at the composition time. Now we prune stale user mentions on every text change middleware execution.

### DIFF
--- a/src/messageComposer/configuration/commands.configuration.ts
+++ b/src/messageComposer/configuration/commands.configuration.ts
@@ -6,7 +6,7 @@ import type {
 import type { DeepPartial } from '../../types.utility';
 import { stripMentionTokens } from '../middleware';
 
-const MENTION_ONLY_COMMANDS = new Set(['mute', 'unmute', 'unban']);
+export const MENTION_ONLY_COMMANDS = new Set(['mute', 'unmute', 'unban']);
 export const defaultCommandSendabilityValidator: CommandSendValidator = ({
   command,
   commandArgsText,
--- a/src/messageComposer/configuration/commands.configuration.ts
+++ b/src/messageComposer/configuration/commands.configuration.ts
@@ -31,17 +31,17 @@ export const defaultCommandSendabilityValidator: CommandSendValidator = ({
   return { command, ready: true };
 };
 export const DEFAULT_COMMANDS_CONFIG: CommandsConfig = {
-  sendValidators: [defaultCommandSendabilityValidator],
+  sendValidator: defaultCommandSendabilityValidator,
 };
 export const applyCommandValidatorOverride = (
   targetConfig: MessageComposerConfig,
   sourceConfig?: DeepPartial<MessageComposerConfig>,
 ) => {
-  const overrideValidators = sourceConfig?.commands?.sendValidators as
-    | CommandSendValidator[]
+  const overrideValidator = sourceConfig?.commands?.sendValidator as
+    | CommandSendValidator
     | undefined;
 
-  if (typeof overrideValidators === 'undefined') {
+  if (typeof overrideValidator === 'undefined') {
     return targetConfig;
   }
 
@@ -49,7 +49,7 @@ export const applyCommandValidatorOverride = (
     ...targetConfig,
     commands: {
       ...targetConfig.commands,
-      sendValidators: overrideValidators,
+      sendValidator: overrideValidator,
     },
   };
 };
--- a/src/messageComposer/configuration/commands.configuration.ts
+++ b/src/messageComposer/configuration/commands.configuration.ts
@@ -15,17 +15,17 @@ export const defaultCommandSendabilityValidator: CommandSendValidator = ({
   if (command.name !== 'ban' && !MENTION_ONLY_COMMANDS.has(command.name ?? '')) return;
 
   if (mentionedUsersInText.length === 0) {
-    return { command, ready: false, reason: 'missing_mention' };
+    return { command, ready: false, reason: 'missing-mention' };
   }
 
   if (command.name !== 'ban') {
     return { command, ready: true };
   }
 
-  const reason = stripMentionTokens(commandArgsText, mentionedUsersInText);
+  const banReason = stripMentionTokens(commandArgsText, mentionedUsersInText);
 
-  if (!reason.length) {
-    return { command, ready: false, reason: 'missing_reason' };
+  if (!banReason.length) {
+    return { command, ready: false, reason: 'missing-ban-reason' };
   }
 
   return { command, ready: true };
--- a/src/messageComposer/configuration/commands.configuration.ts
+++ b/src/messageComposer/configuration/commands.configuration.ts
@@ -1,0 +1,55 @@
+import type {
+  CommandsConfig,
+  CommandSendValidator,
+  MessageComposerConfig,
+} from './types';
+import type { DeepPartial } from '../../types.utility';
+import { stripMentionTokens } from '../middleware';
+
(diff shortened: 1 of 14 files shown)


src/messageComposer/configuration/commands.configuration.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (mentionedUsersInText.length === 0) {
+    return { command, ready: false, reason: 'missing_mention' };
+  }
+
+  if (command.name !== 'ban') {
+    return { command, ready: true };
+  }
+
+  const reason = stripMentionTokens(commandArgsText, mentionedUsersInText);
```
isekovanic: I might be missing something here, but how would this generate a good `reason` for why the `command` is currently not sendable ?

src/messageComposer/configuration/commands.configuration.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!reason.length) {
+    return { command, ready: false, reason: 'missing_reason' };
+  }
+
+  return { command, ready: true };
+};
+export const DEFAULT_COMMANDS_CONFIG: CommandsConfig = {
+  sendValidators: [defaultCommandSendabilityValidator],
```
isekovanic: Why do we keep these as an array ? Since the validation step is a middleware anyway, integrators can either add their own additional validation steps (or replace our current one)

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -11,21 +16,9 @@ import type {
...
-  const normalizedCommandName = commandName.toLowerCase();
-  return searchSource
-    .query(normalizedCommandName)
-    .items.find((command) => command.name?.toLowerCase() === normalizedCommandName);
-};
-
 const getDisabledRawCommand = (
   composer: MessageComposer,
-  searchSource: CommandSearchSource,
+  searchSource: Pick<CommandSearchSource, 'query'>,
```
isekovanic: Why not just use `CommandSearchSource` here ? It should still comply

src/messageComposer/middleware/messageComposer/compositionValidation.ts
```diff
@@ -52,14 +47,28 @@ export const createCompositionValidationMiddleware = (
...
         );
         if (disabledRawCommand) {
           notifyCommandDisabled(composer, disabledRawCommand);
           return await discard();
         }
 
+        const currentCommand =
+          composer.textComposer.command ??
+          getCommandByName(effectiveCommandSearchSource, getRawCommandName(inputText));
+        if (currentCommand) {
```
isekovanic:
```suggestion
        if (currentCommand && notifyCommandNotReady({
              composer,
              sendability: composer.validateCommandSendability(currentCommand, inputText),
            })) {
```

src/messageComposer/messageComposer.ts
```diff
@@ -378,21 +402,49 @@ export class MessageComposer extends WithSubscriptions {
...
 
   get hasSendableData() {
-    return !!(
-      (!this.attachmentManager.uploadsInProgressCount &&
-        (!this.textComposer.textIsEmpty ||
-          this.attachmentManager.successfulUploadsCount > 0)) ||
-      this.pollId ||
-      !!this.locationComposer.validLocation
+    const currentCommand = this.textComposer.command;
+    const commandIsSendable = !currentCommand || this.isCommandSendable(currentCommand);
```
isekovanic: Would it be possible to extract this in a separate getter please ?

So something like:

```
get isCommandSendable() {
  const currentCommand = this.textComposer.command;
  return !currentCommand || this.isCommandSendable(currentCommand);
}
```

src/messageComposer/configuration/commands.configuration.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+export const applyCommandValidatorOverride = (
```
isekovanic, line 36: Since the command validation anyway goes through a middleware, this might not be needed I think. While convenient for integrators, I think convergence towards a middleware approach would be more appropriate (it would separate the need for overrides as well), unless I'm missing something important here

isekovanic (changes requested):

src/messageComposer/configuration/commands.configuration.ts
MartinCupela: This is ban reason - I will rename it.

src/messageComposer/middleware/messageComposer/compositionValidation.ts
MartinCupela: For flexibility?

src/messageComposer/middleware/messageComposer/compositionValidation.ts
MartinCupela: refactored in: 6c1c720e2080a0db3b92804b2a22d80422c34995

src/messageComposer/configuration/commands.configuration.ts
MartinCupela: Renamed in 6c1c720e2080a0db3b92804b2a22d80422c34995

src/messageComposer/messageComposer.ts
MartinCupela: Done in 8f1fb8a94f94829c60b19e8fe34c4be99e5a7c73

src/messageComposer/middleware/messageComposer/compositionValidation.ts
isekovanic: I meant more like, `CommandSearchSource` also has `query` - with this we're specifically looking for `query` from the entire `searchSource` instead of exposing it fully. If we need something else in the future we'll have to change this again.

src/messageComposer/configuration/commands.configuration.ts
MartinCupela: Addressed in 6c096ad3d9780f96558017dd8dea962f73384f43

src/messageComposer/configuration/commands.configuration.ts
MartinCupela: It goes through the middleware but also it is queried when the message composer state changes before performing the composition in the middleware. We want to keep the send button disabled (pre-composition state) until the composer state in command mode is valid. There we do not have any middleware.

isekovanic: Makes sense !

src/messageComposer/middleware/messageComposer/compositionValidation.ts
MartinCupela: Ok, adjusted here to using CommandSearchSource type: 9b45251b6188aa74dc8bf741e6cb5920e30d50cc

isekovanic (approved):

PR #1748, opened 2026-05-15
fix(poll): preserve cleared option identity without collapsing empty rows

MartinCupela:
## Goal

Preserve the cleared option instance instead of replacing it with a later empty placeholder row, so clearing text remains a mutation of the same logical option rather than an implicit remove-and-replace. This prevents re-mounts and loss of focus in the UI. Changing the option id has potentially more unknown side effects.

### DIFF
--- a/src/messageComposer/middleware/pollComposer/state.ts
+++ b/src/messageComposer/middleware/pollComposer/state.ts
@@ -137,6 +137,27 @@ export const pollCompositionStateProcessors: Partial<
     // For single option updates
     const { index, text } = value;
     const prevOptions = data.options || [];
+    const targetOption = prevOptions[index];
+
+    if (!targetOption) {
+      return { options: prevOptions };
+    }
+
+    const nextOptionsAfterTarget = prevOptions.slice(index + 1);
+    const shouldPreserveClearedOption =
+      !text &&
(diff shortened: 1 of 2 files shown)


src/messageComposer/middleware/pollComposer/state.ts
```diff
@@ -137,6 +137,27 @@ export const pollCompositionStateProcessors: Partial<
...
+
+    if (!targetOption) {
+      return { options: prevOptions };
+    }
+
+    const nextOptionsAfterTarget = prevOptions.slice(index + 1);
+    const shouldPreserveClearedOption =
+      !text &&
+      nextOptionsAfterTarget.length > 0 &&
+      nextOptionsAfterTarget.every((option) => !option.text);
```
szuperaz, line 150: I'm just wondering if we need this condition here - to me it seems that if I have 5 options, clearing the text on the 3rd one, that option will still be removed. UX-wise it seems more logical to me to only delete this option when the remove button is clicked, and only automatically remove if the last option's text was cleared. In any case, the PR looks good to me, so feel free to dismiss the suggestion if it doesn't seem reasonable to you.

szuperaz (approved):

PR #1760, opened 2026-05-29
feat: integrate OpenAPI generated clients

arnautov-anton:


src/messageComposer/attachmentManager.ts
```diff
@@ -448,8 +450,12 @@ export class AttachmentManager {
       localAttachment.thumb_url = fileLike.thumb_url;
     }
 
-    if (isFileReference(fileLike) && fileLike.duration) {
-      localAttachment.duration = fileLike.duration;
+    if (
+      isFileReference(fileLike) &&
+      fileLike.duration &&
+      localAttachment.type === 'voiceRecording'
```
arnautov-anton, line 464: 👀

arnautov-anton: From my personal testing, `duration` and `waveform_data` appeared only on attachments of type `voiceRecording` the rest seem to not have been using these properties so I clarified it here to make TS happy.

src/gen/models/index.ts
```diff
@@ -0,0 +1,12662 @@
...
+  next?: string;
+
+  prev?: string;
+}
+
+export interface QueryBannedUsersPayload {
+  /**
+   * Filter conditions to apply to the query
+   */
+  filter_conditions: Filters<{
```
arnautov-anton, line 7511: Filters are now part of the generated files.

src/gen/models/index.ts
```diff
@@ -0,0 +1,12662 @@
...
+
+  /**
+   * User ID
+   */
+  user_id: string;
+
+  /**
+   * Custom data for this object
+   */
+  custom: CustomReactionData;
```
arnautov-anton, line 9029: `custom` fields are populated with a different, JS-based script. I wasn's sure how to integrate this within the OAPI generator - I'm open to suggestions though.

src/logger.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @internal
+ */
+export type ScopedLogger = scopedLogger.Logger<ChatLoggerScope>;
+
+export { LogLevelEnum } from '@stream-io/logger';
+export type { ConfigureLoggersOptions, LogLevel, Sink } from '@stream-io/logger';
+
+export const chatLoggerSystem = scopedLogger.createLoggerSystem<ChatLoggerScope>();
```
arnautov-anton, line 28: The project now uses our logging system, no more `this.logger`.

src/poll.ts
```diff
@@ -74,16 +46,16 @@ export type PollOptionVotesQueryParams = {
...
 type OptionId = string;
 
-export type PollState = Omit<PollResponse, 'own_votes' | 'id'> & {
+export type PollState = Omit<PollResponse_old, 'own_votes' | 'id'> & {
   lastActivityAt: Date; // todo: would be ideal to get this from the BE
   maxVotedOptionIds: OptionId[];
-  ownVotesByOptionId: Record<OptionId, PollVote>;
-  ownAnswer?: PollAnswer; // each user can have only one answer
+  ownVotesByOptionId: Record<OptionId, PollVoteResponseData>;
+  ownAnswer?: PollVoteResponseData; // each user can have only one answer
```
arnautov-anton, line 53: @MartinCupela - I've tried porting poll-related types to the generated ones, but there were some inconsitencies and that's why `PollResponse_old` has been kept. We should make sure to drop it before official v10 release though.

src/utils.ts
```diff
@@ -131,29 +130,12 @@ export function addFileToFormData(
...
-      console.warn(
-        "client._buildSort() - multiple fields in a single sort object detected. Object's field order is not guaranteed",
-      );
-    }
-    for (const [field, direction] of entries) {
-      sortFields.push({ field, direction });
-    }
-  }
-  return sortFields;
-}
```
arnautov-anton, line 173: Sort normalization has been dropped, we no longer accept the object version or even short array version:
```ts
{ pinned_at: -1, archived_at: 1 }
// or
[{ pinned_at: -1 }, { archived_at: 1 }]
```
The new version should look like this:
```ts
[{ field: "pinned_at", direction: -1 }, { field: "archived_at", direction: 1 }]
```
This version also allows to pass `type` property which allows casting of a value to the specified type before sorting.

src/types.ts
```diff
@@ -2203,246 +629,38 @@ export type QueryFilter<ObjectType = string> =
...
-  reply_count?: AscDesc;
-  text?: AscDesc;
-  type?: AscDesc;
-  updated_at?: AscDesc;
-  'user.id'?: AscDesc;
-};
+export type ReactionSort = SortParamRequest[];
 
-export type SearchMessageSort = SearchMessageSortBase | Array<SearchMessageSortBase>;
+export type ChannelSort = SortParamRequest[];
```
arnautov-anton, line 642: All of these are the same and are fixed types without key suggestion - we should revisit this and generate this in a similar manner we generate filters.

src/channel_state.ts
```diff
@@ -228,48 +229,49 @@ export class ChannelState {
...
-        ) {
-          this.last_message_at = new Date(message.created_at.getTime());
-        }
+      if (
+        !shouldSkipLastMessageAtUpdate &&
+        (!this.last_message_at ||
+          message.created_at.getTime() > this.last_message_at.getTime())
+      ) {
+        this.last_message_at = new Date(message.created_at.getTime());
       }
```
arnautov-anton, line 275: This is the biggest change I can think of - first, notice the bracked placement after `message = this.formatMessage(messagesToAdd[i]);` and second message has been considered a `LocalMessage` if it had `created_at` converted to a `Date` instance. Newly - message is considered local if it has been given `status` field which is not being delivered as part of the response. Some of the SDK code relies on this field to be present in the messages received through the WS events - I'm not sure how that worked previously, I'll mark it in the `stream-chat-react` for further discussion.

src/messageComposer/attachmentManager.ts
isekovanic: Can they not also appear in audio attachments in general ? Like an audio file specifically

src/poll.ts
isekovanic: Hmm what were the inconsistencies ? Has the `polls` API changed ?

src/messageComposer/middleware/messageComposer/cleanData.ts
```diff
@@ -39,7 +39,7 @@ export const createCompositionDataCleanupMiddleware = (
           ...editedMessagePayloadToBeSent,
           ...state.message,
           ...common,
-        },
+        } as typeof state.message,
```
isekovanic, line 42: shouldn't this already be correct ? why do we need the type query ?

src/client.ts
```diff
@@ -374,22 +178,43 @@ export class StreamChat {
...
-  userID?: string;
   wsBaseURL?: string;
   wsConnection: StableWSConnection | null;
   wsFallback?: WSConnectionFallback;
   wsPromise: ConnectAPIResponse | null;
-  consecutiveFailures: number;
+  get anonymous(): boolean {
+    return this.user?.anon ?? false;
+  }
+  get userId() {
```
isekovanic, line 197: While I welcome the switch, I have 2 questions:

1. Since `client.userID` is used very, very heavily pretty much everywhere - does it not mean it will make the migration have significantly more surface. Are we sure we want this ? I know it's a breaking change but just thinking about the migration itself.
2. Are we certain `this.user?.id` is always going to be correct ? Is this how we inferred `client.userID` as well in the past ? I genuinely can't remember, asking because of the incredibly large surface it covers

src/client.ts
```diff
@@ -374,22 +178,43 @@ export class StreamChat {
...
+    return this.user?.id;
+  }
+  /**
+   * @deprecated Use `userId` instead.
+   */
+  get userID() {
```
isekovanic, line 203: Probably just ditch these  if we're switching to the (much better) camel case

src/client.ts
```diff
@@ -446,60 +271,37 @@ export class StreamChat {
...
-    }
-
-    this.axiosInstance = axios.create(this.options);
+    this.axiosInstance = axios.create({
+      timeout: 3000,
+      withCredentials: false,
+      httpsAgent: this.node
+        ? new https.Agent({ keepAlive: true, keepAliveMsecs: 3000 })
+        : undefined,
+      ...this.options.axiosRequestConfig,
```
isekovanic, line 292: does the full request config go into `axiosRequestConfig` now ? because on RN we also have an adapter for our HTTP client as we need it to provide native multipart uploads, it looks like this is the case but just wanna double check

src/channel_state.ts
isekovanic: This will all be out of scope on V10 very soon (perhaps it might need to be ported to `messagePaginator` though ?)

src/messageComposer/middleware/messageComposer/cleanData.ts
arnautov-anton: There are some inconsistencies like this one - I tried to reduce them as much as possible but some prevailed because without the assertion they kept throwing errors. Did not dig deeper as to why, that's for part two I suppose.

src/client.ts
arnautov-anton: I'd be worried if v10 was still supporting server-side functionality - that's where I'd expect `userID` to be problematic but since the new version supports client-side only and user is required for all BE interactions then I don't think this should be a problem whatsoever. User should be set during connection, updated through events and cleaned up on disconnects.

arnautov-anton: I may be wrong but I'd rather keep it simple going forward. Given the possibility of AI-accelerated migrations I feel like it should not cause troubles. I have not noticed any hurdles during `stream-chat-react` migration anyway.

src/client.ts
arnautov-anton: Yes - I did not like that Axios-specific settings were both part of the `StreamChatOptions` (spread) and as a separate key in `axiosRequestConfig` - it was a mess. Now everything Axios-related lives in that particular property.

arnautov-anton: Also see: https://github.com/GetStream/stream-chat-js/blob/8becc277a8e8ec604095d7637b7df3a0a7411c2f/src/api-client.ts#L126-L157

src/poll.ts
arnautov-anton: Ooops, forgot to reply to this one. I think it all revolves around `PollEnrichData` which are not present in the OAPI spec, I suppose those are either custom or local data of sorts, not sure.

src/channel_state.ts
MartinCupela: This is already dead on my branch :)

src/poll.ts
isekovanic: Aaaah yes you did mention that on the OAPI presentation, right 👍 

It was neither, but it was like a intermediary type that was added to polls from the BE team if I remember correctly (@MartinCupela can correct me if I'm wrong here). I guess all of the things are now merged into the generic poll type so it should be alright if that's indeed the case

MartinCupela (approved):

src/client.ts
isekovanic: Great, so long as we pass everything still I'm good with it 🤝 Thank you !

isekovanic (approved):

src/messageComposer/attachmentManager.ts
isekovanic: Handled offline, `waveform` and `duration` are custom fields in fact - so we can't have any surprises here specifically. I retract this :D

PR #1761, opened 2026-06-02
fix: increasing/decreasing member count on member event cause incorrect member count

szuperaz:
The issue:
When adding more than one member at once to a channel, this is the WS event sequence I get:
- Channel has 4 members now, we're adding 2 new members in a single API call
- `member.added` event arrives -> stream-chat-js increases member count by one, current value is 5
- `channel.updated` event arrives with member count 6 -> stream-chat-js sets member count from WS event to 6
- `member.added` event arrives -> stream-chat-js increases member count by one, current value is 7 ❌

The solution:
Since `channel.updated` is dispatched every time we add/remove members, it's safe to set member count only from that event, and don't increase/decrease counter on member events.

## CLA

- [ ] Code changes are tested

## Changelog

-

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -2165,9 +2165,6 @@ export class Channel {
             ...channelState.members,
             [memberCopy.user.id]: memberCopy,
           };
-          if (channel.data?.member_count && event.type === 'member.added') {
-            channel.data.member_count += 1;
-          }
         }
 
         const currentUserId = this.getClient().userID;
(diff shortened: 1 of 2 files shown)


MartinCupela (changes requested):
@szuperaz could you add tests for this, please?

szuperaz:
> @szuperaz could you add tests for this, please?

Do you mean a negative test checking we don't change counter on member events?

szuperaz:
> @szuperaz could you add tests for this, please?

Added tests in this commit: https://github.com/GetStream/stream-chat-js/pull/1761/commits/f1db7dc8c1226c28fe7398893251e45bf815aa4d

MartinCupela (approved):

PR #1770, opened 2026-06-12
feat(search): add ChannelMemberSearchSource

MartinCupela:
## Goal

Add ChannelMemberSearchSource that can be used to paginate and search for channel members in channel management UIs.

Also removes condition for existence of searchQuery in MessageSearchSource.

### DIFF
--- a/src/search/ChannelMemberSearchSource.ts
+++ b/src/search/ChannelMemberSearchSource.ts
@@ -1,0 +1,85 @@
+import { BaseSearchSource } from './BaseSearchSource';
+import { FilterBuilder, type FilterBuilderOptions } from '../pagination';
+import type { Channel } from '../channel';
+import type {
+  ChannelMemberResponse,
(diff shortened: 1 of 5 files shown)


src/search/ChannelMemberSearchSource.ts
```diff
@@ -0,0 +1,85 @@
...
+    >({
+      initialFilterConfig: {
+        default: {
+          enabled: true,
+          generate: ({ searchQuery }) =>
+            searchQuery
+              ? {
+                  $or: [
+                    { name: { $autocomplete: searchQuery } },
+                    { id: { $eq: searchQuery } },
```
szuperaz, line 53: I don't really see the benefit of having `id` in the default search config, but I guess since it's not `$autocomplete` but `$eq` it doesn't hurt either, so feel free to leave it in

szuperaz (approved):

PR #1771, opened 2026-06-12
fix: include fixed queue cache in client

### LINKED ISSUE #3642: [🐛] 🔥 Module-level MessageComposer cache in useCreateMessageComposer leaks composers across connectUser cycles — "You can't use a channel after client.disconnect() was called" after switching users
## Issue

`useCreateMessageComposer` (`src/contexts/messageInputContext/hooks/useCreateMessageComposer.ts`) keeps its composer cache at **module scope**:

```js
const queueCache = new FixedSizeQueueCache(64);
```

`MessageComposer` instances for threads (`legacy_thread_<id>`) and edited messages (`message_<id>`) are stored in this cache and survive for the lifetime of the JS process — across `connectUser`/`disconnectUser` cycles and across `StreamChat` client instances. Each cached composer holds a `channel` object bound to the client/session that existed when it was created.

After a logout → login-as-different-user flow, the hook returns a cached composer whose channel object was torn down by the previous `disconnectUser()`. Mounting it then throws:

1. The hook's effect runs `messageComposer.registerSubscriptions()`
2. → the `subscribe*` helpers access `this.client`
3. → `MessageComposer`'s getter: `get client() { return this.channel.getClient() }` (`stream-chat` `src/messageComposer/messageComposer.ts`, ~line 344)
4. → `Channel.getClient()` throws `You can't use a channel after client.disconnect() was called` (`src/channel.ts`, ~line 185)

The throw happens inside a React passive-effect mount, so it propagates to the app's root error boundary and takes down the whole screen.

Captured runtime stack from a device repro (Hermes, frames joined into one line):

```
getClient  <-  get [MessageComposer.client]  <-  [subscribe*]  <-  registerSubscriptions  <-  [useCreateMessageComposer useEffect]  <-  react_stack_bottom_frame  <-  runWithFiberInDEV  <-  commitHookEffectListMount  <-  commitHookPassiveMountEffects  <-  commitPassiveMountOnFiber  <-  recursivelyTraversePassiveMountEffects  ...
```

Two points worth highlighting:

- **It reproduces regardless of how the next session's client is obtained.** `disconnectUser()` permanently marks the existing `Channel` **objects** `disconnected = true`; reconnecting creates fresh channel objects and never resurrects the old ones. We hit the crash both with `StreamChat.getInstance()` (the singleton reconnected as user B, but the cached composer still pointed at a pre-logout channel object) and with a fresh `new StreamChat()` per login (which is how we conclusively attributed the leaked channel to the previous session, by tagging client instances with serial numbers).
- **The cache is module-scoped and not exported, so there is no way for an application to clear or invalidate it on logout.** We're aware of the guidance in #3394 that stale channel objects held in _application_ state must be refreshed by the app on user switch — but this cache is SDK-internal state, so the app has no way to follow that guidance here.

### Steps to reproduce

1. Connect user A and render a `Channel` with a `thread` (or start editing a message) — this creates a `MessageComposer` with a `legacy_thread`/`message` context and stores it in the module-level `queueCache`.
2. Log user A out: `client.disconnectUser()` (marks the composer's channel object `disconnected`).
3. Connect as user B — either by reconnecting the same `getInstance()` singleton or via a new `StreamChat` client; both reproduce.
4. Navigate so that `useCreateMessageComposer` resolves the same tag again (open the same thread; in our app it was also reached on the channel screen during composer resolution for the same channel/thread).
5. The cached composer from step 1 is returned; its `registerSubscriptions()` effect throws `You can't use a channel after client.disconnect() was called` and the screen crashes.

Minimal sketch:

```ts
const clientA = new StreamChat(API_KEY);
await clientA.connectUser(userA, tokenA);
// render <Channel channel={chA} thread={parentMessage}> ... -> composer cached under legacy_thread_<id>

await clientA.disconnectUser();

// same result with `StreamChat.getInstance(API_KEY)` reconnected as user B
const clientB = new StreamChat(API_KEY);
await clientB.connectUser(userB, tokenB);
// render the same thread again -> useCreateMessageComposer returns the cached
// composer bound to a pre-logout (disconnected) channel object -> registerSubscriptions() throws
```

### Expected behavior

Switching users (any `disconnectUser()` followed by a `connectUser`, on the same singleton or a new client) must not resurrect composers bound to disconnected channel objects. The cached composer should either be invalidated when its channel is disconnected, or the cache should be scoped to the client.

### Customization

<details><summary>Click To Expand</summary>
<p>

The crash is independent of our UI customizations (it reproduces through stock `Channel`/`Thread`/`MessageInput` rendering); the relevant usage is the connection lifecycle and a thread render:

```typescript jsx
// Session 1 (user A): standard thread screen
<Chat client={client}>
  <Channel channel={channel} thread={parentMessage} threadList>
    <Thread />
  </Channel>
</Chat>

// Logout: await client.disconnectUser()
// Session 2 (user B): connectUser (same getInstance() singleton OR a fresh client)
// Re-rendering the channel/thread resolves the cached composer from session 1 and crashes.
```

</p>
</details>

### Offline support

Offline support is **not** enabled in our app.

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "expo": "55.0.17",
  "expo-router": "55.0.13",
  "react": "19.2.2",
  "react-native": "0.83.6",
  "stream-chat": "9.43.0",
  "stream-chat-expo": "9.1.2"
}
```

**`react-native info` output:**

Using expo, react-native cli not installed.

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [x] Both
- **`stream-chat-react-native` version you're using that has this issue:**
  - `stream-chat-expo 9.1.2` (`stream-chat-react-native-core 9.1.2`), with `stream-chat 9.43.0`
- Device/Emulator info:
  - [x] I am using a physical device
  - OS version: `reproduced on both iOS and Android (physical devices via TestFlight/internal builds, and dev clients)`
  - Device/Emulator: `iPhone + Android phone`

</p>
</details>

### Additional context

Workaround we're currently shipping — patching `FixedSizeQueueCache.prototype.get/peek` so values bound to a disconnected channel act as cache misses (the hook then rebuilds the composer against the live client, and `add()` replaces the stale entry under the same tag):

```ts
import { FixedSizeQueueCache } from "stream-chat";

const proto = FixedSizeQueueCache.prototype as any;
for (const method of ["get", "peek"]) {
  const orig = proto[method];
  proto[method] = function (key: unknown) {
    const value = orig.call(this, key);
    if (value?.channel?.disconnected) return undefined;
    return value;
  };
}
```

This resolves the crash for us, but it's a monkey-patch we'd love to delete. Possible upstream fixes:

- Scope the cache to the client (include the client/connection identity in the cache key, or store the cache on the client instance), or
- Validate cache hits: treat entries whose `channel.disconnected === true` (or whose client !== the current `useChatContext().client`) as misses and rebuild the composer, or
- Expose a way to clear the cache (e.g., on `disconnectUser`).

Related: #3394 (same error after a user switch; closed with guidance that stale channel objects in app state must be refreshed — this report is about the SDK-internal cache that the app cannot refresh). Happy to provide the full instrumented logs (per-client serial numbers showing the cached composer's channel belonging to the previous, disconnected session) if useful.

### Screenshots

Image

isekovanic:
## CLA

- [ ] Code changes are tested

Move the `FixedSizeQueueCache<string, MessageComposer>` that backs thread/edit composer reuse from module scope in the UI SDKs onto the `StreamChat` client instance and drain it on `disconnectUser()`.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
The React and RN UI SDKs currently keep this cache at module scope inside their respective `useMessageComposerController`/`useCreateMessageComposer` hooks:                                                                                                                        
                                                                                                                                                                                                                                                                                                                          
```ts                                                                                                                                                                                                                                                                                                                     
const queueCache = new FixedSizeQueueCache<string, MessageComposer>(64); 
```                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                          
Because the cache is module scoped and not exported, it survives for the lifetime of the JS process across `connectUser`/`disconnectUser` cycles and across `StreamChat` instances. Each cached composer holds a `Channel` object bound to the session that existed when it was created.
                                                                                                                                                                                                                                                                                                                          
After a **logout** then **login** flow (whether via reconnecting the same `StreamChat.getInstance()` singleton or by constructing a bran new `new StreamChat()`), the hook resolves the same tag (`legacy_thread_<id>`/`message_<id>`) and returns a composer whose channel was torn down by the previous `disconnectUser()` (and is marked as `channel.disconnected`. Mounting it then throws, because:                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                          
- `useEffect` runs `messageComposer.registerSubscriptions()`
- `subscribe*` helpers read `this.client`                                                                                                                                                                                                                                                                                   
- `get client() { return this.channel.getClient() }`
- `Channel.getClient()` throws with `"You can't use a channel after client.disconnect() was called"`
                                                                                                                                                                                                                                                                                                                          
The throw happens inside a React passive effect mount, so it propagates to the root error boundary and takes down the screen. Since the cache is module scoped and not exported, applications have no way to invalidate it on user switch and they cannot follow the existing "refresh stale channel objects on user switch" guidance. Nor should they need to as this should all be handled internally.                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
To tackle this, we do the following:
                                                                                                          
1. `StreamChat.messageComposerFixedSizeQueue` - instantiated in the constructor at capacity `64` as it was before. Owned per client instance, so it can never be reused across instances. Also solves the multiple `client` instances issue.
2. `FixedSizeQueueCache.clear()` - runs the optional `dispose` callback for each entry if it, then resets `map` and `keys`. Lets consumers register subscription teardown via dispose if they want explicit lifecycle cleanup at disconnect. Currently we don't expose a clear way to do this, but since we didn't have that before anyway I'm keeping it that way in this iteration and we can possibly extend later on.                                                                                            
3. `disconnectUser()` - calls `this.messageComposerFixedSizeQueue.clear()` and makes sure the cache is flushed entirely so that no `MessageComposer` instances get to leak.
                                                                                                                                                                                                                                                                                                                          
After ths, both reproduction paths are covered:                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                          
- Singleton reconnect - `clear()` drains the cache at disconnect and the next mount does a cache miss and rebuilds against the fresh channel.                                                                                                                                                                                       
- New `StreamChat()` per login - the new instance has its own empty cache and the old instance's cache is unreachable from the new client even if it lives on in memory.                                                              
                                                                                                                                                                                                                                                                                                                          
Will resolve [this GH issue](https://github.com/GetStream/stream-chat-react-native/issues/3642).

## Changelog

-

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -293,6 +293,7 @@ import { StateStore } from './store';
 import type { MessageComposer } from './messageComposer';
 import type { AbstractOfflineDB } from './offline-support';
 import { getPendingTaskChannelData } from './offline-support/util';
+import { FixedSizeQueueCache } from './utils/FixedSizeQueueCache';
 
 function isString(x: unknown): x is string {
   return typeof x === 'string' || x instanceof String;
@@ -393,6 +394,7 @@ export class StreamChat {
   defaultWSTimeout: number;
   sdkIdentifier?: SdkIdentifier;
   deviceIdentifier?: DeviceIdentifier;
+  messageComposerFixedSizeQueue: FixedSizeQueueCache<string, MessageComposer>;
   private nextRequestAbortController: AbortController | null = null;
   /**
    * @private
@@ -573,6 +575,9 @@ export class StreamChat {
     this.polls = new PollManager({ client: this });
     this.reminders = new ReminderManager({ client: this });
     this.messageDeliveryReporter = new MessageDeliveryReporter({ client: this });
+    this.messageComposerFixedSizeQueue = new FixedSizeQueueCache<string, MessageComposer>(
+      64,
+    );
   }
 
   /**
@@ -1060,6 +1065,7 @@ export class StreamChat {
     // reset thread manager
     this.threads.resetState();
     this.uploadManager.reset();
+    this.messageComposerFixedSizeQueue.clear();
 
     // Since we wipe all user data already, we should reset token manager as well
     closePromise
--- a/src/client.ts
+++ b/src/client.ts
@@ -394,7 +394,7 @@ export class StreamChat {
   defaultWSTimeout: number;
   sdkIdentifier?: SdkIdentifier;
   deviceIdentifier?: DeviceIdentifier;
-  messageComposerFixedSizeQueue: FixedSizeQueueCache<string, MessageComposer>;
+  messageComposerCache: FixedSizeQueueCache<string, MessageComposer>;
   private nextRequestAbortController: AbortController | null = null;
   /**
    * @private
@@ -575,9 +575,7 @@ export class StreamChat {
     this.polls = new PollManager({ client: this });
     this.reminders = new ReminderManager({ client: this });
     this.messageDeliveryReporter = new MessageDeliveryReporter({ client: this });
-    this.messageComposerFixedSizeQueue = new FixedSizeQueueCache<string, MessageComposer>(
-      64,
-    );
+    this.messageComposerCache = new FixedSizeQueueCache<string, MessageComposer>(64);
   }
 
   /**
@@ -1065,7 +1063,7 @@ export class StreamChat {
     // reset thread manager
     this.threads.resetState();
     this.uploadManager.reset();
-    this.messageComposerFixedSizeQueue.clear();
+    this.messageComposerCache.clear();
 
     // Since we wipe all user data already, we should reset token manager as well
     closePromise
--- a/src/client.ts
+++ b/src/client.ts
@@ -394,7 +394,7 @@ export class StreamChat {
   defaultWSTimeout: number;
   sdkIdentifier?: SdkIdentifier;
   deviceIdentifier?: DeviceIdentifier;
-  messageComposerCache: FixedSizeQueueCache<string, MessageComposer>;
+  readonly messageComposerCache: FixedSizeQueueCache<string, MessageComposer>;
   private nextRequestAbortController: AbortController | null = null;
   /**
    * @private
--- a/src/utils/FixedSizeQueueCache.ts
+++ b/src/utils/FixedSizeQueueCache.ts
@@ -77,9 +77,7 @@ export class FixedSizeQueueCache<K, T> {
    */
   clear() {
     if (this.dispose) {
-      for (const [key, entry] of this.map.entries()) {
-        this.dispose(key, entry);
-      }
+      this.map.forEach((entry, key) => this.dispose?.(key, entry));
     }
     this.map.clear();
     this.keys = [];
--- a/src/utils/FixedSizeQueueCache.ts
+++ b/src/utils/FixedSizeQueueCache.ts
@@ -71,4 +71,17 @@ export class FixedSizeQueueCache<K, T> {
 
     return foundItem;
   }
+
+  /**
+   * Clears queue entirely and disposes of each item individually.
+   */
+  clear() {
+    if (this.dispose) {
+      for (const [key, entry] of this.map.entries()) {
+        this.dispose(key, entry);
+      }
+    }
+    this.map.clear();
+    this.keys = [];
+  }
 }
--- a/test/unit/client.test.js
+++ b/test/unit/client.test.js
@@ -368,16 +368,16 @@ describe('Client disconnectUser', () => {
 
 	it('should clear the message composer cache', async () => {
 		const client = new StreamChat('', '');
-		client.messageComposerFixedSizeQueue.add('cid:a', {});
-		client.messageComposerFixedSizeQueue.add('cid:b', {});
+		client.messageComposerCache.add('cid:a', {});
+		client.messageComposerCache.add('cid:b', {});
 
 		const { resolve, promise } = Promise.withResolvers();
 		client.wsConnection = { disconnect: () => promise };
 		client.wsFallback = null;
 		const disconnectPromise = client.disconnectUser();
 
-		expect(client.messageComposerFixedSizeQueue.peek('cid:a')).to.be.undefined;
-		expect(client.messageComposerFixedSizeQueue.peek('cid:b')).to.be.undefined;
+		expect(client.messageComposerCache.peek('cid:a')).to.be.undefined;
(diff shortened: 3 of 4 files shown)


src/utils/FixedSizeQueueCache.ts
```diff
@@ -71,4 +71,17 @@ export class FixedSizeQueueCache<K, T> {
...
   }
+
+  /**
+   * Clears queue entirely and disposes of each item individually.
+   */
+  clear() {
+    if (this.dispose) {
+      for (const [key, entry] of this.map.entries()) {
+        this.dispose(key, entry);
+      }
```
arnautov-anton: One less transform operation.
```suggestion
      this.map.forEach((entry, key) => this.dispose(key, entry));
```

arnautov-anton (approved):

src/client.ts
```diff
@@ -393,6 +394,7 @@ export class StreamChat {
   defaultWSTimeout: number;
   sdkIdentifier?: SdkIdentifier;
   deviceIdentifier?: DeviceIdentifier;
+  messageComposerFixedSizeQueue: FixedSizeQueueCache<string, MessageComposer>;
```
arnautov-anton:
```suggestion
  readonly messageComposerFixedSizeQueue: FixedSizeQueueCache<string, MessageComposer>;
```

arnautov-anton: We should probably start using this thing more often for stuff that should not change whatsoever.

oliverlaz: While here, let's name this field `messageComposerCache` or something similar. I don't like the `*FixedSizeQueue` as it leaks an implementation detail into the field's name.

oliverlaz (approved):

PR #1776, opened 2026-06-18
fix: prevent user mentions when missing CreateMention permission

MartinCupela:
## Goal

Prevent including users / members in mentions suggestions when CreateMention channel permission is missing in `own_capabilities`.

Up until now, the `'create-mention'` was not included in `own_capabilities` and so the client side code could not perform decision, whether user mentions can be used during the message composition.


oliverlaz (approved):

MartinCupela:
V1 permission system does not allow to expose the create-mention permission so we cannot go forward with this PR.

PR #1778, opened 2026-06-21
fix(thread): handle notification.mark_unread event

TabishRiazBajwa:
## Summary

Fixes #3076

`Thread` currently subscribes to `message.read` events to keep its read state up to date, but never subscribes to `notification.mark_unread` events. This means that when a user marks a thread as unread, the `Thread`'s local state never reflects it — `unreadMessageCount` stays stale until the thread is reloaded.

## Root cause

`Channel` already has a working handler for `notification.mark_unread` (see `channel.ts`), and `Thread` already has all the state infrastructure needed (`ThreadUserReadState` with `unreadMessageCount`, `lastReadAt`, `lastReadMessageId`) — it was just never wired up to this event.

## Changes

- Added `subscribeRepliesMarkUnread()` to `Thread`, following the same `.on()` + filter pattern as the existing `subscribeRepliesRead()`
- Registered the new subscription in `registerSubscriptions()`
- Added 3 new test cases in `threads.test.ts`:
  - Ignores events from other threads (filtered by `parent_message_id`)
  - Ignores events for other users (only updates the current user's state)
  - Correctly updates `unreadMessageCount`, `lastReadAt`, and `lastReadMessageId` for the current user

## Testing

All existing tests in `threads.test.ts` pass (87/87), including the 3 new ones added for this fix.

## Notes

- Did not add `first_unread_message_id` tracking since `ThreadUserReadState` doesn't currently have an equivalent field (unlike `ChannelState`) — happy to add this in a follow-up if desired.
- Did not call `messageReceiptsTracker`, consistent with the existing `subscribeRepliesRead()` handler, since thread delivery receipts aren't supported yet.

## CLA

- [x] Code changes are tested

## Changelog

-

### DIFF
--- a/src/thread.ts
+++ b/src/thread.ts
@@ -272,6 +272,7 @@ export class Thread extends WithSubscriptions {
     this.addUnsubscribeFunction(this.subscribeMarkThreadStale());
     this.addUnsubscribeFunction(this.subscribeNewReplies());
     this.addUnsubscribeFunction(this.subscribeRepliesRead());
+    this.addUnsubscribeFunction(this.subscribeRepliesMarkUnread());
     this.addUnsubscribeFunction(this.subscribeMessageDeleted());
     this.addUnsubscribeFunction(this.subscribeMessageUpdated());
   };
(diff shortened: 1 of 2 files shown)


arnautov-anton:
Hey, @TabishRiazBajwa, thank you for your submission - I've changed a few things a bit and created a new PR so I'm going to close yours. I've added you as a co-author so that it still counts as a contribution from you. Unfortunately, while the PR adds a support for this feature client-side, the BE team is working on resolving a bug regarding `unread_messages` property which gets delivered with incorrect count. Until then - you may encounter incorrect unread count until this gets resolved.

TabishRiazBajwa:
Hi @arnautov-anton, thanks for the review and for adding me as co-author, really appreciate that. Good to know about the backend bug with unread_messages, makes sense to hold off until that gets sorted on the BE side.

Happy to pick up something else in the repo if there's anything useful I can help with. Looking forward to seeing it land!

PR #1787, opened 2026-06-29
feat: localized unread count

isekovanic:
## CLA

- [ ] Code changes are tested

Spec: https://stream-wiki.notion.site/Local-Unread-Count-specification-38b6a5d7f9f68073a5e1c73298fe0d71                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                        
Adds an optin, client side unread count for channels that have read events disabled (e.g. livestreams). Those channels normally don't track unread at all and so `_countMessageAsUnread` bails out whenever `own_capabilities` lacks `read-events`, so integrators have no "N new messages" affordance to show while scrolled up.
                                                                                                                                                                                                                                                                                                                        
With the new `isLocalUnreadCountEnabled` client option (default `false`), the SDK maintains a purely local unread count for those channels. It increments on incoming messages and is reset via the new `channel.markReadLocally()` method, which dispatches a client only `message.local_read` event. That event reuses the exact `message.read` state logic in `_handleChannelEvent` (so read state updates stay in one place) but skips `syncDeliveredCandidates()`, so the count is never sent to the backend. `hasReadEvents()` centralizes the capability check.                                                                                       
                                                                                                                                                                                                                                                                                                                        
When offline support is enabled, the local read/unread state is persisted for channels with disabled `read-events` so the count survives a cold start. The offline DB handles `message.local_read` (guarded so it can only affect channels with disabled `read-events` channels with the option on) and falls back gracefully when `state.read[userId]` is absent, since the server never sends a read for these channels.                                                                                                                                
                                                                                                                                                                                                                                                                                                                        
Behavior is fully gated behind the flag, so default off means no change for existing integrations.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                        
## Changelog                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                        
- Add `isLocalUnreadCountEnabled` client option for a client side unread count on channels with disabled `read-events` (e.g. livestreams)                                                                                                                                                                                        
- Add `Channel.markReadLocally()` to reset the local unread count with no backend call                                                                                                                                                                                                                                    
- Add `Channel.hasReadEvents()` helper                                                                                                                                                                                                                                                                                    
- Add client only `message.local_read` event                                                                                                                                                                                                                                                                              
- Persist local unread state to the offline DB so it survives app restarts

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1250,7 +1250,7 @@ export class Channel {
   /**
    * markReadLocally - Resets this user's unread count locally, without any backend call. Intended for
    * channels that have read events disabled (e.g. livestreams) when the client is created with the
-   * `enableLocalUnreadCount` option. Dispatches a dedicated, client-only `message.local_read` event
+   * `isLocalUnreadCountEnabled` option. Dispatches a dedicated, client-only `message.local_read` event
    * that runs through the same `_handleChannelEvent` read logic as a real `message.read` (minus the
    * delivery-report network sync), so the read-state update lives in one place. When offline support
    * is enabled, the offline DB persists the reset for read-events-disabled channels, so the local
@@ -1470,8 +1470,8 @@ export class Channel {
       return false;
 
     // Return false if channel doesn't allow read events, unless the client opted into a local
-    // unread count (e.g. livestreams where read events are disabled). See `enableLocalUnreadCount`.
-    if (!this.getClient().options.enableLocalUnreadCount && !this.hasReadEvents()) {
+    // unread count (e.g. livestreams where read events are disabled). See `isLocalUnreadCountEnabled`.
+    if (!this.getClient().options.isLocalUnreadCountEnabled && !this.hasReadEvents()) {
       return false;
     }
 
@@ -2004,7 +2004,7 @@ export class Channel {
         }
         break;
       // `message.local_read` is the client-only event dispatched by `markReadLocally()` when read
-      // events are disabled (e.g. livestreams with `enableLocalUnreadCount`). It reuses the exact
+      // events are disabled (e.g. livestreams with `isLocalUnreadCountEnabled`). It reuses the exact
       // `message.read` state logic so the read-state update lives in one place — only the
       // delivery-report network sync below is skipped for it.
       case 'message.local_read':
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1248,12 +1248,13 @@ export class Channel {
   }
 
   /**
-   * markReadLocally - Resets this user's unread count locally, without any backend call or offline-DB
-   * write. Intended for channels that have read events disabled (e.g. livestreams) when the client is
-   * created with the `enableLocalUnreadCount` option. Dispatches a dedicated, client-only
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -1591,6 +1549,12 @@ export type StreamChatOptions = AxiosRequestConfig & {
    */
   disableCache?: boolean;
   enableInsights?: boolean;
+  /**
+   * When true, maintains a client-local unread count on channels that have read events disabled
+   * (e.g. livestreams). The count increments on incoming messages and is reset via
+   * `channel.markReadLocally()`. It is never sent to the backend, but is persisted to the offline DB.
+   */
+  isLocalUnreadCountEnabled?: boolean;
```
oliverlaz, line 1599: to match the other props, maybe `enableLocalUnreadCounts`?

isekovanic: It was that exactly initially, however I had to change it to resolve [this comment](https://app.notion.com/p/Local-Unread-Count-specification-38b6a5d7f9f68073a5e1c73298fe0d71?d=38b6a5d7f9f6807daf06001c57266eda&source=copy_link#38b6a5d7f9f6806fadb7c399e9815e23).

szuperaz (approved):

src/events.ts
```diff
@@ -63,6 +63,7 @@ export const EVENT_MAP = {
   'ai_indicator.clear': true,
 
   // local events
+  'message.local_read': true,
```
MartinCupela:
```suggestion
  'message.read_locally': true,
```

MartinCupela: Just to follow the syntax of other WS names `<subject>.<state>_<additional-classification>`

isekovanic: Sure, makes sense, much better 👍

src/channel.ts
```diff
@@ -1421,6 +1447,20 @@ export class Channel {
...
   }
 
+  /**
+   * hasReadEvents - Whether read events are enabled for the current user on this channel. The channel
+   * is treated as not having read events only when its `own_capabilities` are known and exclude
+   * `read-events` (e.g. livestreams); when capabilities are unknown we assume read events are enabled.
+   *
+   * @return {boolean}
+   */
+  hasReadEvents() {
```
MartinCupela: Would `privacy_settings` also interfere here as the read events can be as well enabled / disabled there

isekovanic: Could be, have no clue ! How do I set it on the dashboard ?

isekovanic: Ok so did some digging @MartinCupela - `privacy_settings` actually affect capabilities. I guess the correct way is to always look at capabilities regardless (those 2 cannot diverge).

For example, a `channel` might have `read_events` in config but specific users might have different `privacy_settings` - and those users will not receive capabilities for read events.

It's basically the final resolved version of what we "are allowed to do".

MartinCupela: Yes, privacy_settings should have the last say in to determine, whether the feature is enabled.

MartinCupela (approved):

PR #1806, opened 2026-07-21
refactor: replace legacy ChannelState message/thread/pinned storage with paginators

MartinCupela:
### Summary

Completes the `message-paginator` initiative on the LLC side: the channel's **messages, thread replies, and pinned messages are no longer stored on `channel.state`**. Each list now lives in a paginator that is the single source of truth (interval storage + a canonical `ItemIndex`):

- **Main message list** → `channel.messagePaginator`
- **Thread replies** → `thread.messagePaginator` (owned by the `Thread` object)
- **Pinned messages** → `channel.pinnedMessagesPaginator` (new `PinnedMessagePaginator`)

### What changed

- **`ChannelState` storage removed:** `messages`, `latestMessages`, `messageSets`, `messagePagination`, `threads`, `pinnedMessages`, and their mutators (`addMessageSorted`/`addMessagesSorted`, `removeMessage`, `findMessage`, `findMessageByTimestamp`, `filterErrorMessages`, `loadMessageIntoState`, `clearMessages`, `initMessages`, `pruneOldest`, `addReaction`/`removeReaction`, `updateUserMessages`, `deleteUserMessages`, `addPinnedMessages`/`addPinnedMessage`/`removePinnedMessage`, `removeQuotedMessageReferences`, + internal helpers).
- **`PinnedMessagePaginator`** added and populated from channel events; pin/unpin falls out of `ingestItem` + a `{ pinned: true }` filter (no bespoke branching).
- **`MessageIntervalPaginator`** extracted as the unread-free base of `MessagePaginator`; it tracks the latest message (`latestMessageId` + `latestMessage`), advanced on ingest and mirrored into reactive state.
- **`channel.state.last_message_at`** is now a **read-only getter** derived from `messagePaginator.latestMessage` (setter + backing field removed).
- **`channel.state.isUpToDate` / `setIsUpToDate` removed** — live-message routing (don't disrupt a scrolled-away view) is handled structurally by the paginator's interval model.
- **`Channel._trackLatestMessage` removed.** `user.updated` / `user.deleted` propagation now scans active channels (`reflectUserUpdate` / `applyMessageDeletionForUser` self-filter by author id). A targeted user-reference index is planned (`specs/user-reference-index`).
- **`utils` cleanup:** removed `addToMessageList`, `messageSetPagination`, `binarySearchByDateEqualOrNearestGreater`, `deleteUserMessages`, and the `MessageSet` / message-set pagination types.
- **Reactive `headItems`** added to `PaginatorState` (newest-loaded window).
- **Removed the unused `MessageReplyPaginator`** (dead sibling; the thread reply list uses `MessagePaginator`).

### Breaking changes

Full list + before → after migration table in **`docs/breaking-changes-v14-v15.md`**. Highlights:

| Before (v14) | After (v15) |
| --- | --- |
| `channel.state.messages` | `channel.messagePaginator.state.items` / `.items` |
| `channel.state.threads[parentId]` | `thread.messagePaginator.state.items` |
| `channel.state.pinnedMessages` | `channel.pinnedMessagesPaginator.state.items` |
| `channel.state.addMessageSorted(m)` | `channel.messagePaginator.ingestItem(m)` |
| `channel.state.removeMessage({ id })` | `channel.messagePaginator.removeItem({ id })` |
| `channel.state.findMessage(id)` | `channel.messagePaginator.getItem(id)` |
| `channel.state.isUpToDate` / `setIsUpToDate` | `messagePaginator.isActiveIntervalAtHead` / `hasMoreHead` / `jumpToTheLatestMessage()` |
| `channel.state.last_message_at = …` | read-only (derived); ingest a message on the paginator |

Behavioral note: `last_message_at` now advances on every incoming message regardless of the viewer's scroll position (the old `isUpToDate` suppression is gone) — it's a channel-level fact.

### Follow-ups (not in this PR)

- **`specs/user-reference-index`** — replace the active-channel scan on `user.updated`/`user.deleted` with a `userId → message-reference` index (design-gated).
-  Unifying the remaining flat paginators on interval storage
- Offline migration


src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -0,0 +1,1188 @@
...
+   * permits it. Updates only the instance field (no state write) so it is safe to call from inside a
+   * `state.next` updater; the mirror preprocessor reflects it on the next emission. Returns `true`
+   * when the pointer moved.
+   *
+   * The current latest's timestamp is resolved from the item index rather than cached: there is no
+   * interval trimming, so the tracked message stays resolvable while loaded; the only way it leaves
+   * the index is deletion (its author hard-deleted / an explicit remove), and in that case advancing
+   * to the next newest loaded message is the correct outcome, not a regression.
+   */
+  protected advanceLatestTo(message: LocalMessage): boolean {
```
isekovanic: The comment says removing the tracked latest advances to the next newest, but as far as I can see (might totally be wrong) nothing actually does that and `_latestMessageId` only moves forward in `advanceLatestTo` or gets `null`ed in `clearStateAndCache` and there's no `removeItem` override. So if you hard delete the current latest (or a `user.deleted` + `hard_delete` on the author lands in `applyMessageDeletionForUser` and then `removeItem`), the `id` it points at is gone from the index, `latestMessage` goes `undefined` and `last_message_at` will be `null` until the next message shows up, which flickers the channel list sort/preview. Soft delete's fine since the msg stays indexed. Can we repoint to the head window latest in `removeItem` when the removed id was the tracked one (probably want the same in truncate) ?

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -0,0 +1,1188 @@
...
+        return leftId < rightId ? -1 : leftId > rightId ? 1 : 0;
+      },
+    });
+    this.setFilterResolvers([dataFieldFilterResolver]);
+
+    // Mirror the tracked-latest source-of-truth field into reactive state on every emission. Doing
+    // it in a preprocessor (rather than a `partialNext` at each advance) means advances made from
+    // inside a `state.next` updater - `ingestPage` via `setItems`/query - are reflected by the very
+    // update that commits the page, instead of being clobbered by it.
+    this.state.addPreprocessor((nextValue) => {
```
MartinCupela: Actually updates the state for latestMessageId

src/pagination/paginators/MessageIntervalPaginator.ts
MartinCupela: there is a statestore preprocessor that enriches `nextValue.latestMessageId = this._latestMessageId;`. You can see it here: https://github.com/GetStream/stream-chat-js/pull/1806/changes/BASE..2846f0678cf3df08592b7b1aaa5bd3904862ece7#r3622823887

src/pagination/paginators/BasePaginator.ts
```diff
@@ -457,6 +464,33 @@ export abstract class BasePaginator<T, Q> {
     this._filterFieldToDataResolvers = [];
     this._usesItemIntervalStorage = !!itemIndex;
     this._itemIndex = itemIndex ?? new ItemIndex({ getId: this.getItemId.bind(this) });
+
+    // Materialize the reactive `headItems` (newest-loaded window) on every state emission. There is
```
isekovanic: Ok so I dug into the tracking of latest items (so `headItems`/`latestMessage`/`latestMessageId`) since we want to build the RN `ChannelList` previews on it and there are two things that I think will be a problem with this, but please feel free to correct me if I'm wrong. Both come straight out of how `headItems` is derived since it's not written directly, it's stamped on by a `preprocessor` that runs on every state emission and reads `latestItems` and then goes through `getHeadIntervalFromSortedIntervals` to get the topmost loaded interval. 

Due to this, I believe that we have some sideeffects:
                                                                                                                                                                                                                                                                                                                          
1. `headItems` only reactively updates while you're at the bottom. The preprocessor only runs when state emits and `ingestItem`'s emit is gated on the active interval being hit. So "latest" is really piggybacking on the active window's emit trigger. If we subscribe to `state.headItems` it will stay current only while the head interval is the active one. If we move to a different message set (so head is still loaded, just no longer active) and a new message extends the head interval without emitting, so `state.headItems` silently lags (and anything that relies on this state is now stale). The value is correct in the intervals (`latestItems` getter returns it) but it's purely the notification that never fires. Self heals  
on the next unrelated emit of course, but for the canonical "latest" it won't be this all the time                                                                                                                                                                                                                                                      
2. When the actual head interval is not loaded (for example we have 2 message list pages that are not the latest one for whatever reason) it reports a nonhead window as "latest" instead of `undefined`. `getHeadIntervalFromSortedIntervals` returns the topmost loaded interval unconditionally with no check that it's actually the head. So when we haven't loaded the head (opened around an old message, or jumped far and the newest page was never fetched for some reason, I know this will never be the case with our UI SDK but it might be for the state layer in general), `headItems` returns that older window and presents it as newest and so nothing signals to the consumer that it's not real. Could we return `undefined` when the top interval isn't genuinely head (probs something like`isLogicalInterval(head) || !head.isHead`), so "latest not loaded" is representable? As far as I can see, `seedFirstPageSync` already flags actual first pages as `isHead`, so channel list channels still resolve BUT your note on `latestItems` says `isHead` isn't reliable on the seed, so lemme know what I'm missing; if it genuinely can't be trusted that feels like the root thing to fix, since everything latest keys off "is this the head". If we have any signal we should use it (probably the pagination cursor should be good, right ?).   
3. This is not really an issue, but I'm generally against the prospect of recomputing `headItems` whenever any part of the state changes. The state is likely to grow even larger and we probably don't want this many recomputations (even though they will be mostly thrown away as we do a shallow comparison across all items). Something like this: `candidate.every((item, i) => item === previous[i])` will not be so cheap to do potentially in a livestream with many thousands of messages and this will happen on every single state change.                                                                                                                                           
                                                                                                                                                                                                                                                                                                                          
Some thoughts on how we could approach this:
                                                                                                                                                                                                                                                       
1. Stop recomputing `headItems` in the blanket preprocessor (it reruns `intervalToItems`). Recompute `headItems` + `latestMessageId` into the instance fields at the head affecting mutation sites only, deduped and then the preprocessor can shrink to an `O(1)` computation, kept purely to survive the `ingestPage` thingy I guess.                                                                                                                                                                                                                   
2. Drop the active interval gate for the latest so that on the live paths (`ingestItem`/`removeItem`), emit `latestMessageId` when it actually changed regardless of the active interval and that drags `headItems` along through the mirror, so it's reactive from any interval we are currently viewing.                                                       
                                                                                                                                                                                                                                                                                                                          
That way, we should get the isolation for free from `useStateStore` and end up doing less work. Thoughts on this ?

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: Yeah, that preprocessor is exactly the bit I was looking at, but it mirrors the field, it doesn't rederive it here, so `nextValue.latestMessageId = this._latestMessageId` just copies whatever `_latestMessageId` currently holds onto state each emission (I guess this is the correct term for emits ? :D ). So if the field is pointing at a message that's already been removed, the preprocessor republishes that same stale `id`.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
I think the actual gap is that nothing repoints the field on removal. The only writes to `_latestMessageId` that I could fine are `advanceLatestTo` (addition only) and `clearStateAndCache` (just to clear it), but there's no `removeItem` override. So hard deleting the current latest message would do `removeItem({ id: theLatestMessage.id })` drops and will drop the message from the index, however `_latestMessageId` would still be `theLatestMessage.id`. After this the preprocessor will mirror `theLatestMessage.id` onto then state and `latestMessage = getItem(M.id) = undefined` causing `last_message_at` to be `null`.  
                                                                                                                                                                                                                                                                                                                            
It will only recover on the next message (`advanceLatestTo` sees the current latest is `undefined` and advances) and soft delete's fine since `theLatestMessage` stays indexed so `getItem` still resolves it.
                                                                                                                                                                                                                                                                                                                            
So the preprocessor's doing its job, it's just downstream of the gap and doesn't cover it, so I think we need to move the field itself in `removeItem` (and `truncate` most likely since that's effectively removal, but probably only the full truncate without  a date) when the removed `id === _latestMessageId`, repointing to the head window latest. 

Or am I missing something in the preprocessor here ?

isekovanic: Ah but wait, I think I got it wrong. It would appear that removing a message does not remove it from the item index, is this intentional ? Because then it won't be `null` as I wrote above but it'll actually point to a stale message (which is worse)

src/pagination/paginators/BasePaginator.ts
MartinCupela: I addressed it here: https://github.com/GetStream/stream-chat-js/pull/1806/commits/8bc92437c8903f859252aa7db8ee8a4338ac8b9f

I have added a new StateStore called `aggregateState`, that is updated independently from the pagination activity. Also lastMessageAt value is retrieved solving a possible out of sync issue if we have lastMessageAt timestamp and lastMessage reference in the state. Now we seed what comes in from the server and if newer messages are registered as lastMessage, then the seed is ignored. Last message concept is stored only for MessagePaginator as PinnedMessagesPaginator would not need it.

I have moved all the paginators into itemIndex storage. That allowed me to remove `headItems` from the PaginatorState as they do not make much sense as now we can retrieve head items easily with existing API and they are up to date in the paginator state.

MartinCupela: Next I will try to remove `BasePaginator._usesItemIntervalStorage` as it is redundant.

isekovanic (approved):
Thanks @MartinCupela !

PR #1812, opened 2026-08-04
fix: post-merge/OAPI-reconciliation adjustments

arnautov-anton:


src/utils.ts
```diff
@@ -802,7 +802,7 @@ export const generateChannelTempCid = (channelType: string, members: string[]) =
 export const isDate = (value: unknown): value is Date => !!(value as Date).getTime;
 
 export const isLocalMessage = (message: unknown): message is LocalMessage =>
-  isDate((message as LocalMessage).created_at);
+  typeof (message as LocalMessage | undefined)?.status === 'string';
```
arnautov-anton, line 781: Important one to check, @isekovanic and @MartinCupela.

src/types.ts
```diff
@@ -1107,35 +1107,3 @@ export type GiphyVersions = keyof Images;
...
-  Omit<DraftMessage, 'mentioned_groups'>,
-  'id'
-> & {
-  user_id?: string;
-};
-
-export type QueryRemindersOptions = Pager & {
-  filter?: ReminderFilters;
-  sort?: ReminderSort;
-};
```
arnautov-anton: Replaced all of these with generated equivalents.

oliverlaz (approved):

src/utils.ts
szuperaz: Verified and this one should work

PR #1814, opened 2026-08-05
feat: generalize entity store in V10

isekovanic:
## CLA

- [x] Code changes are tested

Generalizes the `MessageStore` into a reusable generic layer, makes the pinned message list reactive and collapses the `ItemIndex` abstraction to a single implementation.

### What & Why

The reactive `MessageStore` and its `MessageStoreBackedItemIndex` were message specific, but nothing about them actually depends on message and the domain type is only used to pull an `id`. This PR lifts that into a generic, reusable core so any collection can share the same normalized/reactive machinery and removes redundant stuff.

**1. `MessageStore` becomes generic `EntityStore<T>`**
- New `EntityStore<T>` holds the entire former store body; its constructor takes an options object `{ getId }` (`EntityStoreOptions<T>`), matching the `ItemIndex({ getId })` idiom
- The `MessageStore` **subclass is deleted** since it existed only to bind `getId`. `client.messageStore` is now `new EntityStore<LocalMessage>({ getId: (m) => m.id })`
- `MessageStoreBackedItemIndex` becomes a generic `StoreBackedItemIndex<T>`, with an **optional `store`** (falls back to a private `EntityStore`, so a detached index behaves like a plain per-instance one)

**2. Pinned message list is now store backed**
- The store backed `ItemIndex` factory has been moved up to `MessageIntervalPaginator` (common ancestor of the main list, thread replies and pinned list), so `PinnedMessagePaginator` now shares the client global store instead of holding a private `ItemIndex`
- I somehow forgot to do this in the previous reactive collections PR, even though this is exactly what prompted that architectural design in the first place 😅 

**3. Delete the standalone `ItemIndex` class**
- With the store now optional, a storeless `StoreBackedItemIndex` is behaviorally identical to the old `ItemIndex`. `owner` is now optional too and so the four remaining `new ItemIndex(...)` sites (Base default, `Channel/Reminder/UserGroup` paginators) route through `StoreBackedItemIndex`.
- The `ItemIndex` class and `ItemIndexOptions` are removed; the `ItemIndexApi` interface (the contract `BasePaginator` depends on) is kept, now with a single implementation

### Public API surface

| Removed | Added |
|---|---|
| `MessageStore` (class) | `EntityStore` (class) |
| `MessageStoreChangeBatch` | `EntityStoreChangeBatch` |
| `MessageStoreSubscriber` | `EntityStoreSubscriber` |
| — | `EntityStoreOptions` |

### Testing

- `yarn types`, `yarn lint` (0 warnings), `yarn build` all clean.
- Full unit suite green. New coverage: pinned-message fan-out (`reactions.optimistic.test.ts`) and owner-less/store-less CRUD (`StoreBackedItemIndex.test.ts`). The `ItemIndex.test.ts` suite was folded into `StoreBackedItemIndex.test.ts`.
- Consuming SDKs verified unaffected — the renamed symbols have no importers in `stream-chat-react-native`.


src/entityStore/EntityStore.ts
```diff
@@ -0,0 +1,204 @@
+import type { Unsubscribe } from '../store';
```
MartinCupela, line 1: I would rename the parent folder from messageStore to entityStore

isekovanic: Ah, good catch - missed this.

Thanks

src/messageStore/EntityStore.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /**
+   * Immediately flushes any throttled/pending state publish on the holders of `id` (via
+   * {@link EntityStoreSubscriber.flushState}). Called after an optimistic (local-user) write to
+   * `id` so it renders without the throttle delay. Only that id's own holders are flushed — the
+   * write touched no other id — and flushing a holder with nothing pending is a no-op.
+   */
+  flushSubscribers(id: string): void {
+    const holders = this.subscribers.get(id);
```
MartinCupela: is holder a subscriber?

src/messageStore/EntityStore.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  // ---- writes ----
+
+  /**
+   * Replaces the canonical copy of `entity` and notifies every subscriber watching its id —
+   * except `origin`, which is expected to re-render itself (the paginator that performed the
+   * write already emits its own window inline, so it must not be notified a second time through
+   * the subscription).
+   */
+  upsert(entity: T, origin?: EntityStoreSubscriber): void {
```
MartinCupela: Here subscriber is called origin. In other place it is holder. Can we just call it subscriber everywhere?

src/messageStore/StoreBackedItemIndex.ts
```diff
@@ -0,0 +1,115 @@
+import type { ItemIndexApi } from '../pagination/ItemIndex';
+import { EntityStore, type EntityStoreSubscriber } from './EntityStore';
+
+export type StoreBackedItemIndexOptions<T> = {
+  getId: (item: T) => string;
```
MartinCupela:
```suggestion
  getEntityId: (item: T) => string;
```

MartinCupela (approved):

MartinCupela (approved):

PR #1819, opened 2026-08-11
fix: drop server-side remnants, add WS interface

arnautov-anton:


src/utils.ts
```diff
@@ -125,23 +100,14 @@ export const channelTracksReadLocally = (channel?: Channel) =>
...
-    data.append('file', {
-      uri,
-      name: name || (uri as string).split('/').reverse()[0],
-      contentType: contentType || undefined,
-      type: contentType || undefined,
-    });
+    data.append('file', uri);
   }
 
   return data;
```
arnautov-anton: This right here is the biggest change and would need some testing from RN guys too, not sure whether this is the correct typing. @isekovanic

src/connection.ts
```diff
@@ -337,7 +341,10 @@ export class StableWSConnection {
         wsURL,
         requestID: this.requestID,
       });
-      this.ws = new WebSocket(wsURL);
+
+      const WS = this.client.options.WebSocketImpl ?? WebSocket;
+      this.ws = new WS(wsURL);
```
arnautov-anton, line 346: @oliverlaz

src/signing.ts
```diff
@@ -78,253 +10,7 @@ export function UserFromToken(token: string) {
...
- * decompressed), the bytes are returned unchanged.
- *
- * Magic-byte detection (rather than relying on a header) keeps the same handler correct when
- * middleware — Express, Next.js, AWS Lambda — auto-decompresses the request before your code sees it.
- *
- * @param rawBody - The raw HTTP request body, either as a string or a `Buffer`.
- * @returns The uncompressed payload bytes.
- * @throws {@link InvalidWebhookError} when the gzip envelope is malformed.
- */
-export function gunzipPayload(rawBody: string | Buffer): Buffer {
```
oliverlaz, line 186: @szuperaz do we have these APIs in the node-sdk?

oliverlaz (approved):

src/signing.ts
szuperaz: Yes, they were added to stream-node too when they were added to stream-chat-js: https://github.com/GetStream/stream-node/pull/294

src/utils.ts
szuperaz: Checking

szuperaz: Fixed in dd957833d29f776000abdc8614830404af77d42e

PR #1822, opened 2026-08-12
feat: destructive reconciliation

isekovanic:
## CLA

- [x] Code changes are tested

Adds **destructive reconciliation** of the channel message list. On reconnect/rehydrate, messages that were hard-deleted by anyone while the client was offline are removed from the list, for the channel main list and thread replies, with offline support enabled or disabled.

A hard delete emits no event to other clients, and the newest page merge is additive (it never removes), so a message deleted while you were offline lingers as a ghost forever after reconnect and with offline support on, it also survives in SQLite and returns on cold start.

Reconciliation is folded into the existing newest page merge, so every reseed path (`channel.query`, channel list hydrate, `channel.reload`, `Thread.reload`, `recoverState`) reconciles uniformly, without a full reload and without dropping already-loaded older pages. A loaded message that's absent from the authoritative page is pruned; a prefetch snapshot distinguishes a genuine hard-delete from a message that arrived live during the fetch, and local (optimistic/failed/error) messages are always kept. Reconciled ids are mirrored into the offline DB so a cold start doesn't reseed them. I've also added a billion unit tests that should capture this behaviour because it's an incredible pain to test all of the nooks and crannies here.

## Changelog

- **Feature:** hard deletes performed by other users while you were offline are now removed from the message list (and threads) on reconnect/re-hydrate instead of lingering as ghosts, so no full reload and mirrored into the offline DB so they don't return on cold start.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -17,7 +17,6 @@ import {
 } from './utils';
 import type { StreamChat } from './client';
 import { chatLoggerSystem } from './logger';
-import { DEFAULT_QUERY_CHANNEL_MESSAGE_LIST_PAGE_SIZE } from './constants';
 import type {
   AIState,
   APIResponse,
@@ -175,6 +174,8 @@ export class Channel extends ChannelApi {
   lastTypingEvent: Date | null;
   isTyping: boolean;
   disconnected: boolean;
+  /** Re-entrancy guard for {@link Channel.reload} (mirrors Thread.reload's isLoading guard). */
+  private _reloading = false;
   push_preferences?: Gen_ChannelPushPreferencesResponse;
   public readonly configState = new StateStore<ChannelInstanceConfig>({});
   public readonly messageComposer: MessageComposer;
@@ -1298,6 +1299,46 @@ export class Channel extends ChannelApi {
   }
 
   /**
+   * Re-watch the channel and refresh its FULL loaded message window — the channel analog of
+   * {@link Thread.reload}. Used on reconnect to catch up AND reconcile hard deletes that happened
+   * while offline: a hard delete reaches other clients via no event, so an offline client only learns
+   * of it by diffing the re-queried page.
+   *
+   * This is intentionally thin: it only re-issues `watch()` with a limit sized to the loaded window
+   * (`items.length`, so the whole loaded window is refreshed — not the smaller channel-list page). The
+   * actual fold + destructive reconciliation happens inside `query()` → `seedFirstPageSync`
+   * (the same path the channel-list re-hydrate and React's `recoverState` use), driven by the loaded-id
+   * snapshot `query()` captures before its await. Owning that single path is what lets the SDK stop
+   * passing the reconciliation window/snapshot itself.
+   *
+   * Preserves failed (unsent) messages: an overlap merge keeps them (the reconcile's provenance guard
+   * never prunes a non-server message); only a disjoint rebuild can drop them, so any that actually
+   * fell out are re-ingested below.
+   */
+  async reload() {
+    if (this._reloading || (!this.initialized && !this.offlineMode)) return;
+    this._reloading = true;
+    try {
+      const paginator = this.messagePaginator;
+      // Captured BEFORE the await: request our full loaded window (not the list's smaller page), and
+      // remember failed (unsent) messages so a disjoint rebuild does not silently drop them.
+      const requestedLimit = paginator.items?.length || paginator.pageSize;
+      const failedBefore = (paginator.items ?? []).filter(
+        (message) => message.status === 'failed',
+      );
+
+      await this.watch({ messages: { limit: requestedLimit } });
+      this.offlineMode = false;
+
+      for (const failed of failedBefore) {
+        if (!paginator.getItem(failed.id)) paginator.ingestItem(failed);
+      }
+    } finally {
+      this._reloading = false;
+    }
+  }
+
+  /**
    * Stops watching the channel.
    *
    * @param request - The stop-watching request payload (optional).
@@ -1482,6 +1523,24 @@ export class Channel extends ChannelApi {
     options: ChannelGetOrCreateRequest = {},
     messageSetToAddToIfDoesNotExist: MessageSetType = 'current',
   ) {
+    // Snapshot the loaded message ids BEFORE the network await, for a latest-window (re)seed only.
+    // When this query re-seeds an already-loaded window (reconnect / re-hydrate), seedFirstPageSync
+    // reconciles the fresh page against what is loaded, and the snapshot lets it tell an offline
+    // hard-delete (in the snapshot, absent from the page) from a message that arrives live
+    // during the fetch (not in the snapshot). Captured here — the only place with the pre-await state —
+    // so callers (channel.reload, watch) need not thread it. Empty on a cold open, so it is harmless.
+    const candidateIds =
+      messageSetToAddToIfDoesNotExist === 'latest'
+        ? new Set(this.messagePaginator.items?.map((message) => message.id) ?? [])
+        : undefined;
+
+    // The INITIAL channel-open query honors the paginator's OWN pageSize (light on native, 25) rather
+    // than the server's larger default — opening loads the same page size it paginates by. A caller
+    // that already knows how much to fetch passes an explicit messages.limit, which is respected as-is:
+    // a reconnect/re-hydrate sizes it to the loaded window (channel.reload → items.length), and
+    // pagination/around pass their own cursors + limit.
+    const requestedPageSize = options?.messages?.limit ?? this.messagePaginator.pageSize;
+
     // Make sure we wait for the connect promise if there is a pending one
     await this.getClient().wsPromise;
 
@@ -1489,6 +1548,13 @@ export class Channel extends ChannelApi {
       data: this._data,
       state: true,
       ...options,
+      // Ask the server for exactly the initial-open page size (not its default), so the loaded window
+      // matches the paginator's pageSize. Explicit messages (reconnect/around/pagination) pass through.
+      messages:
+        options?.messages ??
+        (messageSetToAddToIfDoesNotExist === 'latest'
+          ? { limit: requestedPageSize }
+          : undefined),
     };
 
     const state = this.id
@@ -1542,8 +1608,6 @@ export class Channel extends ChannelApi {
     // latest-page open paths (watch/create) pass 'latest' — the paginator's own pagination queries
     // use 'current' and must not be reseeded as a first page here.
     if (messageSetToAddToIfDoesNotExist === 'latest' && Array.isArray(state.messages)) {
-      const requestedPageSize =
-        options?.messages?.limit ?? DEFAULT_QUERY_CHANNEL_MESSAGE_LIST_PAGE_SIZE;
       // Pass the query's message pagination options through: a channel can be opened AROUND a
       // message (id_around / created_at_around), in which case the fetched page is a jump window,
       // not the latest page — the paginator must reconcile it with jump semantics.
@@ -1551,6 +1615,8 @@ export class Channel extends ChannelApi {
         state.messages.map(formatMessage),
         requestedPageSize,
         options?.messages,
+        // Re-seed of an already-loaded window folds + reconciles instead of blanking (see above).
+        { candidateIds, reconcile: true },
       );
     }
 
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1328,9 +1328,14 @@ export class Channel extends ChannelApi {
       await this.watch({ messages: { limit: requestedLimit } });
       this.offlineMode = false;
 
-      for (const failed of failedBefore) {
-        if (!paginator.getItem(failed.id)) paginator.ingestItem(failed);
-      }
+      paginator.batch(
+        () => {
+          for (const failed of failedBefore) {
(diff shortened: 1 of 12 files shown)


src/channel.ts
```diff
@@ -1297,6 +1298,46 @@ export class Channel extends ChannelApi {
...
+   * Preserves failed (unsent) messages: an overlap merge keeps them (the reconcile's provenance guard
+   * never prunes a non-server message); only a disjoint rebuild can drop them, so any that actually
+   * fell out are re-ingested below.
+   */
+  async reload() {
+    if (this._reloading || (!this.initialized && !this.offlineMode)) return;
+    this._reloading = true;
+    try {
+      const paginator = this.messagePaginator;
+      // Captured BEFORE the await: request our full loaded window (not the list's smaller page), and
```
MartinCupela: I don't understand this comment

src/channel.ts
```diff
@@ -1297,6 +1298,46 @@ export class Channel extends ChannelApi {
...
+      const requestedLimit = paginator.items?.length || paginator.pageSize;
+      const failedBefore = (paginator.items ?? []).filter(
+        (message) => message.status === 'failed',
+      );
+
+      await this.watch({ messages: { limit: requestedLimit } });
+      this.offlineMode = false;
+
+      for (const failed of failedBefore) {
+        if (!paginator.getItem(failed.id)) paginator.ingestItem(failed);
```
MartinCupela: Can we batch this ingestion to prevent re-emitting new state for each failed message?

src/channel.ts
```diff
@@ -1297,6 +1298,46 @@ export class Channel extends ChannelApi {
...
+      const paginator = this.messagePaginator;
+      // Captured BEFORE the await: request our full loaded window (not the list's smaller page), and
+      // remember failed (unsent) messages so a disjoint rebuild does not silently drop them.
+      const requestedLimit = paginator.items?.length || paginator.pageSize;
+      const failedBefore = (paginator.items ?? []).filter(
+        (message) => message.status === 'failed',
+      );
+
+      await this.watch({ messages: { limit: requestedLimit } });
+      this.offlineMode = false;
```
MartinCupela, line 1328: Would it make sense to set the offlineMode flag inside the watch or query methods instead?

src/channel.ts
isekovanic: You're not meant to, it's a leftover from a previous experiment that never got removed :D I'll remove it, thanks 👍

src/channel.ts
```diff
@@ -1297,6 +1298,46 @@ export class Channel extends ChannelApi {
...
+   * fell out are re-ingested below.
+   */
+  async reload() {
+    if (this._reloading || (!this.initialized && !this.offlineMode)) return;
+    this._reloading = true;
+    try {
+      const paginator = this.messagePaginator;
+      // Captured BEFORE the await: request our full loaded window (not the list's smaller page), and
+      // remember failed (unsent) messages so a disjoint rebuild does not silently drop them.
+      const requestedLimit = paginator.items?.length || paginator.pageSize;
```
MartinCupela: `paginator.items?.length` could contain more than one page and so the requestedLimit could end up being more than the permitted page size on the server side?

src/channel.ts
isekovanic: can do

src/channel.ts
isekovanic: Yes, and this will still return 100 at most (which we still respect for handling these things). Maybe we can do something like `Math.max(requestedLimit, DEFAULT_CHANNEL_PAGE_SIZE` or whatever that constant was called. I believe pagination should not be broken with this but I can try it out and see

src/channel.ts
isekovanic: Eventually probably yes, for now I'm just keeping it ported from what we had on RN so I don't break stuff. Anyway with the channel state refactor a lot of these will (hopefully) go away. It just didn't make sense to clean them up here specifically.

isekovanic: Sorry, by this:

> Eventually probably yes

I mean remove some of these fields if they aren't useful

src/channel.ts
MartinCupela: Not related to this PR:
I am also thinking about having to ask paginator for the item (already updated after the request) and then asking the paginator to ingest it - could look confusing to integrators. Maybe we should have cache service with all the entity stores - from there we would request it and then ingest it into the paginator

src/channel.ts
```diff
@@ -1482,13 +1523,44 @@ export class Channel extends ChannelApi {
...
+    // during the fetch (not in the snapshot). Captured here — the only place with the pre-await state —
+    // so callers (channel.reload, watch) need not thread it. Empty on a cold open, so it is harmless.
+    // TODO(perf/cleanup): `headItems` materializes full message objects (intervalToItems) just to map
+    // them down to ids. A cheaper, clearer equivalent is a straight copy of the paginator index's own
+    // id set — expose `memberIds` on StoreBackedItemIndex (e.g. `snapshotMembers()` returning
+    // `new Set(this.memberIds)`) and use it here AND in client.queryChannelsAndHydrate. The broader
+    // scope (all intervals vs just the head) is inert: the reconcile only consults head ids, older
+    // island ids are never at/above-newest, and local messages are guarded by isServerConfirmedMessage.
+    const candidateIds =
+      messageSetToAddToIfDoesNotExist === 'latest'
```
MartinCupela, line 1541: Out of scope of the PR: we need to get rid of this `messageSetToAddToIfDoesNotExist` flag

MartinCupela: https://linear.app/stream/issue/REACT-1075/get-rid-of-messagesettoaddtoifdoesnotexist-in-llc

isekovanic: Yes, we do, I don't think it's needed anymore

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,192 @@ export class MessageIntervalPaginator extends BasePaginator<
       return;
     }
 
-    // Overlapping window: merge in place, preserving the older boundary.
+    // Overlapping window: merge in place, keeping every already-loaded (incl. older) item.
     const interval = this.ingestPage({ page, isHead: true, setActive: false });
     if (!interval) return;
 
+    // Re-compute hasMoreTail from the FETCHED PAGE, not the merged interval. The interval's isTail is
```
MartinCupela: Do we miss tests for this?

src/channel.ts
isekovanic: I guess we could have an API for something like this, but it is not always the case that you don't want to reingest an item. Here it's basically a specific case where we have explicit control over this and don't really want to even try to reingest because the items are by default local only (the guard is only there for this specific case).

I was also thinking about perhaps making `ingestItem` idempotent but I'm not sure that's correct at all times (and also it will be an issue for newly fetched entities). So maybe this is something that needs to remain case by case.

src/channel.ts
isekovanic: Confirmed that it does not break pagination (handled this due to other reasons anyway) and added a test for this.

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,194 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+    const oldestReturnedTs = getMessageCreatedAtTimestamp(page[0]);
+
+    // Only extend below the oldest returned message when the page proves it reached the channel's
+    // oldest message: it came back shorter than requested. Clamp the request to the server's max page
+    // size so asking for MORE than one page can return (an over-request) is not mistaken for reaching
+    // the start — in that case the shortfall is the server capping, not the channel ending.
+    const { requestedLimit } = options ?? {};
+    const reachedChannelStart =
+      typeof requestedLimit === 'number' &&
+      page.length < Math.min(requestedLimit, DEFAULT_CHANNEL_MESSAGE_LIST_PAGE_SIZE);
```
MartinCupela: The hard-coded channel value would be applied to Threads as well? I am wondering whether it is possible to take the configured value on the paginator instance.

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,194 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+   * Removal goes through {@link removeItem} (batched) so the item index, intervals, shared message
+   * store and — via the {@link MessagePaginator} override — the tracked last message all stay
+   * correct, then the active window is re-emitted once.
+   */
+  protected reconcileLoadedAgainstPage(
+    page: LocalMessage[],
+    options?: MergeNewestPageOptions,
+  ) {
+    const headInterval = this.itemIntervals[0] as Interval | undefined;
+    if (!headInterval?.isHead) return;
```
MartinCupela: This reconciliation is only for head intervals? If so, then we may want to change the method name.

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,194 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+   * size — so it can only ever remove messages the page actually covers:
+   *
+   * - WITHIN the page's span (`oldest returned < created_at < newest returned`): a server-confirmed
+   *   loaded message absent from the page was hard-deleted. Safe with no snapshot — a message that
+   *   arrived live during the caller's fetch is always strictly newer than the newest returned
+   *   message, so it can never fall in this span.
+   * - BELOW the oldest returned message: only reconcilable when the page reached the channel's oldest
+   *   message (`requestedLimit` given and the page came back short, clamped to the server max page
+   *   size). Otherwise older messages simply were not fetched and are left untouched.
+   * - AT/ABOVE the newest returned message (a hard-deleted newest message) and the empty-page case:
```
MartinCupela: It is hard to read the docstrings as they are sometimes missing context around the substantives. "The newest returned message" from where returned?

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,194 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+  }
+
+  /**
+   * Remove a set of reconciled (hard-deleted) ids in one batch — coalescing the shared-store fan-out
+   * to a single flush — then flush the deferred window publish so the list drops the ghosts
+   * synchronously (blanking to `[]` if the active window emptied, per {@link flushWindowPublish}).
+   * Finally, mirror the removal into the offline DB so a cold start does not re-seed the ghosts from
+   * SQLite. No-op for an empty set, so an unaffected merge does not touch state a second time.
+   */
+  private removeReconciledIds(ids: string[]) {
```
MartinCupela: From the method name it is not entirely clear, from where the ids are being removed

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: Not really, why do you have that impression ?

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -603,19 +688,194 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+    if (!ids.length) return;
+    this.batch(
```
MartinCupela, line 946: I think that batch method actually does not prevent re-emission from the underlying paginator state store for each id. For each removed id the store re-emits I think. I do not see, what would prevent that.

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: It would and it needs to apply to those as well. The point is not page size (100 is anyway too much for a page size at least for mobile), but rather the maximum BE can return. I'll try resolving this a different way and make sure pagination doesn't break. But in general we need to fetch as many as we can as there is no other way to reconcile updates without ditching the entire list altogether.

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: Yes, we do not reconcile disjoint intervals. That's out of scope for now, maybe in the future if we decide to do so but it complicates things significantly for no good reason. We can rename I guess

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: Ah in that case I misunderstood you.

Batching of course cannot prevent updates to the paginator itself, it's designed to batch updates to sibling collections. I'll update this then

isekovanic: should be resolved here: https://github.com/GetStream/stream-chat-js/pull/1822/commits/f7626b731971ccc24d7e4b56daba66566447ae31

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: renamed this to `removeReconciledItems`, the "from where" is clearly implied (as it's the state of the paginator).

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: Should be fixed here: https://github.com/GetStream/stream-chat-js/pull/1822/commits/679e8e417d9af06435e56474a800d3bb09c08400

MartinCupela (approved):

PR #1823, opened 2026-08-13
feat(search): adapt debounce to query length and cancel superseded requests

oliverlaz:
## CLA

- [x] Code changes are tested

Short search queries (1-2 characters) have low selectivity, so the backend `search` and `QueryChannels` endpoints are slow or time out on them. Two things made that worse than it needed to be:

1. **The debounce did not adapt to query length.** `debounce()` was built once in the constructor with a fixed 300ms, so a 1-character query fired just as eagerly as a precise one.
2. **Nothing was ever cancelled.** `cancelScheduledQuery()` only cancelled a *scheduled* call. An already-dispatched request ran to completion and wrote its result into state via a `finally` block. Worse, `canExecuteQuery` had a hard `!this.isLoading` guard, so a new search string typed while a request was in flight was **silently dropped** and the UI kept showing stale results.

### Dynamic debounce

`debounce()` in `src/utils.ts` now accepts a function for its timeout, resolved on every call right before `setTimeout`. That is the whole switching mechanism: each keystroke clears the pending timer and reschedules using the delay computed from *that* call's arguments.

| Option | Default | Applies to |
| --- | --- | --- |
| `shortQueryDebounceMs` | 500 | queries of at most `shortQueryMaxLength` characters |
| `longQueryDebounceMs` | 300 | longer queries |
| `shortQueryMaxLength` | 2 | the boundary, inclusive |

Typing `a` -> `ab` -> `abc` schedules at 500ms, 500ms, then 300ms, and only the last runs. An explicit `debounceMs` applies to **both** buckets, so anyone who already tuned it (say `debounceMs: 800`) keeps exactly their interval rather than being sped up to 500ms on short queries.

### Request cancellation

Each source holds one `AbortController`, aborted at the top of every dispatch. The state write stays in `finally` but is skipped when the signal is aborted, so a late response cannot clobber the newer query. `cancelScheduledQuery()` and `resetState()` abort as well.

I did **not** use `withCancellation` from `src/utils/concurrency.ts`: it serializes same-tag callbacks, which would make the new query queue behind the aborted one. A per-instance `AbortController` is a few lines and has no such coupling.

The signal reaches axios through a new optional trailing `ApiRequestOptions` on `search`, `queryUsers`, `queryChannels` (via the non-serialized `ChannelStateOptions`), `searchRoles`, `searchUserGroups` and `channel.queryMembers`, plus an optional `config` argument on `client.get`/`post`, which were previously dropping any axios config. `axios.isCancel` errors are no longer logged or counted toward the token-refresh backoff.

### Applies to

`MessageSearchSource`, `ChannelSearchSource`, `UserSearchSource`, `ChannelMemberSearchSource` and `MentionsSearchSource` all inherit this from the two base classes. `$autocomplete` / `$q` filter generation is unchanged. `CommandSearchSource` resolves locally, so it gets the dynamic debounce but no cancellation.

Short queries are **not** blocked, only debounced harder.

### Behaviour changes worth calling out for downstream SDKs

- Queries of at most 2 characters now wait 500ms instead of 300ms.
- **`canExecuteQuery` semantics changed**: a new search query now preempts an in-flight request instead of being dropped. Pagination still waits. This is a public method that `stream-chat-react` and custom sources both call and override.
- Custom `query()` implementations with their own `try`/`catch` will start seeing `CanceledError`.

All additions are optional parameters or widened optional fields, so this is a **minor** release, not a breaking one.

### Verification

`yarn lint`, `yarn types`, `yarn test` (3064 passing, 77 files) and `yarn build` are clean. 22 new tests cover the threshold switch in both directions, the config resolution rules including legacy `debounceMs`, stale-response discard, `isLoading` release on cancel, and the signal reaching axios.

## Changelog

- Search sources debounce short, low-selectivity queries harder than long ones, configurable via `shortQueryDebounceMs`, `longQueryDebounceMs` and `shortQueryMaxLength`
- A new search cancels the in-flight request instead of letting it finish, and no longer gets dropped while one is loading
- `search`, `queryUsers`, `queryChannels`, `searchRoles`, `searchUserGroups` and `channel.queryMembers` accept an abort signal

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1931,7 +1931,7 @@ export class StreamChat {
    * searchUserGroups - Search user groups by prefix for autocomplete
    *
    * @param {SearchUserGroupsOptions} options The search options
-   * @param requestOptions the api request options
+   * @param {RequestOptions} [requestOptions] Carries an abort signal. Not sent in the request.
    * @return {Promise<SearchUserGroupsResponse>} User Group Search Response
    */
   async searchUserGroups(
@@ -2078,7 +2078,7 @@ export class StreamChat {
    * @param {ChannelSort} [sort] Sort options, for instance {created_at: -1}.
    * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_updated: -1}, {created_at: 1}]
    * @param {ChannelOptions} [options] Options object. Can include predefined_filter, filter_values, and sort_values for using predefined filters.
-   * @param requestOptions the api request options.
+   * @param {RequestOptions} [requestOptions] Carries an abort signal. Not sent in the request.
    *
    * @return {Promise<QueryChannelsAPIResponse>} full search channels response
    */
--- a/src/client.ts
+++ b/src/client.ts
@@ -41,11 +41,11 @@ import {
 } from './utils';
 
 import type {
+  AbortOptions,
   ActiveLiveLocationsAPIResponse,
   AddUserGroupMembersOptions,
   AddUserGroupMembersResponse,
   APIErrorResponse,
-  ApiRequestOptions,
   APIResponse,
   AppIdentifier,
   AppSettings,
@@ -1340,7 +1340,9 @@ export class StreamChat {
   get<T>(
     url: string,
     params?: AxiosRequestConfig['params'],
-    config?: AxiosRequestConfig,
+    // `params` is omitted deliberately: _enrichAxiosOptions spreads `config` after the
+    // enriched params, so a `config.params` would drop api_key/user_id/connection_id.
+    config?: Omit<AxiosRequestConfig, 'params'>,
   ) {
     return this.doAxiosRequest<T>('get', url, null, { params, config });
   }
@@ -1349,7 +1351,7 @@ export class StreamChat {
     return this.doAxiosRequest<T>('put', url, data);
   }
 
-  post<T>(url: string, data?: unknown, config?: AxiosRequestConfig) {
+  post<T>(url: string, data?: unknown, config?: Omit<AxiosRequestConfig, 'params'>) {
     return this.doAxiosRequest<T>('post', url, data, { config });
   }
 
@@ -1845,7 +1847,7 @@ export class StreamChat {
    * @param {UserSort} sort Sort options, for instance [{last_active: -1}].
    * When using multiple fields, make sure you use array of objects to guarantee field order, for instance [{last_active: -1}, {created_at: 1}]
    * @param {UserOptions} options Option object, {presence: true}
-   * @param {ApiRequestOptions} [apiOptions] Request-level options such as an abort signal. Not sent in the request.
+   * @param {AbortOptions} [abortOptions] Carries an abort signal. Not sent in the request.
    *
    * @return {Promise<{ users: Array<UserResponse> }>} User Query Response
    */
@@ -1853,7 +1855,7 @@ export class StreamChat {
     filterConditions: UserFilters,
     sort: UserSort = [],
     options: UserOptions = {},
-    apiOptions: ApiRequestOptions = {},
+    abortOptions: AbortOptions = {},
   ) {
     const defaultOptions = {
       presence: false,
@@ -1877,7 +1879,7 @@ export class StreamChat {
           ...options,
         },
       },
-      apiOptions,
+      abortOptions,
     );
 
     this.state.updateUsers(data.users);
@@ -1929,17 +1931,17 @@ export class StreamChat {
    * searchUserGroups - Search user groups by prefix for autocomplete
    *
    * @param {SearchUserGroupsOptions} options The search options
-   * @param apiOptions the api request options
+   * @param abortOptions the api request options
    * @return {Promise<SearchUserGroupsResponse>} User Group Search Response
    */
   async searchUserGroups(
     options: SearchUserGroupsOptions,
-    apiOptions: ApiRequestOptions = {},
+    abortOptions: AbortOptions = {},
   ) {
     return await this.get<SearchUserGroupsResponse>(
       this.baseURL + '/usergroups/search',
       options,
-      apiOptions,
+      abortOptions,
     );
   }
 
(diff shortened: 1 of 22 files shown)


src/client.ts
```diff
@@ -1333,16 +1337,20 @@ export class StreamChat {
     }
   };
 
-  get<T>(url: string, params?: AxiosRequestConfig['params']) {
-    return this.doAxiosRequest<T>('get', url, null, { params });
+  get<T>(
+    url: string,
+    params?: AxiosRequestConfig['params'],
+    config?: AxiosRequestConfig,
```
MartinCupela: Shouldn't we omit params to prevent duplication?
```suggestion
    config?: Omit<AxiosRequestConfig, 'params'>,
```

src/client.ts
```diff
@@ -1333,16 +1337,20 @@ export class StreamChat {
...
+  ) {
+    return this.doAxiosRequest<T>('get', url, null, { params, config });
```
MartinCupela, line 1347: I could not find property `config` in `AxiosRequestConfig`, the last param of `doAxiosRequest`

src/client.ts
```diff
@@ -1837,13 +1845,15 @@ export class StreamChat {
...
    * @param {UserOptions} options Option object, {presence: true}
+   * @param {ApiRequestOptions} [apiOptions] Request-level options such as an abort signal. Not sent in the request.
    *
    * @return {Promise<{ users: Array<UserResponse> }>} User Query Response
    */
   async queryUsers(
     filterConditions: UserFilters,
     sort: UserSort = [],
     options: UserOptions = {},
+    apiOptions: ApiRequestOptions = {},
```
MartinCupela: Not sure if I would include word `api` in the naming as it could imply config params related to the Chat API, meanwhile these options are more about the request itself independent from the Chat API.

src/client.ts
oliverlaz: Good catch, and it is worse than just duplication: it silently breaks auth.

`_enrichAxiosOptions` spreads `...options.config` **after** the enriched `params` object, so a caller-supplied `config.params` replaces the whole thing rather than merging into it. Probed it on this branch before the fix:

```
params sent: {"mine":1}
api_key present? false
```

`user_id`, `connection_id` and `api_key` all disappear. Applied your suggestion, and the same to `post` for the identical reason (it has no separate `params` argument, so `config.params` would be the only way in and would clobber just as thoroughly). Fixed in 5b1ec134.

src/client.ts
oliverlaz: It is not on `AxiosRequestConfig` itself, it comes from the intersection on `doAxiosRequest`{'}s own parameter (client.ts:1275-1277):

```ts
options: AxiosRequestConfig & {
  config?: AxiosRequestConfig & { maxBodyLength?: number };
} = {},
```

So `{ params, config }` is valid: `params` lands on the top-level `AxiosRequestConfig` part and `config` on the added member. `sendFile` already uses the same shape.

That said, your confusion is fair and I think the type is the problem rather than the usage: `options` accepts axios config keys at the top level *and* a nested `config` of the same type, and `_enrichAxiosOptions` treats them differently (a top-level `options.signal` is silently ignored, only `options.config.signal` is honoured). Worth flattening, but it predates this PR so I left it alone here. Happy to open a follow-up if you want it cleaned up.

src/client.ts
oliverlaz: Agreed, and the point goes further than the Chat API: the bag is not necessarily request-related at all. `BaseSearchSource.query()` receives it too, and a custom source may resolve from local data and never issue a request.

Renamed to `AbortOptions` (params to `abortOptions`) in 5b1ec134, which also matches the existing `UploadRequestOptions` precedent of not prefixing transport concerns with "Api". Doc samples updated in GetStream/docs-content#1511.

oliverlaz:
Naming and the `config.params` hazard are both addressed; summary of what changed since review, since the branch history shows an intermediate name that did not survive.

| Thread | Outcome |
| --- | --- |
| `config.params` duplication | Fixed in 5b1ec134. It was worse than duplication: `_enrichAxiosOptions` spreads `config` after the enriched params, so a `config.params` replaced the whole object and dropped `api_key`, `user_id` and `connection_id`. Verified on the branch before the fix (`api_key present? false`). Applied your `Omit<AxiosRequestConfig, 'params'>` to `get`, and to `post` for the same reason. |
| `config` not on `AxiosRequestConfig` | No change needed, it comes from the intersection on `doAxiosRequest`'s own parameter (client.ts:1275-1277). The awkward shape predates this PR; happy to open a follow-up to flatten it. |
| `ApiRequestOptions` naming | Split in d25f63d2 into `SearchQueryOptions` for `query()` and `RequestOptions` for the client/channel methods. See the thread for why one name could not cover both. |

`yarn lint`, `yarn types`, `yarn test` (3064 passing) and `yarn build` are clean.

src/client.ts
oliverlaz: Correction to my reply above: `AbortOptions` did not survive, it was a bad name and I should not have picked one unilaterally. Settled in d25f63d2 as a split, because one type could not serve both roles honestly:

- **`SearchQueryOptions`** (declared in `src/search/BaseSearchSource.ts`) for a search source's `query()`. Your point goes further than the Chat API: a custom source may resolve from local data and never issue a request at all, so naming its options after a request was wrong regardless of the prefix. The doc comment now says ignoring the signal is valid.
- **`RequestOptions`** (in `types.ts`) for the client and channel methods, which do perform HTTP requests.

They could not share one name in any case: `client.search()` already takes a `SearchOptions`, so a `SearchQueryOptions` alongside it would have put two near-identical names on one signature. Parameters are `queryOptions` and `requestOptions`; sources forward the former into the latter where they do issue requests.

Doc samples follow in GetStream/docs-content#1511.

szuperaz (approved):
Didn't review the code as I've seen there are comments already, but tested with RN and was wokring fine for me.

MartinCupela (approved):

PR #1830, opened 2026-08-18
feat(i18n)!: share the translation runtime as stream-chat/i18n

oliverlaz:
Moves the translation runtime shared by `stream-chat-react` and `stream-chat-react-native` into this package as a new **`stream-chat/i18n`** subpath, plus **`stream-chat/i18n/codegen`** for the catalog generator. Both UI SDKs carried ~1,300 lines of near-duplicate runtime that had drifted apart, each reverse-mapping this package's English notification prose against its own hand-maintained table of sentences.

**Breaking, and shipped together deliberately** — the first commit is independently revertable if you'd rather split it:

- Notifications are keyed on `CORE_NOTIFICATION_TYPE` / `CoreNotificationType`. `Notification.message` is now documented as a developer-facing fallback whose wording is not contractual. Two identifiers renamed (`api:message{s}:query:failed` → `messageJumpFailed` / `messageJumpToLatestFailed`).
- Poll-composer field errors become `{ code, message, metadata? }` keyed on `POLL_VALIDATION_CODE`, instead of plain English strings.
- `engines.node` → `>=22.18.0` (the release that unflagged type stripping, which the `.mts` build scripts need). Node 18/20 are no longer tested — see the guide for what that means if you deploy the WS client there.
- `i18next` and `dayjs` become direct dependencies. They stay out of the root bundle; the subpath is what isolates them, and `assertBundleBoundaries` fails the build if that ever regresses.

Also here: the generator moves out of `src/` to `codegen/i18n/` (it reads the filesystem — not library source), ships ESM-only, and `scripts/bundle.mjs` becomes `.mts` with a real typecheck gate.

**Docs:** `v9-to-v10-migration-guide-i18n.md` (new), plus a Node-floor section in `-other.md`. Initiative record in `specs/i18n-to-core/`.

**Verified:** lint, all three `tsc` projects, 2,804 tests, and a clean-install check of the packed tarball across all four export conditions. The root bundle reaches neither `src/i18n/`, `i18next` nor `dayjs` — machine-checked, not reviewed.

Consumer PRs, both blocked on this publishing as `10.0.0-rc.3`: GetStream/stream-chat-react#3271 · GetStream/stream-chat-react-native#3777


src/messageComposer/middleware/pollComposer/validation.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Stable identifiers for poll-composer field validation failures.
+ *
+ * Same `domain:entity:operation:result` convention as `CORE_NOTIFICATION_TYPE`, but these are *field*
+ * errors rendered inline next to an input, so they deliberately do not go through
+ * `NotificationManager` — that would surface a toast per keystroke.
+ *
+ * **These values are public API.** UI SDKs key their translation tables on them, so renaming one is a
+ * breaking change.
+ */
+export const POLL_VALIDATION_CODE = {
```
MartinCupela: I would maybe change the name to be more specific what is the code related to. Probably it would make the purpose clearer if we included word 'notification' in the variable name.

test/unit/i18n/Streami18nGuarantees.test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+/** Core has no catalog, so every instance is handed the fixture's bundled defaults. */
+const setup = (options: Record<string, unknown> = {}) =>
+  new Streami18n<FixtureCatalog, FixtureBundledKey>({
+    logger: () => {},
+    runtimeDefaults: fixtureRuntimeDefaults,
+    ...options,
+  });
+
+describe('G1 — bundled defaults are layered under every language', () => {
```
MartinCupela, line 34: I would remove the agent's jargon e.g. G1

src/messageComposer/middleware/pollComposer/validation.ts
oliverlaz: Fixed in: https://github.com/GetStream/stream-chat-js/pull/1830/commits/4ab62e0da573cc88e41f9320cb50c80136d03b00

src/i18n/Streami18n.ts
```diff
@@ -0,0 +1,509 @@
...
+
+      this.state.partialNext({
+        initialized: true,
+        ...(this.tOverridden
+          ? {}
+          : { t: t as unknown as StreamTFunctionFor<C, Bundled> }),
+      });
+    } catch (error) {
+      this.logger(`Streami18n: initialization failed: ${describeError(error)}`);
+      this.state.partialNext({ initialized: true });
```
isekovanic: Is this intentional ? I think that `registerTranslations` on the instance later on is going to fail (probably other methods too) in that case

src/i18n/Streami18n.ts
```diff
@@ -0,0 +1,509 @@
...
+
+  /**
+   * Initializes i18next. Idempotent and safe to call concurrently.
+   *
+   * Memoized on success, so two independent consumers (a chat root and its overlay host) share one
+   * initialization. Cleared on rejection, so a retry is possible rather than the instance staying
+   * uninitialized for the process lifetime.
+   */
+  init(): Promise<Streami18nState<C, Bundled>> {
+    this.initPromise ??= this.runInit().catch((error: unknown) => {
```
isekovanic: I think the `.catch()` here will never run, as `runInit()` swallows all potential errors (and does not reflect this through `initialized` or anything else)

isekovanic (approved):

isekovanic (approved):

PR #1832, opened 2026-08-18
feat: remove handwritten upload methods; align method names

szuperaz:
## CLA

- [ ] Code changes are tested

https://linear.app/stream/issue/REACT-1084/remove-uploadfile-and-uploadimage-and-use-generated-method-instead

- remove handwritten `client.uploadImage_` and `client.uploadFile_` methods
- Rename `channel.sendImage` and `channel.sendFile` to `upload*` to match client method names
- Align method signature to rest of the API: `{ file, name }` instead of positional args

## Changelog

-


isekovanic (approved):

src/types.ts
```diff
@@ -981,8 +981,37 @@ export type UpdateMessageAPIResponse = StreamResponse<UpdateMessageResponse>;
...
 export type TranslationLanguage = TranslateMessageRequest['language'];
 
+/**
+ * A file descriptor, as produced by React Native image/document pickers. RN's `FormData`
+ * reads `uri`, `name` and `type` off this object; there is no `Blob` with a real body behind
+ * a picker URI, which is why the MIME type has to be carried explicitly.
+ *
+ * `name` falls back to the last path segment of `uri` when omitted.
+ */
+export type StreamFileDescriptor = { uri: string; name?: string; type?: string };
```
MartinCupela: @szuperaz there is `FileReference` type in messageComposer/types.ts that serves the same purpose. We should reconciliate

src/utils.ts
```diff
@@ -100,26 +98,57 @@ export const channelTracksReadLocally = (channel?: Channel) =>
...
-    if (name) data.append('file', uri, name);
-    else data.append('file', uri);
-  } else {
-    // React Native path
-    data.append('file', {
-      uri,
-      name: name || uri.split('/').reverse()[0],
-      contentType: contentType || undefined,
-      type: contentType || undefined,
+const isFileDescriptor = (value: unknown): value is StreamFileDescriptor =>
```
MartinCupela: @szuperaz I would move all the file related utils into a dedicated utils.file.ts or file.utils.ts file so that we do not have one long utils.ts made of a mix of functionalities

src/types.ts
szuperaz: We can't fully consolidate them because `FileReference` requires more fields, but I simplified it a bit here: 868fe82c2c0d22853631fa8f886403271e7af774

src/utils.ts
szuperaz: Fixed in d0af0a337db3b3379e897c8fef5627e22ce8704b

src/types.ts
MartinCupela: Thank you

src/utils.ts
MartinCupela: Thank you

MartinCupela (approved):

PR #1833, opened 2026-08-18
feat: move to WS API v2

szuperaz:
https://linear.app/stream/issue/REACT-1086/move-to-ws-connect-api-v2

## CLA

- [ ] Code changes are tested

WS API v2 sends different WS event shapes, so we need to use that WS endpoint to receive WS events where custom data is sent in `event.channel.custom` object.

No SDK change is required; tested both with React and RN, and regular and anon users.

Claude flagged some backend issues: https://getstream.slack.com/archives/C07AL9Q0T2T/p1787086052824189 - only one has a practical impact: the connected user's channel/user mutes are not returned on WS connect, so that will be a known regression until fixed.

Half related to this PR, but guest creation was broken, that is fixed too.

## Breaking changes

- WS API v2 doesn't have long polling implemented, so WS fallback is removed
- API v1 allowed providing a device ID for client connection, but it wasn't documented, nor was it used by RN SDK/sample apps; that functionality is removed (integrators can register a device in a separate API call)


MartinCupela:
Is this a fix of feat?

MartinCupela:
> Is this a fix of feat?

Or feat!:

src/types.ts
```diff
@@ -479,7 +475,35 @@ type LocalEvent = (
...
+export type ConnectedEvent = {
+  type: 'connection.ok';
+  connection_id: string;
+  created_at: Date;
+  me: OwnUserResponse;
+  /**
+   * Duplicates the mute and unread fields already present on `me`. Prefer `me`: it is
+   * the same shape the v1 `health.check` carries, and unlike `chat` it is decoded.
+   */
+  chat?: {
```
MartinCupela: Why is this added?

szuperaz: It's technically part of the API response,  but I removed it in 6e19bb41aef4094cc35d6e9427eb47234e84737d as `event.me` already has the same data and since we don't decode it, it's better not to expose. It should all be irrelevant if backend can include `ConnectedEvent` in API spec

MartinCupela (approved):

PR #1834, opened 2026-08-18
feat: channel state migration

isekovanic:
## CLA

- [x] Code changes are tested

### What

Converges a channel's per-domain reactive stores into **one** reactive state — `channel.state`, a `StateStore<ChannelStateData>` — subscribed to exactly like `thread.state`:

```ts
useStateStore(channel.state, (s) => ({ read: s.read }));
```

The per-domain `*Store` handles (`readStore`, `typingStore`, `membersStore`, `watcherStore`, `ownCapabilitiesStore`) are **removed** in favour of this single store, and the state surface is extended with the channel-level slices the UI SDKs previously hand-rolled in React (`data`, `membership`, `muteStatus`, lifecycle flags, `aiState`, `active`). `messagePaginator` and friends stay separate.

This targets `release-v10`.

### Why

Consumers had to know which of ~6 sub-stores held a given field and subscribe to each individually; channel-level state (mute status, membership, AI indicator) lived outside the reactive system entirely, forcing every UI SDK to re-derive it with bespoke `client.on(...)` listeners and local React state. One flat, reactive `channel.state` lets a consumer subscribe to any slice through a single selector and deletes that per-SDK plumbing.

### How

`ChannelState` now `extends StateStore<ChannelStateData>` (not composition — the store's `protected` members mean a wrapper isn't assignable to `StateStore<T>`, so `useStateStore(channel.state, …)` wouldn't typecheck otherwise). `ChannelStateData` is **flat**, with the same top-level keys the old per-store shapes used, so existing selectors stay contravariantly assignable and **do not need retyping**:

```ts
type ChannelStateData = WatcherState &     // watchers, watcher_count
  TypingUsersState &                       // typing
  ReadState &                              // read
  MembersState &                           // members, member_count
  MembershipState &                        // membership             (new)
  OwnCapabilitiesState &                   // ownCapabilities
  ChannelDataState &                       // data                   (new)
  MuteStatusState &                        // muteStatus             (new)
  ChannelLifecycleState &                  // initialized/offlineMode/pendingDisposal (new)
  ChannelActivationState &                 // active                 (new)
  AIIndicatorState;                        // aiState                (new)
```

Convenience getters/setters (`channel.state.members`, `.read`, `.typing`, `.watchers`, `.member_count`, `.watcher_count`) are kept and now proxy the single store; all writes go through `partialNext`, so a single-key write never wipes sibling slices.

Touched: `src/channel_state.ts` (the unified store + slices), `src/channel.ts` (`active`/`activate()`/`deactivate()`, `_syncMuteStatus`, `_setOwnUnreadCount`, AI event handling + connection-loss reset), `src/client.ts` (`_reflectMutedChannelsToActiveChannels`, `_resetAIStateOnActiveChannels`, re-seed guard), `src/messageDelivery/MessageReceiptsTracker.ts` (subscribe via `subscribeWithSelector` over the `read` slice), `src/types.ts` (`AIState`/`AIStates`).

#### Breaking vs v9

| Before (v9) | After | Notes |
|---|---|---|
| `channel.disconnected` | `channel.pendingDisposal` | The flag is one-way and terminal — `_disconnect()` disposes the paginators, unregisters the subscriptions, and the client drops the channel from `activeChannels` right after; the instance is never reconnected, which the old name implied. The old name is **removed outright** (no deprecation alias — v10 is a major, so the rename lands in one step). Rename every read/write, and note the state **slice key** changed with it: `(s) => ({ pendingDisposal: s.pendingDisposal })`. |
| `AIState` incl. `AI_STATE_CHECKING_SOURCES` | reconciled to `AI_STATE_EXTERNAL_SOURCES`, plus `AI_STATE_IDLE` / `AI_STATE_STOP`; canonical `AIStates` const now exported | Soft — `(string & {})` keeps arbitrary strings assignable. |

#### Changes relative to earlier v10 pre-releases (not v9 → v10 breaking)

None of these ever shipped in v9, so they are listed for anyone tracking `release-v10` rather than as migration steps:

| Earlier v10 pre-release | Now |
|---|---|
| `useStateStore(channel.state.readStore \| typingStore \| membersStore \| watcherStore \| ownCapabilitiesStore, sel)` | `useStateStore(channel.state, sel)` — drop the `.<X>Store`, keep the selector verbatim |
| `channel.state.mutedUsersStore` | `client.mutedUsersStore` (muted **users** are client-global) |
| `MutedUsersState` type | removed |
| in-place `channel.data.member_count = n` / `.own_capabilities = [...]` synced to state | no longer syncs — reassign `channel.data = { ...channel.data, member_count: n }` (the `Object.defineProperty` accessors are gone) |

#### Additive

- **`channel.state`** — the single unified `StateStore<ChannelStateData>`.
- **New reactive slices:** `data` (reactive channel data POJO), `membership`, `muteStatus` (`{ muted, createdAt, expiresAt }`, from `client.mutedChannels`), `initialized` / `offlineMode` / `pendingDisposal`, `aiState`, `active`.
- **`channel.activate()` / `channel.deactivate()` / `channel.active`** — refcount-backed "a consumer is currently consuming this channel" flag (a `Channel` instance is shared across preview + open + threads, so it's refcounted, not last-writer-wins). Gates the list-hydrate re-seed suppression on reconnect. Deliberately carries no rendering semantics — the type is `ChannelActivationState`, not a UI state.
- **`AIStates`** const exported from the package.

#### Behavioral

- **AI indicator is now LLC-owned and reactive.** Driven from `ai_indicator.update` / `.clear` / `.stop` in `_handleChannelEvent` (`.stop` is newly honoured). It **auto-resets to `Idle` on connection loss** — transient/internet drop via the health-gated channel cleaning sweep (`clean()`), deliberate close (e.g. mobile backgrounding) via `client.closeConnection()` — so a stuck "Generating" can't outlive a lost socket. WS-driven only (no optimistic self-update); ephemeral `ai_indicator.*` events are not replayed on reconnect by design.
- **`muteStatus` publishes only on real change** (recomputed from `client.mutedChannels`; no churn on the frequent `health.check` fan-out) via `client._reflectMutedChannelsToActiveChannels()`.
- **The own unread count now lives in exactly one place.** `channel.state.unreadCount` was a plain (non-store) field kept in parallel with `read[ownUserId].unread_messages`; it is now a **derived getter** over that read slice, so the number `channel.countUnread()` returns can no longer drift from the number an unread badge reads. Consequence worth knowing: the own read row now carries the own-unread gating that only the counter had before — a message that isn't unread-worthy (silent / shadowed / from a muted user / muted channel) no longer bumps it, and neither does one arriving while `messagePaginator.isViewingLive` (the consumer is looking at the newest message and is about to mark it read). The row is seeded when a channel has none yet (never queried, or `query({ watch: false, state: false })`) so the count still accumulates there.
- **`unreadCount` / `read[me]` kept consistent on channel-wide resets** — `channel.truncated` and "all channels read" route through `_setOwnUnreadCount`, which writes the read row (guarded; only reconciles an existing entry).
- **Reconnect re-seed of an active channel is suppressed** — the list-hydrate seed is gated on `!c.active` (only in `hydrateActiveChannels`, never on `channel.query`).
- Teardown ordering hardened in `_disconnect` (subscriptions down before flipping `pendingDisposal`, since store writes during teardown would trip a `getClient()`-reading subscription).

#### Kept deliberately separate

`messagePaginator` (+ `unreadStateSnapshot` / live-view state), `pinnedMessagesPaginator`, `messageComposer`, and `channel.configState`.

#### Review follow-ups (addressed)

- `Channel._syncStateFromChannelData` wrapper **removed** — every call site already passed both arguments, so its default duplicated `ChannelState.syncStateFromChannelData`'s own.
- `ChannelUIState` → **`ChannelActivationState`**, and the `active` / `activate()` / `deactivate()` docs no longer describe UI ("mounted", "on-screen") — the client stays headless.
- `ChannelState.unreadCount` field → **derived getter** (single source of truth, see Behavioral above).
- `isDirectChannel` slice **dropped** — classifying 1:1 by `memberCount === 2` is an opinionated definition (products that treat a DM as a creation-time property would disagree, and a 4 → 3 → 2 group would flip). Consumers derive `memberCount === 2` themselves at no extra cost: `memberCount` publishes on the same rare events the slice did.
- `channel.disconnected` → **`channel.pendingDisposal`**, old name removed outright rather than deprecated (see Breaking vs v9).
- `currentReadStoreState` → `currentState` in `_patchReadState`; leftover `TODO #29` references removed.

#### Testing

Full suite green: **2740 passed / 1 todo**; `yarn types` clean; `yarn lint` clean; dist builds. Coverage added in `test/unit/channel_state.test.js`, `channel.test.js`, `client.test.js`, and `messageDelivery/MessageReceiptsTracker.test.ts` (handle-identity asserts dropped; sibling-preservation, subscribe-by-selector, the derived `unreadCount`, and the lifecycle / mute / AI / `active` slices covered).


src/channel.ts
MartinCupela: Do we need this method if we have this.state.syncStateFromChannelData? I suppose that `_` in the name means it should be private so probably no need to have this wrapper that would make sense in case of public API.

src/channel_state.ts
```diff
@@ -36,163 +38,280 @@ export type ReadState = {
...
+  active: boolean;
+};
+
+/**
+ * Reactive AI-indicator state — driven by the `ai_indicator.update` / `.clear` / `.stop` events (see
+ * `Channel._handleChannelEvent`). Seeded to `AIStates.Idle` and subscribable via
+ * `useStateStore(channel.state, (s) => ({ aiState: s.aiState }))`. Reset to `Idle` on unwatch /
+ * disconnect (a live server sends `ai_indicator.clear` when the AI response starts streaming).
+ */
+export type AIIndicatorState = {
```
MartinCupela, line 171: Could this be in ChannelUIState? it is also related to what is being displayed.

src/client.ts
```diff
@@ -1022,9 +1026,10 @@ export class StreamChat extends ChatApi {
 
     if (event.type === 'notification.mark_read' && event.unread_channels === 0) {
       const activeChannelKeys = Object.keys(this.activeChannels);
-      activeChannelKeys.forEach(
-        (activeChannelKey) =>
-          (this.activeChannels[activeChannelKey].state.unreadCount = 0),
+      activeChannelKeys.forEach((activeChannelKey) =>
+        // resets both `state.unreadCount` and `read[userId].unread_messages` so the badge, which
+        // reads the latter, does not stay stale (TODO #29).
```
MartinCupela: do we need to keep this todo?

src/channel_state.ts
```diff
@@ -36,163 +38,280 @@ export type ReadState = {
...
   _channel: Channel;
-  readonly watcherStore: StateStore<WatcherState>;
-  readonly typingStore: StateStore<TypingUsersState>;
-  readonly readStore: StateStore<ReadState>;
-  readonly membersStore: StateStore<MembersState>;
-  readonly ownCapabilitiesStore: StateStore<OwnCapabilitiesState>;
-  // todo: is this actually used somewhere?
-  readonly mutedUsersStore: StateStore<MutedUsersState>;
   pending_messages: Array<PendingMessageResponse>;
   unreadCount: number;
```
MartinCupela: Do we need to keep this variable if we have unread count in the read state?

src/channel_state.ts
```diff
@@ -36,163 +38,280 @@ export type ReadState = {
...
+        !Array.isArray(data.own_capabilities) &&
+        Array.isArray(fallbackData?.own_capabilities)
+      ) {
+        data.own_capabilities = [...fallbackData.own_capabilities];
+      }
+    }
+
+    this.partialNext({
+      data,
+      isDirectChannel: (memberCount ?? 0) === 2,
```
MartinCupela: Why do we need the isDirectChannel flag? I am thinking about a scenario where a group chat has 4 -> 3 -> 2 members. Originally it was a group chat (the creation intent was a group chat). Should we start considering it a 1:1 channel?

src/channel_state.ts
isekovanic: It probably can, but reading through it now I'm not happy with `ChannelUIState` in general and I don't think it should exist in the LLC at all. I will instead make it more explicit, because as far as the LLC is concerned the `active` state means something different.

The `AIIndicatorState` similarly CAN affect UI, but in reality is the state of the current AI agent within that `channel`. The LLC should be headless and not related to any UI semantics.

isekovanic: So TL;DR, I'll change this in a bit

src/client.ts
isekovanic: nah, it's a leftover

src/channel_state.ts
isekovanic: It should derive it directly as a shorthand, I can remove it if you prefer but doesn't really hurt I guess

MartinCupela:
In the PR description there is the following:

- [Breaking] Removed the per-domain channel.state.*Store handles (readStore, typingStore, membersStore, watcherStore, ownCapabilitiesStore); subscribe via useStateStore(channel.state, selector) instead.
- [Breaking] Removed channel.state.mutedUsersStore and the MutedUsersState type; use the client-global client.mutedUsersStore.
- [Breaking] In-place mutation of channel.data.member_count / channel.data.own_capabilities no longer syncs to state; reassign channel.data.

We should not mark this change as breaking as these have never made it to v9 (master).

src/channel_state.ts
isekovanic: There is often the need to react differently to a channel based on whether it's a direct one or not, hence it's quite useful to have a state that will not change very often (i.e on every member add/remove) but rather have it ready to be used.

Naturally, it can be also derived anywhere SDK side through `memberCount` and it will also stay stable, this just helps us skip extra steps

src/channel_state.ts
MartinCupela: I would prefer to have a single place to store a value to prevent having to sync multiple places.

isekovanic:
> In the PR description there is the following:
> 
> * [Breaking] Removed the per-domain channel.state.*Store handles (readStore, typingStore, membersStore, watcherStore, ownCapabilitiesStore); subscribe via useStateStore(channel.state, selector) instead.
> * [Breaking] Removed channel.state.mutedUsersStore and the MutedUsersState type; use the client-global client.mutedUsersStore.
> * [Breaking] In-place mutation of channel.data.member_count / channel.data.own_capabilities no longer syncs to state; reassign channel.data.
> 
> We should not mark this change as breaking as these have never made it to v9 (master).

it's breaking in terms of V10 of course and it can be useful if someone pulls stuff in so that they know what's changed, since V10 is anyway going to be squashed at the end (probably) I can't see why it matters

src/channel.ts
isekovanic: it's useful I guess so you don't need to pass `this.data` everywhere (what it defaults to) and so you're able to just `_syncStateFromChannelData(someNewData)` locally, but if you prefer I can also remove it, don't have strong feelings for it either way

src/channel.ts
```diff
@@ -148,27 +149,13 @@ export class Channel extends ChannelApi {
...
-  initialized: boolean;
-  /**
-   * Indicates whether channel has been initialized by manually populating the state with some messages, members etc.
-   * Static state indicates that channel exists on backend, but is not being watched yet.
-   */
-  offlineMode: boolean;
   lastKeyStroke?: Date;
   lastTypingEvent: Date | null;
   isTyping: boolean;
-  disconnected: boolean;
```
MartinCupela, line 166: What is actually this flag for? I see that we prevent state updates if it is enabled. In which scenario do we need to keep a reference to a disconnected instance?

src/channel_state.ts
isekovanic: that's fine then, I'll remove it

src/channel.ts
MartinCupela: What the flag buys you — four behaviors:

1. Fail loudly on stale usage. src/channel.ts:186 throws "You can't use a channel after client.disconnect() was called". Since nearly every channel method goes through getClient()/getConfig(), holding onto a channel across a disconnectUser() produces a clear error instead of silent requests on a client with no user.
2. Cache eviction guard. src/client.ts:2680 and the members-based lookup in getChannelByMembers (client.ts:2625) skip disconnected entries, so client.channel(...) mints a fresh instance rather than returning a dead one. This matters because the activeChannels deletion happens in a postListenerCallbacks deferral — the flag covers the window where the dead channel is still in the map.
3. Prevents re-watching on reconnect. Marking the channel dead is what stops recoverState from re-subscribing to a channel you were removed from or that was deleted (the reason cited in the removed_from_channel comment referencing #2599).
4. Offline DB guard. offline_support_api.ts:551 refuses to harvest channel.data from a disconnected instance, so stale data isn't written back into the local DB.

Note it's one-way: there's no _reconnect(). A disconnected Channel is garbage — the client always constructs a new one.

MartinCupela: I think the name is however misleading. I would maybe think about renaming it to `pendingDisposal`.

isekovanic: Yes so it's there precisely for those reasons. I've thought about removing it but I'm not sure what other issues it was solving (for example a stale reference to a `Channel` object sitting in the SDK or some integrator code while the channel itself is no longer "connected"). While a disconnected channel is indeed garbage, the flag is needed to prevent errors from being thrown upwards unreasonably whenever something inevitably disconnects it.

src/channel_state.ts
MartinCupela: I have doubts about this flag being misleading and opinionated - some integrators can approach the 1:1 channel classification in a strict way where channel is created as 1:1 (as in Slack, WhatsApp) and cannot be changed to a group channel.

MartinCupela:
> > In the PR description there is the following:
> > 
> > * [Breaking] Removed the per-domain channel.state.*Store handles (readStore, typingStore, membersStore, watcherStore, ownCapabilitiesStore); subscribe via useStateStore(channel.state, selector) instead.
> > * [Breaking] Removed channel.state.mutedUsersStore and the MutedUsersState type; use the client-global client.mutedUsersStore.
> > * [Breaking] In-place mutation of channel.data.member_count / channel.data.own_capabilities no longer syncs to state; reassign channel.data.
> > 
> > We should not mark this change as breaking as these have never made it to v9 (master).
> 
> it's breaking in terms of V10 of course and it can be useful if someone pulls stuff in so that they know what's changed, since V10 is anyway going to be squashed at the end (probably) I can't see why it matters

Because as an integrator you are looking at a diff v9 - v10. If we say there is a breaking change that is actually not a breaking change, it is misleading. I would just not put these into the description of the commit that will be created with this PR being merged.

src/channel.ts
MartinCupela: I think that if we use it internally only, then no need to duplicate the naming - makes stuff harder to read, if you are not familiar with the SDK.

src/channel_state.ts
isekovanic: I don't really think it's opinionated nor misleading, as then you could say this for any part of the state really that relies on business logic for integrators - but I can remove it since this:

> I am thinking about a scenario where a group chat has 4 -> 3 -> 2 members

probably doesn't make sense in that case

isekovanic:
> > > In the PR description there is the following:
> > > 
> > > * [Breaking] Removed the per-domain channel.state.*Store handles (readStore, typingStore, membersStore, watcherStore, ownCapabilitiesStore); subscribe via useStateStore(channel.state, selector) instead.
> > > * [Breaking] Removed channel.state.mutedUsersStore and the MutedUsersState type; use the client-global client.mutedUsersStore.
> > > * [Breaking] In-place mutation of channel.data.member_count / channel.data.own_capabilities no longer syncs to state; reassign channel.data.
> > > 
> > > We should not mark this change as breaking as these have never made it to v9 (master).
> > 
> > 
> > it's breaking in terms of V10 of course and it can be useful if someone pulls stuff in so that they know what's changed, since V10 is anyway going to be squashed at the end (probably) I can't see why it matters
> 
> Because as an integrator you are looking at a diff v9 - v10. If we say there is a breaking change that is actually not a breaking change, it is misleading. I would just not put these into the description of the commit that will be created with this PR being merged.

You're actually just looking at a changelog, which will contain all changes that are relevant and breaking :D 

And also a migration guide. 

But I'll remove them

src/channel.ts
```diff
@@ -2111,7 +2202,7 @@ export class Channel extends ChannelApi {
     let hasStateChanged = false;
     this.messageReceiptsTracker.setPendingReadStoreReconcileMeta(reconcileMeta);
 
-    this.state.readStore.next((currentReadStoreState) => {
+    this.state.next((currentReadStoreState) => {
```
MartinCupela:
```suggestion
    this.state.next((currentState) => {
```

The `currentReadStoreState` sounds misleading as it contains the whole state not only read state.

src/channel_state.ts
MartinCupela: It is opinionated in the sense, that if you want to apply the Slack, WhatsApp approach (different approach / opinion), then you cannot rely on this flag.

src/channel.ts
isekovanic: How so ? What's harder to read ? It's the exact same thing but as a wrapper around the original so that we don't need to keep passing the fallback all the time.

src/channel.ts
MartinCupela: I think that `pendingDisposal` signals, what will happen with the channel, so the developers will naturally avoid touching such an instance. Meanwhile `disconnected` makes impression of something that can be reconnected and thus to be kept used.

src/channel.ts
MartinCupela: I see the default value provided to `fallbackData` also in the channel state's class. So it seems there would not be added value to `_syncStateFromChannelData`. Maybe I am missing something.

Image

MartinCupela:
> > > > In the PR description there is the following:
> > > > 
> > > > * [Breaking] Removed the per-domain channel.state.*Store handles (readStore, typingStore, membersStore, watcherStore, ownCapabilitiesStore); subscribe via useStateStore(channel.state, selector) instead.
> > > > * [Breaking] Removed channel.state.mutedUsersStore and the MutedUsersState type; use the client-global client.mutedUsersStore.
> > > > * [Breaking] In-place mutation of channel.data.member_count / channel.data.own_capabilities no longer syncs to state; reassign channel.data.
> > > > 
> > > > We should not mark this change as breaking as these have never made it to v9 (master).
> > > 
> > > 
> > > it's breaking in terms of V10 of course and it can be useful if someone pulls stuff in so that they know what's changed, since V10 is anyway going to be squashed at the end (probably) I can't see why it matters
> > 
> > 
> > Because as an integrator you are looking at a diff v9 - v10. If we say there is a breaking change that is actually not a breaking change, it is misleading. I would just not put these into the description of the commit that will be created with this PR being merged.
> 
> You're actually just looking at a changelog, which will contain all changes that are relevant and breaking :D
> 
> And also a migration guide.
> 
> But I'll remove them

You mean breaking btw v10 rc.2 and v10 rc.x?

src/channel.ts
isekovanic: removed

src/channel.ts
```diff
@@ -1357,6 +1367,100 @@ export class Channel extends ChannelApi {
...
+  get pendingDisposal() {
+    return this.state.getLatestValue().pendingDisposal;
+  }
+
+  set pendingDisposal(pendingDisposal: boolean) {
+    this.state.partialNext({ pendingDisposal });
+  }
+
+  /**
+   * @deprecated Renamed to {@link Channel.pendingDisposal} — the flag is one-way and the instance is
```
MartinCupela: Maybe we could directly remove the methods / symbols instead of deprecating them as we are doing breaking changes anyway.

isekovanic: yes agreed

src/channel.ts
```diff
@@ -311,7 +300,7 @@ export class Channel extends ChannelApi {
    * @returns The chat client.
    */
   getClient(): StreamChat {
-    if (this.disconnected === true) {
+    if (this.pendingDisposal === true) {
       throw Error(`You can't use a channel after client.disconnect() was called`);
```
MartinCupela: I would re-word the error message to something like: "Channel instance is pending disposal and cannot be used". Not sure whether mentioning client.disconnect is a good thing.

isekovanic: bah, I forgot this one naturally, thanks

src/channel.ts
```diff
@@ -1501,6 +1614,9 @@ export class Channel extends ChannelApi {
   override async stopWatching(...args: Parameters<ChannelApi['stopWatching']>) {
     const response = await super.stopWatching(...args);
 
+    // Deliberate: unlike a connection loss this must NOT be restored on reconnect.
+    this.watchStatus = ChannelWatchStatus.NotWatching;
```
MartinCupela, line 1578:
```suggestion
    this.watchStatus = ChannelWatchStatus.WatchingStopped;
```

NotWatching semantically subsumes also WasWatching. So I was thinking we could express the intentional aspect of not watching by saying it was stopped.

isekovanic: Don't think so, because `NotWatching` also means "never started watching".

`WatchingStopped` essentially means "we were watching but we stopped", which is not always going to be the case

MartinCupela: Reading the comment on the value meaning I realized it is state 0 and 3 combined. So then NotWatching is ok with me.

MartinCupela (approved):

PR #1836, opened 2026-08-19
feat!: remove hand-written types and API wrappers

szuperaz:
## Breaking changes
- Several type renames (hand-written types replaced with generated ones)
- Moderation endpoints migrated from hand-written API v1 to generated API v2
- Permission v1 remainders removed

The full list of changes is in the migration guides

- Remove hand-written API types
- Migrate the few hand-written API calls to generated API -> mainly moderation endpoints
- Removed 3 "proxy" methods from `channel`: `channel.getReplies`, `channel.search` and `channel.getReactions` -> technically none of these have a channel cid param, but having them on `channel` is handy, but my logic here was that if we start adding these methods to `channel`, it's something we have to maintain manually when new API calls introduced by regeneration
- Removed the permission v1 remainders

https://linear.app/stream/issue/REACT-1066/remove-unnecessary-types-from-typests 
https://linear.app/stream/issue/REACT-1065/check-poll-types-in-stream-chat-js
https://linear.app/stream/issue/REACT-1087/check-few-api-methods-bypass-the-generated-layer


src/gen/chat/ChannelApi.ts
```diff
@@ -316,6 +318,37 @@ export class ChannelApi {
...
+      pinned_at_before_or_equal?: Date;
+      pinned_at_before?: Date;
+      id_around?: string;
+      pinned_at_around?: Date;
+      sort?: Array<SortParamRequest>;
+      member_custom_include?: Array<string>;
+    },
+    requestOptions?: StreamRequestOptions,
+  ): Promise<StreamResponse<GetPinnedMessagesResponse>> {
+    if (!this.id) {
```
MartinCupela, line 340: Is it possible that `id` would be undefined? Can we create a Channel instance without an `id`?

src/gen/chat/ChannelApi.ts
```diff
@@ -316,6 +318,37 @@ export class ChannelApi {
     );
   }
 
+  getPinnedMessages(
+    request?: {
```
MartinCupela, line 322: Would it make sense to create a type for this request object?

MartinCupela (approved):

oliverlaz (approved):

isekovanic (approved):

src/gen/chat/ChannelApi.ts
szuperaz: Yes, it is, it's basically a simplified version of the `_checkInitialized` check https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L2398. This generated code was made with node SDK in mind, for stream-chat-js we need to call `_checkInitialized` here, so added a linear task to do this: https://linear.app/stream/issue/REACT-1173/generator-allow-overriding-initalized-check

src/gen/chat/ChannelApi.ts
szuperaz: I don't think it adds too much value, it's nicer when we have a named type for an API request, but since the generator can't generate one here (it comes down to how api spec defines an endpoint), I don't think the manual work is worth it (adding it for a single endpoint of course is not a big deal, but what I mean is that I don't think it's worth creating a pattern where we add a named type for all endpoints like this, and keep adding new types when a similar endpoint is added to code).

PR #1841, opened 2026-08-25
feat: connection recovery

isekovanic:
## CLA

- [x] Code changes are tested

The client handles connection recovery now, so the UI SDKs don't have to.

`recoverState()` was unusable, because of the fact that it never guaranteed proper rewatching of channels (nor reordering of the channel list properly and so we had an implementation that ended up spread across three places that didn't know about each other.

What was wrong with the old one:

- one `queryChannelsAndHydrate({ cid: { $in: activeChannels }, limit: 30 })`, so it used a query shape no list actually had, and quietly stopped at 30 channels
- never looked at `watchStatus`
- wasn't ordered against offline pending-task replay or `sync()`
- only called from `_reconnect()`, so it never ran when the app backgrounded (`closeConnection()` → `openConnection()`), which is the common case on mobile

`ChannelWatchStatus.WasWatching` and `channel.active` were already there and read by nothing. They were
added for this.

### What recovery does

`ConnectionRecoveryManager` does three things:

1. every loaded channel list re-runs its own first-page query (`ChannelManager.recover()`). Since `queryChannels` watches by default, that re-watches the channels on that page too. Non-destructive, the list never blanks.
2. every active channel reloads (`channel.reload()`). A list page carries far fewer messages per channel than an open channel's window, so the list query can't cover it.
3. every active thread reloads its replies (`thread.reload()`). 

It doesn't loop over `activeChannels`. That cache gets large after any scrolling and watches are a limited server-side resource. Channels outside the refreshed pages come back when an event proves them relevant: `ChannelManager` re-watches a routed channel marked `WasWatching`.

Triggers differ per surface, so lists use `connection.changed { online: true }`, because `ChannelPaginator.executeQuery` already defers its own first page. Channels and threads use the offline DB's sync-status edge case when offline support is on and `connection.changed` otherwise, because `reload()` has no deferral of its own and has to run after replay and `sync()`. `OfflineDBSyncManager` publishes that edge unconditionally after `syncAndExecutePendingTasks()`, which is what makes the ordering hold on every path.

`connection.recovered` now fires from the manager once the reloads settle, on every reconnect path rather than only from `_reconnect()`.

### Bug fixes

Three, all pre-existing, found while testing the above. Each has regression tests.

**Reload asked for as many items as were already loaded.** A thread with one reply asked the server for one reply, so anything that arrived while offline couldn't be discovered, and the single result was disjoint from the loaded window so the merge rebuilt and dropped what was there. Asks for at least a page now.

**`mergeNewestPage` returned early when there was no anchored head interval,** throwing away the page it had just been handed. Happens when the first query comes back empty and the only content was ingested live, i.e. a thread or channel created in the same session. It anchors the page now, and doesn't move the view if the reader has jumped somewhere else.

**Opening a thread on a parent with no replies 404s,** since there's no thread server-side yet. That was published as `lastReloadError`, and RN ORs that into its channel error state, so a brand-new thread looked broken. Treated as the expected answer now: `reload()` resolves quietly. A thread that *did* have replies and comes back 404 still reports it. `replyCount` tells the two apart, and unlike `deletedAt` it still works when you missed the delete event while offline.

### Breaking

`client.recoverState()` is gone rather than a no-op, so callers fail loudly. Use. `client.connectionRecovery.recover()`. The one thing it did that mattered, resetting `wsPromise`/`setUserPromise`, moved to `client._settleConnectPromises()` and is called from `_reconnect()` in the same place.

`recoverStateOnReconnect` keeps its name, type and `true` default. It gates the new flow instead of the old bulk query. If you set it `false` and recover yourself, nothing changes.

### Added

`client.connectionRecovery` · `ChannelManager.recover()` · `channel.state.lastReloadError` (plus the `channel.lastReloadError` getter) · `thread.state.lastReloadError` · `isDoesNotExistError` in `errors.ts`. `Thread.activate()`/`deactivate()` are refcounted now, like `Channel.activate()`.

The `lastReloadError` fields exist because the manager runs reloads inside `Promise.allSettled`, so a failure never reaches the UI as a throw. Cleared on entry, set in the catch, and still rethrown.

### Known gaps

Recovery finds threads via `client.threads.threadsById` filtered on `state.active`. That's the thread *list*, not a cache - a thread opened from a message list only gets there because the UI SDKs adopt it once its replies load. A reconnect before that, or after `ThreadManager.reload()` evicts a thread the user doesn't participate in, misses it. There's a test named `KNOWN GAP` pinning this. The fix is `ThreadManager`'s commented-out `threadCache` (probably in the form of a `StoreBackedItemIndex` or something) that is anyway in our todo list.

Underneath that: live-ingested content produces a window with no head interval at all. It renders fine and `hasMoreHead` says `true`, which isn't true. Three places each work around it differently (`mergeNewestPage` got it wrong, `jumpToTheLatestMessage` queries to recover, `reconcileHeadAgainstPage` returns early). Anchoring at ingest time instead was tried and reverted: it breaks optimistic message placement, since an unsent message would claim it came from a fetched page, and it stops scroll-to-bottom loading the real head. The fix within the page merging mechanism should still stay for a variety of different edge cases, but this one should also be addressed. Needs its own PR.

`isDoesNotExistError` looks the code up in `APIErrorCodes` by name instead of comparing to `16`, so the table stays the only place that mapping lives. `name` is typed `string` there, so a typo compiles and silently returns false. Making those names a union would be nice.

## Changelog

- Added `client.connectionRecovery`. On reconnect it re-runs each loaded channel list's first-page query,reloads every active channel and thread, then dispatches `connection.recovered`. Runs after offline replay and `sync()` on every reconnect path, including backgrounding.
- Added `ChannelManager.recover()`, a non-destructive re-query of every initialized list.
- Added `channel.state.lastReloadError` (and `channel.lastReloadError`) and `thread.state.lastReloadError` so a failed reconnect refresh stays visible to the UI.
- Added `isDoesNotExistError` to `errors.ts`.
- `ChannelManager` re-watches a routed channel whose `watchStatus` is `WasWatching`, restoring a watchlost with the socket. Narrower than v9, which re-watched any routed channel. Skipped for`channel.hidden`, pending disposal and `NotWatching`.
- `connection.recovered` is dispatched by `ConnectionRecoveryManager` after the reloads, so it fires onevery reconnect path instead of only from `_reconnect()`.
- `Thread.activate()`/`deactivate()` are refcounted, matching `Channel.activate()`.
- Fixed the reconnect refresh requesting only as many items as were loaded, which hid messages that arrived while offline and could drop the loaded window.
- Fixed `mergeNewestPage` discarding the fetched page when no page had ever anchored the window.
- Fixed opening a thread on a reply-less parent raising an error state.
- [BREAKING CHANGE] Removed `client.recoverState()`. Use `client.connectionRecovery.recover()`.
- [BREAKING CHANGE] `recoverStateOnReconnect` now gates `client.connectionRecovery` instead of the removed bulk query. Same name, type and default.

### DIFF
--- a/src/ChannelManager.ts
+++ b/src/ChannelManager.ts
@@ -15,15 +15,13 @@ import type {
 import { getChannel } from './pagination/utility.queryChannel';
 import type { Channel } from './channel';
 import { ChannelWatchStatus } from './channel_state';
-import { chatLoggerSystem } from './logger';
+import { runDetached } from './utils';
 
 export type ChannelManagerEventHandlerContext = {
   channelManager: ChannelManager;
 };
 
 type EventHandlerContext = ChannelManagerEventHandlerContext;
-
-const logger = chatLoggerSystem.getLogger('channel');
 
 type SupportedEventType = EventType | (string & {});
 
@@ -159,10 +157,8 @@ const restoreInterruptedWatch = (channel: Channel, client: StreamChat) => {
   // Takes the client as an argument rather than calling `channel.getClient()`, which THROWS for a
   // channel pending disposal — the guard above makes that unreachable today, but a throw here would
   // reject the whole event handler, so it is not a hazard worth leaving one edit away.
-  getChannel({ channel, client }).catch((error) => {
-    logger
-      .withExtraTags('restoreInterruptedWatch', channel.cid)
-      .warn('Failed to restore an interrupted channel watch.', { error });
+  runDetached(getChannel({ channel, client }), {
+    context: `restoreInterruptedWatch(${channel.cid})`,
   });
 };
 
--- a/src/ChannelManager.ts
+++ b/src/ChannelManager.ts
@@ -14,12 +14,16 @@ import type {
 } from './EventHandlerPipeline';
 import { getChannel } from './pagination/utility.queryChannel';
 import type { Channel } from './channel';
+import { ChannelWatchStatus } from './channel_state';
+import { chatLoggerSystem } from './logger';
 
 export type ChannelManagerEventHandlerContext = {
   channelManager: ChannelManager;
 };
 
 type EventHandlerContext = ChannelManagerEventHandlerContext;
+
+const logger = chatLoggerSystem.getLogger('channel');
 
 type SupportedEventType = EventType | (string & {});
 
@@ -134,6 +138,34 @@ export const ignoreEventsForUnknownChannels: EventHandlerPipelineHandler<
   if (!channel) return { action: 'stop' };
 };
 
+/**
+ * Restore a watch this client held and lost to a dropped socket.
+ *
+ * The server keys watches by connection ID, so a reconnect ends every watch the previous connection
+ * held. Reconnect recovery re-watches the first page of each list plus the active channel; everything
+ * else stays `WasWatching` — pages 2+ of a scrolled list, channels visited and navigated away from,
+ * channels matching no mounted list. Such a channel still receives member-level events (e.g.
+ * `notification.message_new`), which relocate its row but carry no message body, so its preview would
+ * sit frozen until the channel was opened. Re-watching it the moment an event proves it relevant is
+ * what closes that gap, without ever eagerly re-watching the whole cache.
+ *
+ * The request is idempotent, so a successful watch flips the status to `Watching`, and `getChannel`
+ * dedupes concurrent watches for the same cid, so a burst of events cannot produce a burst of requests.
+ */
+const restoreInterruptedWatch = (channel: Channel, client: StreamChat) => {
+  if (channel.pendingDisposal) return;
+  if (channel.watchStatus !== ChannelWatchStatus.WasWatching) return;
+
+  // Takes the client as an argument rather than calling `channel.getClient()`, which THROWS for a
+  // channel pending disposal — the guard above makes that unreachable today, but a throw here would
+  // reject the whole event handler, so it is not a hazard worth leaving one edit away.
+  getChannel({ channel, client }).catch((error) => {
+    logger
+      .withExtraTags('restoreInterruptedWatch', channel.cid)
+      .warn('Failed to restore an interrupted channel watch.', { error });
+  });
+};
+
 const updateLists: EventHandlerPipelineHandler<EventHandlerContext> = async ({
   event,
   ctx: { channelManager },
@@ -183,6 +215,13 @@ const updateLists: EventHandlerPipelineHandler<EventHandlerContext> = async ({
     // (`paginator.boost`) for integrators to opt into for specific channels (VIP/mention/deep-link).
     paginator.ingestItem(channel);
   });
+
+  // AFTER routing, and not awaited: the row relocates immediately off the event, and the watch (plus
+  // the state it hydrates) lands whenever it lands. `channel.hidden` is excluded — a channel being
+  // hidden is the one routed event that must not resurrect a watch.
+  if (event.type !== 'channel.hidden') {
+    restoreInterruptedWatch(channel, channelManager.client);
+  }
 };
 
 // we have to make sure that client.activeChannels is always up-to-date
@@ -720,5 +759,36 @@ export class ChannelManager extends WithSubscriptions {
       this.paginators.map(async (paginator) => {
         await paginator.reload();
       }),
+    );
+
+  /**
+   * Re-run every loaded list's own first-page query — the channel-list half of connection recovery,
+   * driven by {@link ConnectionRecoveryManager}.
+   *
+   * Each list re-asserts its *own* `filters` / `sort` / `pageSize` rather than some substitute query,
+   * and because `client.queryChannels` watches by default, that page's channels are re-watched as a
+   * side effect. Channels below the first page are deliberately not touched — see
+   * `restoreInterruptedWatch` for how they come back.
+   *
+   * `{ keepPreviousItems: true, reset: 'yes' }` is the non-destructive refresh: `reset` restarts the
+   * window at page 1, while `keepPreviousItems` merges the fetched page into the loaded list instead
+   * of replacing it and keeps the item index populated — the latter is what stops a concurrent
+   * offline-replay ingest from rebuilding the list out of a cleared index. Notably NOT `silent`: the
+   * `isLoading` transition it would suppress is the in-flight guard that keeps a second overlapping
+   * query from being issued.
+   *
+   * Distinct from {@link ChannelManager.reload}, which is a destructive `reset` that blanks each list
+   * to its loading state.
+   *
+   * Paginators that were never queried are skipped: they run their own first query when they mount,
+   * and querying them here would race it.
+   */
+  recover = async () =>
+    await Promise.allSettled(
+      this.paginators
+        .filter((paginator) => paginator.isInitialized)
+        .map(async (paginator) => {
+          await paginator.toTail({ keepPreviousItems: true, reset: 'yes' });
+        }),
     );
 }
--- a/src/ConnectionRecoveryManager.ts
+++ b/src/ConnectionRecoveryManager.ts
@@ -212,10 +212,10 @@ export class ConnectionRecoveryManager extends WithSubscriptions {
         `Recovering ${channels.length} active channel(s) and ${threads.length} active thread(s).`,
       );
 
-    // allSettled: one channel or thread failing must not stop the others. Each failure is published
-    // on `channel.state.lastLoadError` / `thread.state.lastLoadError` by the reload itself, so it
-    // is not lost here. Channels and threads are independent requests, so they run together rather
-    // than in sequence.
+    // allSettled: one channel or thread failing must not stop the others. A failed refresh is not
+    // surfaced anywhere — the loaded window simply stays as it was until the next reconnect, or until
+    // a query the consumer itself issues records its own `lastQueryError` on the paginator. Channels
+    // and threads are independent requests, so they run together rather than in sequence.
     await Promise.allSettled([
       ...channels.map((channel) => channel.reload()),
       ...threads.map((thread) => thread.reload()),
--- a/src/ConnectionRecoveryManager.ts
+++ b/src/ConnectionRecoveryManager.ts
@@ -1,12 +1,28 @@
 import { WithSubscriptions } from './utils/WithSubscriptions';
 import { chatLoggerSystem } from './logger';
 import { runDetached } from './utils';
+import { ConfigController } from './configuration/ConfigController';
+import { deepFreezeConfig } from './configuration/utils/deepFreezeConfig';
 import type { StreamChat } from './client';
 import type { Channel } from './channel';
 import type { Thread } from './thread';
-import type { Unsubscribe } from './store';
+import type { StateStore, Unsubscribe } from './store';
 
 const logger = chatLoggerSystem.getLogger('client');
+
+export type ConnectionRecoveryManagerConfig = {
+  /**
+   * Whether the client recovers its own state when the connection comes back.
+   *
+   * Set it to `false` only if the application recovers state itself — nothing is then re-queried or
+   * re-watched on reconnect, and refreshing whatever is on screen becomes the consumer's job.
+   * Read when a recovery actually runs, so flipping it takes effect from the next reconnect.
+   */
+  enabled: boolean;
+};
+
+export const DEFAULT_CONNECTION_RECOVERY_MANAGER_CONFIG: ConnectionRecoveryManagerConfig =
+  deepFreezeConfig({ enabled: true });
 
 /**
  * Owns connection recovery: after the socket comes back, brings the surfaces the app is actually
@@ -59,6 +75,8 @@ const logger = chatLoggerSystem.getLogger('client');
  */
 export class ConnectionRecoveryManager extends WithSubscriptions {
   client: StreamChat;
+  /** The shared configuration machinery — see {@link ConfigController}. */
+  private readonly configController: ConfigController<ConnectionRecoveryManagerConfig>;
   /** Set while a recovery is in flight; only used to keep the logs readable. */
   private isRecovering = false;
   /** Undone by `unregisterSubscriptions`, so re-registering does not stack listeners. */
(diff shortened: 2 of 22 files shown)


oliverlaz (approved):

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -706,7 +706,46 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+      //
+      // It's narrow on purpose, `liveHeadLogical` rather than `itemIntervals[0]`, which is only the head
+      // by convention and could be a logical tail and no anchored head, whose presence would mean
+      // this window is anchored with an older island merely in view. Keeps both neighbouring no-op
+      // contracts, so nothing loaded has no logical head (seeding is `postQueryReconcile`'s job).
+      const logicalHead = this.liveHeadLogical;
+      const hasAnchoredHead = this.itemIntervals.some(
+        (interval) => !isLogicalInterval(interval) && interval.isHead,
+      );
+      if (!logicalHead || hasAnchoredHead || !page?.length) return;
```
MartinCupela, line 730: Shouldn't `!page?.length) return` be at the top of the method execution to avoid unnecessary work and return early?

src/pagination/paginators/MessageIntervalPaginator.ts
```diff
@@ -706,7 +706,46 @@ export class MessageIntervalPaginator extends BasePaginator<
...
+
+      // Take over the view only if the reader is on the live content being absorbed, so if they jumped
+      // to another interval, anchor for their return and leave them there (the rule the jumped away
+      // branch below enforces).
+      const takeOverView = this.isActiveInterval(logicalHead);
+
+      const anchored = this.ingestPage({ page, isHead: true, setActive: false });
+      if (!anchored || !takeOverView) return;
+
+      const hasMoreTailFromLogical = !this.pageReachedChannelStart(page, options);
```
MartinCupela: Why is this called `hasMoreTailFromLogical`? Not sure what does it have in common with logical interval.

src/pagination/paginators/MessageIntervalPaginator.ts
MartinCupela: Probably it is ok here

src/channel_state.ts
```diff
@@ -144,6 +144,21 @@ export type ChannelLifecycleState = {
...
+   * state instead of holding a latch of its own. `watch()` still rethrows, so callers that await it
+   * are unaffected.
+   *
+   * Store-backed because the two failures a UI cares about reach it differently and neither is
+   * catchable at the call site: the reconnect reload is issued by `ConnectionRecoveryManager` inside
+   * a `Promise.allSettled`, and an open-while-offline `watch()` throws at mount, long before
+   * anything that could later prove it stale. Recording it here is what lets it be *invalidated* by
+   * the next attempt rather than guessed at.
+   */
+  lastLoadError?: Error;
```
MartinCupela: Why do we duplicate the information that is already in the paginator?

src/channel.ts
```diff
@@ -1874,20 +1899,28 @@ export class Channel extends ChannelApi {
...
+      if (failedBefore.length) {
+        // Membership is checked against the visible window, NOT `getItem`: that reads the item index,
+        // which can still hold a message the rebuild dropped from the loaded window — so an
+        // index-based guard skips the reingest on exactly the disjoint set rebuild that needs it,
+        // and the user's unsent message silently disappears from the list.
+        const visible = new Set((paginator.items ?? []).map((message) => message.id));
+        paginator.batch(
+          () => {
+            for (const failed of failedBefore) {
+              if (!visible.has(failed.id)) paginator.ingestItem(failed);
```
MartinCupela, line 1891: instead of invoking ingestItem wouldn't it be more effective just call removeItem as we know we want to remove the item?

src/ChannelManager.ts
```diff
@@ -134,6 +136,32 @@ export const ignoreEventsForUnknownChannels: EventHandlerPipelineHandler<
...
+ * dedupes concurrent watches for the same cid, so a burst of events cannot produce a burst of requests.
+ */
+const restoreInterruptedWatch = (channel: Channel, client: StreamChat) => {
+  if (channel.pendingDisposal) return;
+  if (channel.watchStatus !== ChannelWatchStatus.WasWatching) return;
+
+  // Takes the client as an argument rather than calling `channel.getClient()`, which THROWS for a
+  // channel pending disposal — the guard above makes that unreachable today, but a throw here would
+  // reject the whole event handler, so it is not a hazard worth leaving one edit away.
+  runDetached(getChannel({ channel, client }), {
```
MartinCupela, line 160: What is the added value of using `runDetached` here?

src/client.ts
```diff
@@ -158,13 +167,13 @@ export class StreamChat extends ChatApi {
...
-   * `recoverStateOnReconnect` parameter can be used in such cases, to disable state recovery within js client.
-   * When false, user/consumer of this client will need to make sure all the channels present on UI by
-   * manually calling queryChannels endpoint.
+   * When enabled, {@link ConnectionRecoveryManager} re-runs each loaded channel list's own first-page
+   * query and reloads each active channel, then dispatches `connection.recovered`.
+   *
+   * Set it to `false` only if the application recovers state itself — nothing will then be re-queried
+   * or re-watched on reconnect, and it becomes the consumer's job to refresh whatever is on screen.
    */
   recoverStateOnReconnect?: boolean;
```
MartinCupela: Feels like this should be part of the InstanceConfigurationRegistry. And remove from the StreamChatOptions.

src/client.ts
```diff
@@ -339,6 +348,11 @@ export class StreamChat extends ChatApi {
     this.threads = new ThreadManager({ client: this });
     this.polls = new PollManager({ client: this });
     this.channelManager = new ChannelManager({ client: this });
+    this.connectionRecovery = new ConnectionRecoveryManager({ client: this });
+    // Subscribed here rather than by the consumer: recovery is not an opt-in feature, it is what the
```
MartinCupela: I would remove the comment to reduce the bloat

src/client.ts
```diff
@@ -140,6 +141,14 @@ export class StreamChat extends ChatApi {
...
    */
   channelManager: ChannelManager;
+  /**
+   * Owns connection recovery: on reconnect it re-runs each loaded channel list's own first-page query
+   * and reloads each active channel, then dispatches `connection.recovered`. Instantiated with the
+   * client and subscribed for its lifetime.
+   *
+   * Turn it off with the `recoverStateOnReconnect` option if the application recovers state itself.
+   */
+  connectionRecovery: ConnectionRecoveryManager;
```
MartinCupela, line 152: I would remove this option and just keep the configuration via the InstanceConfigurationRegistry. ConnectionRecoveryManager should have own ConfigController to consume the setup from the registry.

src/ConnectionRecoveryManager.ts
```diff
@@ -0,0 +1,245 @@
...
+   *
+   * Deliberately NOT filtered on `watchStatus`. A channel opened while offline never reached
+   * `Watching` at all (its `watch()` threw, leaving `offlineMode`), so a `WasWatching` filter here
+   * would permanently strand the single most important offline case. `Channel.reload()`'s own
+   * `initialized || offlineMode` guard is the correct gate, and this is not a "restore a lost watch"
+   * decision anyway — it is "the channel being read must show the truth".
+   */
+  private get recoverableActiveChannels(): Channel[] {
+    const channels: Channel[] = [];
+    for (const cid in this.client.activeChannels) {
```
MartinCupela, line 195: Wouldn't it be more performant to keep a registry (Array or Set) of active channels instead of keeping the prop on instance and having to iterate over the whole cache of activeChannels? I think there will be 1 or 2 active channels at any time, so going through 100 cached channels may be too much unnecessary work.

Another thing is that we should be probably deprecating client.activeChannels and look up inside the corresponding EntityStore instead.

src/ConnectionRecoveryManager.ts
```diff
@@ -0,0 +1,245 @@
...
+   * The fix is a real off-list registry — see the commented-out `threadCache` in `ThreadManager` — at
+   * which point this getter reads from that instead, with no change to the `active` filter.
+   */
+  private get recoverableActiveThreads(): Thread[] {
+    const threads: Thread[] = [];
+    for (const thread of Object.values(this.client.threads.threadsById)) {
+      if (!thread) continue;
+      // Read off state rather than a getter: `Thread` has no `active` accessor the way `Channel`
+      // does, and adding one just for this would grow the public surface for a single internal read.
+      const { active } = thread.state.getLatestValue();
```
MartinCupela, line 239: I would keep the active threads again in a set or array as I proposed with the channel due to the same reasons described with channels.

MartinCupela: If the `active` info is needed only for purposes of state recovery I would even keep the active channels and threads registry (sets) inside the ConnectionRecoveryManager

src/thread.ts
```diff
@@ -57,6 +57,12 @@ export type ThreadState = {
   deletedAt: Date | null;
   isLoading: boolean;
   isStateStale: boolean;
+  /**
+   * The error thrown by the most recent {@link Thread.reload}, or `undefined` when it succeeded.
+   * Mirrors `channel.state.lastLoadError`: connection recovery runs reloads inside
+   * `Promise.allSettled`, so a failure has to be published rather than only thrown.
+   */
+  lastLoadError?: Error;
```
MartinCupela: I would avoid duplicating the info if we already have it on the paginator. Single source of truth looks more maintainable and better documentable to me.

src/ConnectionRecoveryManager.ts
MartinCupela: I think the unread state does somewhere rely on whether the channel is active, but we could be looking into a set of active channel CIDs to verify it as well.

src/channel.ts
isekovanic: That is not at all what we're trying to do here, we're actually just making sure that the failed messages are persisted so that they are not removed on reconnect. So no removal here.

src/ChannelManager.ts
isekovanic: Same as the other places it's used, it's a primitive that will prevent throwing upwards and log if anything bad happened.

src/channel_state.ts
isekovanic: Not really duplicated, perhaps the comment can be modified a little bit. While the paginator will hold something like an error during `toTail` or `toHead`, it will not be able to figure out an error that might happen for a raw `channel.watch()` somewhere. This does not really get reflected on the paginator itself at all as far as I can see. Maybe it's an angle to drill it down to the paginator if it does happen

src/client.ts
isekovanic: Yeah I can get behind that, it sounds better

src/ConnectionRecoveryManager.ts
isekovanic: I would not agree with this statement. It was evaluated and discarded, because it's simply due to the reason that this is never going to be a set of god knows how many instances (it naturally cannot). 

> may be too much unnecessary work

Let's please evaluate what the unnecessary work might mean. 100s of channels (and even 1000s) of channels are a piece of cake for V8/Hermes to run through shallowly and check for active ones. So we're talking nanoseconds of extra compute time that happens only on reconnects.

Adding a set for this introduces bookkeeping that is extremely unnecessary for something like this. Additionally that would then later completely prevent us acting on `state.active` if we need it elsewhere, as it belongs to a channel as part of the state. 

At a certain point we have to stop and think what we're optimizing against and what gains we'd get as compared to readability.

> Another thing is that we should be probably deprecating client.activeChannels and look up inside the corresponding EntityStore instead.

This is something that can be done whenever `activeChannels` is gone, until then we have to stick to it. Doesn't  make sense to pollute the PR with more changes doing that.

src/ConnectionRecoveryManager.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841#discussion_r3862472272 same as here

src/thread.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841#discussion_r3862399814 same as here, whatever we decide for that one will be reflected on threads as well

src/ConnectionRecoveryManager.ts
isekovanic: And to top it all off, all of this happens only on reconnect.

MartinCupela: The only thing I am saying that all the operations add up. Iterations over what we have, then ingestions which neither are light operations etc. We can take a look at the optimization later, in few weeks time.

src/channel.ts
```diff
@@ -1874,20 +1899,28 @@ export class Channel extends ChannelApi {
...
-            if (!paginator.getItem(failed.id)) paginator.ingestItem(failed);
-          }
-        },
-        { coalesce: true },
-      );
+      if (failedBefore.length) {
+        // Membership is checked against the visible window, NOT `getItem`: that reads the item index,
+        // which can still hold a message the rebuild dropped from the loaded window — so an
+        // index-based guard skips the reingest on exactly the disjoint set rebuild that needs it,
+        // and the user's unsent message silently disappears from the list.
```
MartinCupela:
```suggestion
        // Make sure the messages that failed to be sent before the watch() request are re-ingested after the watch() query.
```

src/channel.ts
MartinCupela: Ok, proposed comment adjustment: https://github.com/GetStream/stream-chat-js/pull/1841/changes#r3862879116

src/ConnectionRecoveryManager.ts
isekovanic: Yeah, they could add up - however we don't do any ingestion here at all, we just extract the `active: true` channels from whatever caching mechanism is present pretty much. I am all for performance naturally, but probably where it counts and on the actually hot paths, which this is probably not.

> Iterations over what we have

This is extremely cheap, not sure why we think these aren't light operations. The ingestions are going to happen regardless of how we consume the currently active channels (for which, keep in mind there's going to be 1 active channel most of the time and tens of loaded channels at all). 

If we can see some potential in which this might balloon to a very high number I'd be more than happy to add alternative ways to bookmark what `active` means, but otherwise I don't think we should be optimizing stuff that do not hurt us and complicate the code more. If you have some specific way in mind where this could explode please let me know, I might just be missing some edge case

src/channel.ts
isekovanic: sounds good, thanks

src/channel_state.ts
MartinCupela: Would it make sense to have it as a boolean flag - e.g. lastRequestFailed? Not sure if we need error reference, not sure about the purpose.

isekovanic: It could work but then we'd be swallowing the error essentially, I guess it's always better to have the actual error of what happened if it needs to be used later (whereas a boolean would just be something like "something went wrong, we don't know what")

isekovanic: Maybe it could make more sense to unify the 2 errors (either in paginator or on `channel`), as this later gives even integrators the opportunity to either listen to it or simply do something like:

```
const { lastError } = useStateStore((state) => ({ lastError: state.lastLoadError }), channel.state)

if (lastError) {
  throw lastError
} 
```

so that it can be caught within an `ErrorBoundary` or something

src/pagination/paginators/MessageIntervalPaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841/commits/213fe3bfa8ad4c016ff329dda6646e55c89974f7

src/channel_state.ts
isekovanic: removed here: https://github.com/GetStream/stream-chat-js/pull/1841/commits/213fe3bfa8ad4c016ff329dda6646e55c89974f7

src/channel.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841/changes/213fe3bfa8ad4c016ff329dda6646e55c89974f7

src/thread.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841/changes/213fe3bfa8ad4c016ff329dda6646e55c89974f7

src/client.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841/changes/e14bf893ba28d3d1e518005daf055898214f3c88

src/client.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1841/changes/e14bf893ba28d3d1e518005daf055898214f3c88

MartinCupela (approved):

PR #1843, opened 2026-08-27
feat: remaining optimistic updates

isekovanic:
## CLA

- [x] Code changes are tested

`MessageOperationStatePolicy` is now operation-aware: `optimistic(kind, params)` returns what it wrote,
and `success` / `failure` take the `kind`, because the three operations genuinely want different state
transitions. Four new context members — `remove`, `persist`, `purge`, `isQueued` — are supplied by
`Channel` and `Thread` through a shared `createMessageOperationsPersistence`, so the two can't drift.
`MessageOperations.delete` now goes through the same `run()` lifecycle as the rest, which is why it had
no optimism before: it was the one operation sitting outside it.

The behaviour, which is deliberate and worth reviewing as such:

| | on a definitive failure | on a QUEUED (offline) failure |
|---|---|---|
| **edit** | keeps the edit, marks the message failed + error | keeps the edit, **no** failed state |
| **delete** | **reverts** | leaves the optimistic delete in place |
| **send / retry** | marks failed | still marks **failed** |

- **An edit is never reverted** — rolling it back would destroy text the user typed.
- **A delete is reverted**, because unlike an edit there is nothing of the user's to lose, and a
  "Message deleted" placeholder on a message that still exists is a lie that only self-corrects on the
  next query. The realistic trigger is offline support disabled with no connectivity: nothing is
  queued, so nothing was deleted.
- **The queued short-circuit is scoped to update/delete.** A queued *send* still settles `failed`,
  because the retry affordance is the only way the user gets that message out. This is v9 behaviour;
  treating it as merely pending leaves it spinning forever.
- Promises still reject in every failure case, deliberately, so an integrator's error notification keeps
  firing. Snackbar yes, failed state on the message no.

Reverts are guarded by object identity — `EntityStore.upsert` replaces the canonical copy, so
`get(id) === applied` is a sound "nothing else has written since" check, and a WS event landing
mid-flight is not clobbered.

### 2. A failed message now survives a restart

Sends write ahead to the offline DB **pessimistically as `failed`**, so a process death between compose
and server ack leaves a message that hydrates as failed-and-retryable rather than vanishing. Success
overwrites it.

V9 recovered failed messages by diffing a second, lagging copy of the list (the React SDK's own state).
With a single reactive source of truth there is no second copy, so the **persisted row is that buffer**:
`Channel.reload()` unions its in-memory failed messages with the DB's, in-memory winning.

Two new **concrete** helpers on `AbstractOfflineDB`, composed from existing primitives so a custom
offline DB inherits them and has nothing new to implement:

- `getFailedMessages({ cid })` — read back through `getChannels`.
- `upsertMessageWithChannelGuard({ message })` — upserts one message, creating its channel row first if
  the DB has never seen that channel, so the optimistic write can't fail on the foreign key.
  `updateMessage` can't serve this: it is an UPDATE and no-ops when the row doesn't exist, which is
  exactly the write-ahead case.

### 3. Bug fixes found along the way

**Editing or deleting a thread parent from inside a thread was a silent no-op.** The reply paginator's
local filter is `{ cid, parent_id }` and a parent message has no `parent_id`, so routing a parent edit
through the open thread's instance — which is what the SDKs do while a thread is on screen — handed it
to a collection that could not hold it. Optimistic writes now fall back to the client-global message
store, reaching the message wherever it is held and fanning out to every collection that holds it. Same
addressing `applyReactionLocally` already used.

**An empty request response no longer overwrites local state.** `formatMessage(undefined)` doesn't
throw — it returns `{ status: 'received', created_at: <now>, updated_at: <now> }`, and because that
`updated_at` is *now* it beat the freshness check and was ingested as an id-less message. Reachable from
a custom request handler that resolves without a `message`.

**The offline channel guard is lazy now.** It attempts the write first and only probes for the channel
row when the write fails, repairing and retrying once; every statement involved is an upsert, so the
retry is idempotent. The eager probe is kept for `execute: false` callers and for `forceUpdate`. This
matters because the probe is a native round-trip, not a cheap read — measured through op-sqlite on
device at ~0.5ms against ~8–12ms for the message upsert it protects, and it sat on every message
received as well as every message written.

**The optimistic layer owns the offline-DB row for deletes.** `client.deleteMessage` soft- or
hard-deleted the row *before* it queued the request: a second uncoordinated writer, the mirror moving
ahead of the state it mirrors, and — the live bug — **a custom `deleteMessageRequest` never reaches
`client.deleteMessage`**, so with one configured the row was never marked while memory showed the
message deleted. Moved into `policy.optimistic`, after the "nothing holds this message" guard, since the
DB mirrors local state and shouldn't be told about state that doesn't exist.

### 4. Queueable operations are declared in one place

`PendingTask`'s payload union was a hand-written seven-arm type kept in step with `executeTask`'s
seven-arm `if` ladder by hand — it carried a `// TODO: Please rethink the definition of PendingTasks as
it seems awkward`.

Each operation is now declared as **the method that performs it**:

```ts
export type QueueableOperationSignatures = {
  'send-message': Channel['_sendMessage'];
  'update-message': StreamChat['_updateMessage'];
  …
};
```

`PendingTask`'s payloads derive from that map, and `QUEUEABLE_OPERATIONS` is a mapped type over its
keys — so **a task type without a resolver does not compile**, rather than failing at runtime with
`Tried to execute invalid pending task type` the first time someone queues one.

Payload types are byte-identical to before (`Parameters<Channel['sendReaction']>` ≡
`Channel['_sendReaction']`), so tasks already serialized in a user's SQLite still deserialize. **No
migration.**

`executeTask` is a lookup now, and `onReplay` became a declared field — which makes visible that
`send-message` does thread `upsertReplyLocally` + `trackLastMessage` **only on a replay**, previously
buried in an `if (isPendingTask)`. The registry is `@internal` and not exported: it's a plain mutable
object, and exporting it would let an integrator reassign what a replay runs.

## Changelog

- feat: optimistic updates for editing and deleting a message
- feat: a message that failed to send now survives an app restart and stays retryable
- fix: editing or deleting a thread parent from inside an open thread is no longer a silent no-op
- fix: an offline edit whose task the queue refuses (an attachment still on a local URL) now settles as
  failed instead of appearing pending forever
- fix: a custom `deleteMessageRequest` no longer leaves the offline-DB row unmarked
- fix: a request response carrying no message no longer overwrites local state with an id-less message
- perf: the offline channel guard attempts the write first and only probes for the channel row on
  failure, taking a native round-trip off every message written and every message received
- chore: queueable operations are declared in one typed registry, so a task type without a resolver is a
  compile error

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -1,5 +1,6 @@
 import { ChannelState, ChannelWatchStatus } from './channel_state';
 import { CooldownTimer } from './CooldownTimer';
+import { queueOrRun } from './offline-support/queueableOperations';
 import { MessageComposer } from './messageComposer';
 import { MessageReceiptsTracker } from './messageDelivery';
 import type { ReadStoreReconcileMeta } from './messageDelivery';
@@ -42,7 +43,6 @@ import type {
   ChannelStateResponseFields,
   ChannelUpdateOptions,
   Command,
-  CreateDraftResponse,
   DeleteMessageOptions,
   Event,
   EventHandler,
@@ -85,7 +85,6 @@ import type { ChatApi } from './gen/chat/ChatApi';
 import { ChannelApi } from './gen/chat/ChannelApi';
 
 const logger = chatLoggerSystem.getLogger('channel');
-const offlineDbLogger = chatLoggerSystem.getLogger('offline-db');
 
 // todo: move to dedicated file
 export type SendMessageWithStateUpdateParams = {
@@ -678,26 +677,20 @@ export class Channel extends ChannelApi {
    */
   override async sendMessage(...args: Parameters<ChannelApi['sendMessage']>) {
     const [request] = args;
-    try {
-      const offlineDb = this.getClient().offlineDb;
-      const messageId = request.message?.id;
-      if (offlineDb && messageId) {
-        return await offlineDb.queueTask<Awaited<ReturnType<ChatApi['sendMessage']>>>({
-          task: {
-            channelId: this.id as string,
-            channelType: this.type,
-            messageId,
-            payload: args,
-            type: 'send-message',
-          },
-        });
-      }
-    } catch (error) {
-      offlineDbLogger
-        .withExtraTags('sendMessage', this.cid)
-        .error('Sending the message failed.', { error });
-    }
-    return await this._sendMessage(...args);
+    const messageId = request.message?.id;
+
+    return await queueOrRun({
+      client: this.getClient(),
+      // Nothing to key a queue entry on without a message id, so it runs but is not queued.
+      queue: !!messageId,
+      task: {
+        channelId: this.id as string,
+        channelType: this.type,
+        messageId,
+        payload: args,
+        type: 'send-message',
+      },
+    });
   }
 
   /**
@@ -911,28 +904,18 @@ export class Channel extends ChannelApi {
   async sendReaction(...args: Parameters<ChatApi['sendReaction']>) {
     const [{ id: messageId }] = args;
 
-    try {
-      const offlineDb = this.getClient().offlineDb;
-      if (offlineDb) {
-        // The optimistic reaction row is written by the local-update layer
-        // (`applyReactionLocally`); here we only queue the request for replay.
-        return await offlineDb.queueTask<Awaited<ReturnType<ChatApi['sendReaction']>>>({
-          task: {
-            channelId: this.id as string,
-            channelType: this.type,
-            messageId,
-            payload: args,
-            type: 'send-reaction',
-          },
-        });
-      }
-    } catch (error) {
-      offlineDbLogger
-        .withExtraTags('sendReaction', this.cid)
-        .error('Sending the reaction failed.', { error });
-    }
-
-    return this._sendReaction(...args);
+    // The optimistic reaction row is written by the local-update layer (`applyReactionLocally`); here
+    // we only queue the request for replay.
+    return await queueOrRun({
+      client: this.getClient(),
+      task: {
+        channelId: this.id as string,
+        channelType: this.type,
+        messageId,
+        payload: args,
+        type: 'send-reaction',
+      },
+    });
   }
 
   _sendReaction(...args: Parameters<ChatApi['sendReaction']>) {
@@ -943,28 +926,18 @@ export class Channel extends ChannelApi {
     this._checkInitialized();
     const [request] = args;
 
-    try {
-      const offlineDb = this.getClient().offlineDb;
-      if (offlineDb) {
-        // The optimistic reaction-row removal is handled by the local-update layer
-        // (`applyReactionLocally`); here we only queue the request for replay.
-        return await offlineDb.queueTask<Awaited<ReturnType<ChatApi['deleteReaction']>>>({
-          task: {
-            channelId: this.id as string,
-            channelType: this.type,
-            messageId: request.id,
-            payload: args,
-            type: 'delete-reaction',
-          },
-        });
-      }
-    } catch (error) {
-      offlineDbLogger
-        .withExtraTags('deleteReaction', this.cid)
-        .error('Deleting the reaction failed.', { error });
-    }
-
-    return await this._deleteReaction(...args);
+    // The optimistic reaction-row removal is handled by the local-update layer
+    // (`applyReactionLocally`); here we only queue the request for replay.
+    return await queueOrRun({
+      client: this.getClient(),
+      task: {
+        channelId: this.id as string,
+        channelType: this.type,
+        messageId: request.id,
+        payload: args,
+        type: 'delete-reaction',
+      },
+    });
   }
 
   /**
@@ -2303,26 +2276,17 @@ export class Channel extends ChannelApi {
    */
   override async createDraft(...args: Parameters<ChannelApi['createDraft']>) {
     const [request] = args;
-    try {
-      const offlineDb = this.getClient().offlineDb;
-      if (offlineDb) {
-        return (await offlineDb.queueTask<CreateDraftResponse>({
-          task: {
-            channelId: this.id as string,
-            channelType: this.type,
-            threadId: request.message?.parent_id,
-            payload: args,
-            type: 'create-draft',
-          },
-        })) as Awaited<ReturnType<ChannelApi['createDraft']>>;
-      }
-    } catch (error) {
-      offlineDbLogger
-        .withExtraTags('createDraft', this.cid)
-        .error('Creating the draft in the offline database failed.', { error });
-    }
 
-    return this._createDraft(...args);
+    return await queueOrRun({
+      client: this.getClient(),
+      task: {
+        channelId: this.id as string,
+        channelType: this.type,
+        threadId: request.message?.parent_id,
+        payload: args,
+        type: 'create-draft',
+      },
+    });
   }
 
   async _deleteDraft(...args: Parameters<ChannelApi['deleteDraft']>) {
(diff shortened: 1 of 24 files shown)


src/messageOperations/optimistic.ts
```diff
@@ -0,0 +1,240 @@
...
+export type MessageChange = LocalMessage | typeof REMOVE_MESSAGE | undefined;
+
+/** An operation's local-state shape: given the copy we hold, what should be there instead. */
+export type MessageProducer = (current: LocalMessage | undefined) => MessageChange;
+
+/**
+ * Where an optimistic operation reads and writes local state. `Channel` and `Thread` each expose one
+ * as `messageState`, routed paginator-first with the client-global message store as the fallback.
+ */
+export type MessageLocalState = {
```
MartinCupela: The name `MessageLocalState` sounds like the state store shape. But here it looks more like an API to interact with some state that holds LocalMessages. It is more like a LocalMessageAccessor?

src/messageOperations/optimistic.ts
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * What a {@link MessageProducer} returns to say "this operation takes the message out of local state"
+ * — a hard delete. Distinct from `undefined`, which means "nothing to do".
+ */
+export const REMOVE_MESSAGE = Symbol('REMOVE_MESSAGE');
+
+export type MessageChange = LocalMessage | typeof REMOVE_MESSAGE | undefined;
+
+/** An operation's local-state shape: given the copy we hold, what should be there instead. */
+export type MessageProducer = (current: LocalMessage | undefined) => MessageChange;
```
MartinCupela:
```suggestion
export type MessageChangeProducer = (current: LocalMessage | undefined) => MessageChange;
```

src/messageOperations/optimistic.ts
```diff
@@ -0,0 +1,240 @@
...
+ * Everything operation-specific is in `produce`. Everything an operation would otherwise hand-roll —
+ * the read, the write, the "has anything else written since" guard, working out what the inverse of the
+ * write is — is here.
+ *
+ * Writes go through `state.ingest` / `state.remove` rather than a bare store upsert, so a change that
+ * affects COLLECTION MEMBERSHIP works and not just a content change: a store upsert replaces the
+ * message wherever it is already held but never consults a collection's filter, so flipping `pinned`
+ * could never make it appear in a list it now belongs to.
+ */
+export const applyMessageChangeLocally = (
```
MartinCupela, line 95: Is this used anywhere? Cannot find any call to `applyMessageChangeLocally()`.

src/messageOperations/optimistic.ts
```diff
@@ -0,0 +1,240 @@
...
+): Promise<boolean> => {
+  const { offlineDb } = client;
+  // No queue at all, so nothing can be pending. `initialized` matters as much as existence: a DB whose
+  // `init()` never succeeded cannot hold a pending task.
+  if (!offlineDb?.state.getLatestValue().initialized) return false;
+
+  // Optional-chained: `getPendingTasks` is part of the `OfflineDBApi` an integrator can implement, so
+  // this must not assume a well-formed return.
+  const pending = await offlineDb.getPendingTasks({ messageId });
+  return !!pending?.length;
```
MartinCupela: Would this be false positive in case the pending tasks are only of type `'send-reaction'` or `'delete-reaction'`? Not sure if we want to report `true` only if there are pending tasks for 'update-message',  'delete-message', 'delete-reaction', 'send-message'.

src/messageOperations/MessageOperationStatePolicy.ts
```diff
@@ -14,32 +40,133 @@ const parseError = (error: unknown): StreamAPIError => {
...
+
+    if (kind === 'update') {
+      // Preserve the status: an edit must not turn a received message into `sending`, and an edit of a
+      // message that never left the device has to stay `failed`.
+      const isFailed = localMessage.status === 'failed';
+      const editedAt = new Date();
+      const applied: LocalMessage = {
+        ...localMessage,
+        error: isFailed ? localMessage.error : undefined,
+        updated_at: editedAt,
```
MartinCupela, line 114: Is there any chance that server's `updated_at` timestamp < `editedAt = new Date()` and thus not allowing to apply server-side state?

src/messageOperations/optimistic.ts
isekovanic: yes that's much better, thank you

src/messageOperations/optimistic.ts
isekovanic: Currently not - the idea behind adding it was so that we can:

1. Add optimistic updates more easily in the future (a good example for this is pinning a message)
2. Integrators can add their own optimistic updates for maybe some peculiar business logic with some part of the heavylifting done for them

If you prefer we can defer this maybe to a cookbook or something ?

MartinCupela: I am ok to keep. Just did not know if I missed anything

src/channel.ts
```diff
@@ -375,11 +377,34 @@ export class Channel extends ChannelApi {
...
+        store.flushSubscribers(m.id);
+      },
+      // Mirrors `ingest`'s routing: a message this paginator does not hold can still be held by the
+      // client-global store (a thread parent, a message displayed by another collection), and the
+      // policy uses this both for its freshness comparison and to decide whether there is anything to
+      // update optimistically at all. Reading only the paginator would make those two disagree.
+      get: (id) =>
+        this.messagePaginator.getItem(id) ?? this.getClient().messageStore.get(id),
+      remove: (id) => {
+        this.messagePaginator.removeItem({ id });
```
MartinCupela, line 409: We may want to handle the messages with property `show_in_channel` in thread paginator as well. Thread does this

src/messageOperations/optimistic.ts
isekovanic: No no, you're good - just wanted to actually experiment with trying to make this generic as then we can add way more optimistic operations later on (for messages at least, I haven't yet figured out how to make it truly generic for any type of entity)

src/channel.ts
isekovanic: Great catch, missed this. I fixed it here for now, but created this as a followup:

https://linear.app/stream/issue/REACT-1119/introduce-eviction-to-entitystore

since we do not want to keep remembering to do this. I experimented with a short PoC and it looks good, but I wouldn't rush it in this PR as it (naturally) has some odd ends that need to be polished.

oliverlaz (approved):

src/messageOperations/MessageOperationStatePolicy.ts
MartinCupela: Is this relevant question?

src/messageOperations/optimistic.ts
MartinCupela: Is this relevant question?

isekovanic: Apologies, missed this comment and resolved it by accident. It is indeed an issue, fixing it now

src/messageOperations/MessageOperationStatePolicy.ts
isekovanic: This should be resolved here: https://github.com/GetStream/stream-chat-js/pull/1843/changes/49ef0256eb78d47f8c941772cf6da5830f079382

src/messageOperations/optimistic.ts
isekovanic: Should be resolved here now: https://github.com/GetStream/stream-chat-js/pull/1843/changes/abbe285751262c631f76e8f2a2965646e5722f04

src/messageOperations/MessageOperationStatePolicy.ts
isekovanic: And here: https://github.com/GetStream/stream-chat-js/pull/1843/changes/dd39e0ef01319172b4cc2b647d7243794503d3d5

src/messageOperations/optimistic.ts
```diff
@@ -0,0 +1,249 @@
...
+  const { offlineDb } = client;
+  // No queue at all, so nothing can be pending. `initialized` matters as much as existence: a DB whose
+  // `init()` never succeeded cannot hold a pending task.
+  if (!offlineDb?.state.getLatestValue().initialized) return false;
+
+  // Optional-chained: `getPendingTasks` is part of the `OfflineDBApi` an integrator can implement, so
+  // this must not assume a well-formed return.
+  const pending = await offlineDb.getPendingTasks({ messageId });
+
+  return pending?.some((task) => types.includes(task.type));
```
MartinCupela:
```suggestion
  return !!pending?.some((task) => types.includes(task.type));
```

Just in case `pending` would be undefined - to match the declared return value by `isQueuedForReplay`

isekovanic: https://github.com/GetStream/stream-chat-js/pull/1843/changes/d1024c53a135d26d90a9e1c1a6a642750a7edc3b

PR #1845, opened 2026-08-27
feat(MessageComposer): add composition middleware for pending attachment uploads

MartinCupela:
## Goal

Make it possible to compose a message while its attachment uploads are still in flight.

Today that is refused twice over: the attachments composition middleware discards any composition with uploads running (warning `Wait until all attachments have uploaded`), and `MessageComposer.hasSendableData` returns `false` for as long as `uploadsInProgressCount > 0`. A UI SDK therefore cannot offer sending before every transfer has finished, however long that takes.

This PR adds the composer-side half of that flow, plus the fixes it exposed in upload state, cancellation and preview lifecycle.

It is deliberately only half. The composition produced is not ready for the wire — `message.attachments` omits every attachment that has no URL yet — so the other half belongs to whoever performs the send: await the pending uploads, write the resolved URLs into the payload, then make the request. That is why nothing here is enabled by a config option.

Backwards compatible except in one respect: outgoing attachments are now sanitized on every send and every update, which strips composer-internal state that was previously forwarded to the API. Everything else takes effect only once a UI SDK installs the new middleware.

### Composing with pending uploads

`createSendWithPendingUploadsAttachmentsMiddleware` replaces the default attachments composition middleware and shares its id, so `replace()` keeps its position in the chain. It splits the two payloads instead of discarding the composition:

- `localMessage.attachments` keeps `localMetadata` for anything still uploading — the `UploadManager` id, the file handle needed to await the upload, the local preview URI;
- `message.attachments` carries only attachments that already resolved to a URL.

Neither key is introduced when its list is empty, matching the default middleware — an empty `attachments` array is not "nothing to say", it reads as "remove every attachment" on an edit.

Attachments still uploading stay in the composer when the composition carries a poll. A poll composition does not clear the composer, so handing the attachment over would leave the same `localMetadata.id` owned by both the message and the composer, and a second `UploadManager.upload` call would restart a request whose in-flight entry had already been cleaned up.

### One switch: the middleware declares what it does

`MessageCompositionMiddleware.allowsPendingUploads` is a declaration a composition middleware sets on itself, and `MessageComposer.hasSendableData` reads it: while such a middleware is installed, an upload in flight is no longer a blocker and a pending attachment counts as content in its own right. `failed` and `blocked` attachments still do not count, so a message whose only attachment was rejected is not sendable.

Installing the middleware is therefore the whole switch — there is no second flag for a UI SDK to keep in sync with it, and a custom middleware implementing the same contract is recognised regardless of the id it uses.

Reading the declaration needs `MiddlewareExecutor.installedMiddleware`, a new read-only view of the chain (`MessageComposerMiddlewareExecutor` narrows it to the composition middleware type). Both the declaration and `MessageComposer.allowsPendingUploads` are marked temporary: v10 moves this into the composer configuration.

### The window between the last byte and the response

`UploadRecord.uploadConfirmationPending` marks the gap between the request body being flushed and the server answering, and is mirrored onto `localMetadata`. Upload progress counts bytes written to the connection, not bytes acknowledged, so it reaches 100% while the file is still being ingested — long enough to matter for a large file on a slow link. The flag is lowered only by a progress report carrying a number below 100: a report with no number means the transport cannot measure this upload, not that flushed bytes were un-sent.

### Outgoing attachments are sanitized

`sanitizeOutgoingAttachments` strips `localMetadata` and drops any attachment that never resolved, reporting the drop through `client.logger` and `console.warn`. It runs in `channel._sendMessage` and `client._updateMessage` — the two methods every path converges on, including the offline replay of a queued task, which calls them directly with a payload that a merged failed edit may have rewritten since it was stored.

"Resolved" means a remote `http(s)` URL, not merely the presence of one. A local reference — `blob:`, `file:`, `content:`, an inline `data:` payload — resolves only on the device that produced it, and forwarding it would store an attachment nobody else can load. Reaching that branch means the code that composed the message has a bug, so it is logged rather than raised as a notification, which is a channel for things an end user can act on. `partialUpdateMessage` is deliberately not covered: its `set` is a caller-authored patch that no composer output flows through.

### Cancellation is not a failure

Removing an attachment mid-upload aborts the request through its `AbortController`. `isUploadCancellation` recognises the result — an axios cancellation, or a `DOMException` named `AbortError`, which is what a custom transport conventionally throws — and both the post-upload error middleware and the deprecated `uploadAttachment` path now stay quiet instead of raising an upload error for something the user asked for.

### Preview lifecycle

`removeAttachments` releases the removed attachment's `blob:` preview, which previously leaked one object URL per removal.

The post-upload enrichment middleware releases a preview only while the composer still holds the attachment, which its new optional `composer` argument is what tells it. The middleware runs when an upload finishes, and that can now happen after the composer was cleared — at which point something else is rendering from that preview and releasing it would blank it out.

### New exports

`isPendingUpload` / `isFinishedUpload` (upload-state predicates, typed against `AttachmentLoadingState`), `isUploadCancellation`, and the tag-keyed async runners `withoutConcurrency`, `withCancellation`, `hasPending`, `settled` — so a UI SDK can serialise its own actions with the primitive this SDK already uses internally instead of hand-rolling promise chains.

### What a UI SDK has to implement

1. Install `createSendWithPendingUploadsAttachmentsMiddleware` on the composers it wants this for. Sendability follows from the declaration; nothing else needs switching.
2. In its send path, await the uploads still in flight (`UploadManager.upload` is idempotent by `localMetadata.id`, so calling it again returns the in-flight promise), write the resolved URLs into `message.attachments`, and only then make the request. Doing this in the send path rather than at submit time also covers resending a failed message, and skipping attachments that already carry a URL makes a retry re-upload only what failed.
3. Decide what ordering it wants. Two sends started in quick succession now differ in duration by however long an upload takes, and `created_at` is stamped when the request arrives — `withoutConcurrency` is exported for serialising them per channel.

`uploadConfirmationPending` is available for both a composer attachment (from `localMetadata`) and a message attachment (from the live `UploadManager` record) if the SDK wants to distinguish "sending" from "sent but unconfirmed".

### Tests

39 cases across the composition middleware, `hasSendableData`, `UploadManager`, `AttachmentManager`, the post-upload middlewares, `sanitizeOutgoingAttachments`, and the `channel` / `client` send and update paths.

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -9,6 +9,7 @@ import {
   logChatPromiseExecution,
   messageSetPagination,
   normalizeQuerySort,
+  sanitizeOutgoingAttachments,
 } from './utils';
 import type { StreamChat } from './client';
 import { DEFAULT_QUERY_CHANNEL_MESSAGE_LIST_PAGE_SIZE } from './constants';
@@ -222,7 +223,9 @@ export class Channel {
     );
   }
 
-  async sendMessage(message: Message, options?: SendMessageOptions) {
+  async sendMessage(rawMessage: Message, options?: SendMessageOptions) {
+    const message = sanitizeOutgoingAttachments({ channel: this, message: rawMessage });
+
     try {
       const offlineDb = this.getClient().offlineDb;
       if (offlineDb) {
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -213,19 +213,20 @@ export class Channel {
    *
    * @return {Promise<SendMessageAPIResponse>} The Server Response
    */
-  async _sendMessage(message: Message, options?: SendMessageOptions) {
-    return await this.getClient().post<SendMessageAPIResponse>(
-      this._channelURL() + '/message',
-      {
-        message,
-        ...options,
-      },
-    );
+  async _sendMessage(rawMessage: Message, options?: SendMessageOptions) {
+    const client = this.getClient();
+    const message = sanitizeOutgoingAttachments({
+      client,
+      message: rawMessage,
+    });
+
+    return await client.post<SendMessageAPIResponse>(this._channelURL() + '/message', {
+      message,
+      ...options,
+    });
   }
 
-  async sendMessage(rawMessage: Message, options?: SendMessageOptions) {
-    const message = sanitizeOutgoingAttachments({ channel: this, message: rawMessage });
-
+  async sendMessage(message: Message, options?: SendMessageOptions) {
     try {
       const offlineDb = this.getClient().offlineDb;
       if (offlineDb) {
--- a/src/messageComposer/attachmentIdentity.ts
+++ b/src/messageComposer/attachmentIdentity.ts
@@ -1,5 +1,6 @@
 import type { Attachment, SharedLocationResponse } from '../types';
 import type {
+  AttachmentLoadingState,
   AudioAttachment,
   FileAttachment,
   GiphyAttachment,
@@ -30,8 +31,14 @@ export const isLocalUploadAttachment = (
 /**
  * Upload states meaning "the bytes are not on the CDN yet, but they are on their way".
  * `blocked` and `failed` are excluded - those never resolve on their own.
+ *
+ * Typed against {@link AttachmentLoadingState} rather than `string`, so renaming a state fails
+ * the build instead of silently making this list match nothing - it decides what gets sent.
  */
-const PENDING_UPLOAD_STATES: ReadonlyArray<string> = ['pending', 'uploading'];
+const PENDING_UPLOAD_STATES: ReadonlyArray<AttachmentLoadingState> = [
+  'pending',
+  'uploading',
+];
 
 /** Whether an upload for this attachment is still expected to resolve. */
 export const isPendingUpload = (
--- a/src/messageComposer/attachmentIdentity.ts
+++ b/src/messageComposer/attachmentIdentity.ts
@@ -27,6 +27,26 @@ export const isLocalUploadAttachment = (
 ): attachment is LocalUploadAttachment =>
   !!(attachment as LocalAttachment)?.localMetadata?.uploadState;
 
+/**
+ * Upload states meaning "the bytes are not on the CDN yet, but they are on their way".
+ * `blocked` and `failed` are excluded - those never resolve on their own.
+ */
+const PENDING_UPLOAD_STATES: ReadonlyArray<string> = ['pending', 'uploading'];
+
+/** Whether an upload for this attachment is still expected to resolve. */
+export const isPendingUpload = (
+  attachment: unknown,
+): attachment is LocalUploadAttachment =>
+  isLocalUploadAttachment(attachment) &&
+  PENDING_UPLOAD_STATES.includes(attachment.localMetadata.uploadState);
+
+/** Whether this attachment's upload already resolved to a URL. */
+export const isFinishedUpload = (
+  attachment: unknown,
+): attachment is LocalUploadAttachment =>
+  isLocalUploadAttachment(attachment) &&
+  attachment.localMetadata.uploadState === 'finished';
+
 export const isFileAttachment = (
   attachment: Attachment | LocalAttachment,
   supportedVideoFormat: string[] = [],
--- a/src/messageComposer/messageComposer.ts
+++ b/src/messageComposer/messageComposer.ts
@@ -434,44 +434,49 @@ export class MessageComposer extends WithSubscriptions {
   }
 
   /**
-   * Sendability for a UI SDK that supports sending while attachments are still uploading.
+   * Whether the installed composition middleware composes messages whose uploads are still in
+   * flight — see `MessageCompositionMiddleware.allowsPendingUploads`.
    *
-   * Differs from {@link hasSendableData} in that an upload in flight does not block the send, and
-   * a pending attachment counts as content in its own right — otherwise an attachment-only
-   * message could never be sent before its upload finished, which is the point of the flow.
-   * `failed` and `blocked` attachments still do not count: a message whose only attachment was
-   * rejected must not look sendable.
+   * Derived from the chain rather than from config, so there is exactly one switch: the UI SDK
+   * that implements the rest of the flow installs the middleware, and everything that has to
+   * follow from that (starting with {@link hasSendableData}) follows on its own.
    *
-   * Read this instead of `hasSendableData` when
-   * {@link createSendWithPendingUploadsAttachmentsMiddleware} is installed. There is no config
-   * option choosing between the two — the switch belongs to the UI SDK that implements the flow.
+   * **Temporary API.** In v10 sending with pending uploads becomes part of the composer
+   * configuration (`MessageComposerConfig`, set through `updateConfig`), and this getter,
+   * together with the middleware declaration it reads, is replaced by that config field.
    */
-  get hasSendableDataWithPendingUploads() {
-    const hasSendableAttachment = this.attachmentManager.attachments.some(
-      (attachment) => isFinishedUpload(attachment) || isPendingUpload(attachment),
-    );
-
-    return (
-      this.isCommandSendable &&
-      !!(
-        !this.textComposer.textIsEmpty ||
-        hasSendableAttachment ||
-        this.pollId ||
-        this.locationComposer.validLocation
-      )
+  get allowsPendingUploads() {
+    return this.compositionMiddlewareExecutor.installedMiddleware.some(
+      (middleware) => middleware.allowsPendingUploads,
     );
   }
 
   get hasSendableData() {
-    return (
-      this.isCommandSendable &&
-      !!(
-        (!this.attachmentManager.uploadsInProgressCount &&
-          (!this.textComposer.textIsEmpty ||
-            this.attachmentManager.successfulUploadsCount > 0)) ||
+    if (!this.isCommandSendable) return false;
+
+    if (this.allowsPendingUploads) {
+      // An upload in flight is no longer a blocker, and a pending attachment counts as content
+      // in its own right — otherwise an attachment-only message could never be sent before its
+      // upload finished, which is the point of the flow. `failed` and `blocked` attachments
+      // still do not count: a message whose only attachment was rejected must not look sendable.
+      const hasSendableAttachment = this.attachmentManager.attachments.some(
+        (attachment) => isFinishedUpload(attachment) || isPendingUpload(attachment),
+      );
+
+      return !!(
+        !this.textComposer.textIsEmpty ||
+        hasSendableAttachment ||
         this.pollId ||
-        !!this.locationComposer.validLocation
-      )
+        this.locationComposer.validLocation
+      );
+    }
+
+    return !!(
+      (!this.attachmentManager.uploadsInProgressCount &&
+        (!this.textComposer.textIsEmpty ||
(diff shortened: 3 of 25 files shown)


src/channel.ts
```diff
@@ -222,7 +223,9 @@ export class Channel {
     );
   }
 
-  async sendMessage(message: Message, options?: SendMessageOptions) {
+  async sendMessage(rawMessage: Message, options?: SendMessageOptions) {
+    const message = sanitizeOutgoingAttachments({ channel: this, message: rawMessage });
```
isekovanic: Does the update path need this too? `client.updateMessage` isn't sanitized, and `handleOfflineFailedUpdateMessagePendingTask` merges an unsanitized `update-message` payload into the pending `send-message` task, which then replays through `_sendMessage`. So composing an edit with pending uploads still puts `localMetadata` at risk I think. Or is edit out of scope for now ?

src/messageComposer/attachmentIdentity.ts
```diff
@@ -27,6 +27,26 @@ export const isLocalUploadAttachment = (
 ): attachment is LocalUploadAttachment =>
   !!(attachment as LocalAttachment)?.localMetadata?.uploadState;
 
+/**
+ * Upload states meaning "the bytes are not on the CDN yet, but they are on their way".
+ * `blocked` and `failed` are excluded - those never resolve on their own.
+ */
+const PENDING_UPLOAD_STATES: ReadonlyArray<string> = ['pending', 'uploading'];
```
isekovanic: Can this be `ReadonlyArray<AttachmentLoadingState>` rather than `string`? It decides what gets sent now, so I'd rather the TS compiler complain if a state ever gets renamed than have it silently stop matching.

src/messageComposer/messageComposer.ts
```diff
@@ -432,6 +433,35 @@ export class MessageComposer extends WithSubscriptions {
...
+   *
+   * Differs from {@link hasSendableData} in that an upload in flight does not block the send, and
+   * a pending attachment counts as content in its own right — otherwise an attachment-only
+   * message could never be sent before its upload finished, which is the point of the flow.
+   * `failed` and `blocked` attachments still do not count: a message whose only attachment was
+   * rejected must not look sendable.
+   *
+   * Read this instead of `hasSendableData` when
+   * {@link createSendWithPendingUploadsAttachmentsMiddleware} is installed. There is no config
+   * option choosing between the two — the switch belongs to the UI SDK that implements the flow.
```
isekovanic: Because the new middleware keeps the default's `id`, nothing can tell which of the two is installed. So the composer can't pick the getter itself and every UI SDK ends up keeping two switches in sync by hand, which is easy to get wrong I think.

Maybe we can give it its own id and insert it at the same position? Then `hasSendableData` could just do the right thing and this second getter goes away.

src/messageComposer/middleware/attachmentManager/postUpload/attachmentEnrichment.ts
```diff
@@ -4,41 +4,66 @@ import type {
...
+ * This runs when an upload finishes, which can be after the composer was cleared: `uploadFile`
+ * executes this chain before the (by then no-op) `updateAttachment`. If the attachment is gone
+ * from the composer, a message is rendering from the preview instead, and releasing it would
+ * blank that message out. A message can be sent with uploads still running — see
+ * `createSendWithPendingUploadsAttachmentsMiddleware`.
+ *
+ * Omit `composer` and the preview is always released.
+ */
+export const createPostUploadAttachmentEnrichmentMiddleware = (
+  composer?: MessageComposer,
```
isekovanic, line 29: The executor always passes composer now, so the no composer branch is only reachable from tests. Shall we make it required?

src/messageComposer/middleware/messageComposer/attachments.ts
```diff
@@ -15,6 +16,102 @@ const localAttachmentToAttachment = (localAttachment: LocalAttachment) => {
...
+
+  // Composer order is preserved in both payloads so previews do not reshuffle when an upload
+  // settles.
+  const relevantAttachments = composer.attachmentManager.attachments.filter(
+    (attachment) =>
+      isFinishedUpload(attachment) ||
+      (!composerIsKeptAsDraft && isPendingUpload(attachment)),
+  );
+
+  const localAttachments = (state.localMessage.attachments ?? []).concat(
```
isekovanic, line 53: Question about the shape that gets persisted, RN stringifies message attachments straight into SQLite, so this `localMetadata` survives a restart and comes back as uploading with nothing in flight. `normalizeSnapshotAttachment` covers the composer, but nothing covers a stored message.

Is that ours to deal with or maybe we can add a normalizer here? Changes whether we need our own hydration pass (right now the implementation will stay in RN but eventually I would like to integrate it using the new middleware for consistency)

src/messageComposer/middleware/messageComposer/attachments.ts
```diff
@@ -15,6 +16,102 @@ const localAttachmentToAttachment = (localAttachment: LocalAttachment) => {
...
+    relevantAttachments.map((attachment) =>
+      isPendingUpload(attachment) ? attachment : localAttachmentToAttachment(attachment),
+    ),
+  );
+  const messageAttachments = (state.message.attachments ?? []).concat(
+    relevantAttachments.filter(isFinishedUpload).map(localAttachmentToAttachment),
+  );
+
+  if (!localAttachments.length && !messageAttachments.length) return state;
```
isekovanic, line 62: If nothing has finished yet, `messageAttachments` is empty but `localAttachments` isn't, so we fall through and set `message.attachments = []`. The default deliberately avoids introducing that key at all and on an edit an empty array clears the attachments.

It also defeats the new warning so `sanitizeOutgoingAttachments` returns early on `!attachments?.length`, so the "installed the middleware but didn't await the uploads" case is silent. It only fires when something already resolved. Can we set `message.attachments` only when it's non empty?

src/uploadManager.ts
```diff
@@ -126,13 +160,18 @@ export class UploadManager {
...
           id,
           uploadProgress: trackProgress ? 0 : undefined,
         });
 
         const onProgress = trackProgress
           ? (progress?: number) => {
               this.updateUpload({
+                // The record is removed in the `finally` below, so a record that still exists
+                // while reporting 100% can only mean "flushed, awaiting the response".
+                uploadConfirmationPending:
```
isekovanic: If a progress callback arrives with `undefined` after hitting `100`, this flips back to `false` and the UI drops out of indeterminate. We only want to lower it on a number below `100` I think.

src/utils.ts
```diff
@@ -382,6 +383,75 @@ export function unformatMessage(message: LocalMessage): MessageResponse {
...
+
+    if (!localMetadata) {
+      acc.push(attachment);
+      return acc;
+    }
+
+    changed = true;
+
+    // Any of these means the attachment resolved to something the API can store.
+    const hasSource = !!(
```
isekovanic: `hasSource` accepts local URLs, so an attachment carrying both `localMetadata` and `image_url: 'blob:...'` gets its `localMetadata` stripped and the local URL forwarded to the API with no warning. So it would come off as something like `[{"type":"image","image_url":"blob:http://x/y"}]`

That's RN's normal shape (the local file URI is what sits in `image_url`), so the moment RN adopts this the check stops catching anything. I think it needs to be a remote URL check, not just "is there a URL".

src/utils.ts
```diff
@@ -382,6 +383,75 @@ export function unformatMessage(message: LocalMessage): MessageResponse {
...
+    } else {
+      unresolved.push(rest as Attachment);
+    }
+
+    return acc;
+  }, []);
+
+  if (unresolved.length) {
+    channel.getClient().notifications.addWarning({
+      message: `Dropped ${unresolved.length} attachment(s) whose upload never completed. Sending while attachments are still uploading requires a UI SDK that awaits them before sending.`,
```
isekovanic: I'm not sure that the warning notification reflects what should really be shown, this is something that will be shown to users. Maybe it should be cut down to just the first sentence

src/messageComposer/middleware/attachmentManager/postUpload/attachmentEnrichment.ts
MartinCupela: I left it as optional in case anybody uses this factory function to customize the functionality. It is just a precaution and I would change it to required in v10.

src/utils.ts
MartinCupela: We will need to sync in v10, how the asset_url and image_url are used. Seems like in RN they are used for any file-like ref. In React we reference only server resource with it

src/messageComposer/middleware/messageComposer/attachments.ts
MartinCupela: Fixed in [7ab9a3f](https://github.com/GetStream/stream-chat-js/pull/1845/commits/7ab9a3f6c4068209739c48a913ff490fd93c98ef)

src/utils.ts
MartinCupela: Fixed in [c494955](https://github.com/GetStream/stream-chat-js/pull/1845/commits/c494955513e34b8d1b6177dfae8c21fd1f4fe302)

src/uploadManager.ts
MartinCupela: Fixed in [3796544](https://github.com/GetStream/stream-chat-js/pull/1845/commits/3796544e9390c6359dfe472e4a3b1eb0cd0e258a)

src/utils.ts
MartinCupela: I decided just to log the info and prevent this from being user-facing notification: [165c0a1](https://github.com/GetStream/stream-chat-js/pull/1845/commits/165c0a175563238cd330c421135395154309d903)

src/channel.ts
MartinCupela: Fixed for both send and update path in [165c0a1](https://github.com/GetStream/stream-chat-js/pull/1845/commits/165c0a175563238cd330c421135395154309d903)

src/messageComposer/messageComposer.ts
MartinCupela: I think we should not rely on middleware's id to determine whether a feature is enabled. That is fragile as anybody can introduce a middleware with different id, that enables the feature under the hood. As we do not have config parameter on message composer, we cannot rely on a single source of truth for now. Introducing a config param would, given false impression to RN users that enabling it, would change the RN SDK behavior.

I have decided to signal the enablement with adding a flag on installed middleware. That is a temp solution and the flag would be moved to config service in v10. This would not break anything in RN SDK.

MartinCupela: Added with [20e5e96](https://github.com/GetStream/stream-chat-js/pull/1845/commits/20e5e964a7cdd4f18b642cd97f197377aeb253c1)

src/messageComposer/middleware/messageComposer/attachments.ts
MartinCupela: My understanding is that this is currently covered by the RN SDK, so I would defer the introduction into the LLC to a next PR to make sure RN is not broken with changes that could be introduced in this subject.

src/messageComposer/attachmentIdentity.ts
MartinCupela: adjusted here: [e26ea8b](https://github.com/GetStream/stream-chat-js/pull/1845/commits/e26ea8ba01e42dcf4e680f109e69563e3089fd76)

PR #1859, opened 2026-09-10
feat!: establish network connection observer services

MartinCupela:

**The SDK told users "you're offline" when their internet was fine.** The offline banner was
driven by WebSocket health, and a socket dies for reasons unrelated to the network — the
server closes it, a token expires, a keep-alive times out. It also missed the opposite case:
when a phone really loses signal, the socket takes up to 35s to notice.

### Three facts, kept apart

| Question | Answered by |
| --- | --- |
| Does this device have a network? | `client.networkConnection` — **new** |
| Is our WebSocket up, and on which connection id? | `client.wsConnection` |
| Have we finished re-syncing after a reconnect? | `client.connectionRecovery`, which dispatches `connection.recovered` |

Apps can now say "you're offline" for the first network connection loss and "reconnecting…" for the WS connection loss.

### Telling the SDK about the network

One function that subscribes to whatever the platform offers:

```ts
client.config.set({
  client: {
    networkConnection: {
      statusListenerRegistrar: (onStatusChange) => {
        onStatusChange(navigator.onLine); // report the current value immediately
        const handle = () => onStatusChange(navigator.onLine);
        window.addEventListener('online', handle);
        window.addEventListener('offline', handle);
        return () => {
          window.removeEventListener('online', handle);
          window.removeEventListener('offline', handle);
        };
      },
    },
  },
});
```

**Browsers get that registrar for free** — the SDK installs it itself. On React Native, in Node
or during SSR you supply one, and until you do `client.networkConnection.isOnline` is
`undefined` rather than `false`: the SDK won't guess, because a fabricated "online" can't be
told apart from a real reading.

So `isOnline` has three values, and the code deciding whether to render an offline banner has
to handle all three:

```ts
if (client.networkConnection.isOnline === false) {
  // definitely offline — show the banner
}

if (!client.networkConnection.isOnline) {
  // WRONG: `undefined` is falsy too, so this also shows the banner when the status is
  // unknown — a React Native app with no registrar would show "you're offline" forever
}
```

You don't have to supply a registrar at all. The socket still connects and reconnects, messages
still send, queries still run: the SDK never checks network status before doing any of that. It
uses it for one thing only — noticing a dropped connection sooner than the socket's own
35-second keep-alive check would.

### Bugs fixed along the way

Independent of the feature, found while auditing it, each its own commit:

- **Channels were watched against dead connections.** "Watching" a channel means asking the
  server to push that channel's realtime events to this client, and the server ties the
  subscription to a WebSocket by its connection id. `channel.watch()` checked that the client
  had such an id before asking. But the id is set once when the socket connects and is never
  cleared, so during a reconnect the check still saw the previous id and went ahead — subscribing
  over a socket that was already gone. The channel then believed it was watching while the server
  sent it nothing.
- **A backgrounded app came back to a stale thread list.** `ThreadManager` reloaded only if it had
  seen `connection.changed { online: false }` — an event `client.closeConnection()` never
  dispatches, and `closeConnection()` is the path a mobile app takes when backgrounded. It also
  never reset the `lastConnectionDropAt` timestamp it recorded, so after the first drop of a
  session that check stopped gating anything at all.
- **`connection.recovered` fired when nothing had recovered.** Recovery reloads every open channel
  with `Promise.allSettled`, so one failure doesn't stop the rest — but a network drop mid-recovery
  could fail *all* of them, and the event was dispatched regardless. The UI SDKs mark messages read
  when they see it, so they marked messages read that had never been fetched.
- **Replaced sockets were abandoned rather than closed.** `client.openConnection()` overwrote the
  current socket without disconnecting it, so the old one kept its keep-alive timers and its own
  reconnect logic running alongside the new one.

## Breaking changes

Each item appears in a `BREAKING CHANGE:` footer on the commit named.

**`feat!: derive network status from a platform listener, not the WebSocket`**

| Before | Now |
| --- | --- |
| `connection.changed` / `connection.recovered` carried only `online` | both carry `connection: 'network' \| 'ws'` — check it first, or a socket drop on a working network reads as the device going offline |
| `StableWSConnection.isHealthy` | `.isOnline` |
| `client.defaultWSTimeout` | `client.config.set({ client: { wsConnection: { connectTimeoutMs } } })` |
| `WebSocketImpl` / `wsUrlParams` / `wsConnection` client options | `wsConnection` config: `webSocketImpl` / `urlParams` / `connection` |
| `client._getConnectionID()` | `client.wsConnection.connectionID` |

Unlike the fields and options they replace, these survive a reconnect.

**`fix: reload the thread list on reconnect without a self-owned drop flag`**

| Before | Now |
| --- | --- |
| `client.threads.state.lastConnectionDropAt` | `client.wsConnection.state.lastOfflineAt` |

**`feat!: wait for a live socket in channel.watch() and queryChannels()`**

Both now wait for a live socket (up to `connectTimeoutMs`, 15s) instead of returning unwatched
data, and throw if it doesn't come back. The channel then stays unwatched, offline support
renders it locally, and the next reconnect reloads it. An explicit `watch: false` is still
honoured. `client._hasConnectionID()` is removed.

> ⚠️ **If you have tests that mock a connected client, this is the one to watch.** A fixture
> faking a connected user without a live socket will now see `watch()` wait and throw. One line
> to fix — mark the socket up — but it broke 552 tests in `stream-chat-react` first. The error
> names it: `Cannot wait for a WebSocket connection: none has been opened.`

**Behaviour change, no API change:** `connection.recovered` is withheld when the network drops
mid-recovery, so you'll stop seeing it for recoveries that recovered nothing. Unaffected when
no registrar is installed.

Ready to paste into the squash commit message — these are `BREAKING CHANGE:` footers in the
form commitlint and semantic-release parse, verified against this repo's config.

```
BREAKING CHANGE: `connection.changed` and `connection.recovered` now carry a `connection:
'network' | 'ws'` field. Narrow on it before reading `online`, or a socket drop on a working
network reads as the device going offline.

BREAKING CHANGE: `StableWSConnection.isHealthy` is renamed to `isOnline`.

BREAKING CHANGE: `client.defaultWSTimeout` is removed, along with the `WebSocketImpl`,
`wsUrlParams` and `wsConnection` client options. They move to `client.wsConnection.config` as
`connectTimeoutMs`, `webSocketImpl`, `urlParams` and `connection`, set with
`client.config.set({ client: { wsConnection: { … } } })`. Unlike the fields and options they
replace, these survive a reconnect.

BREAKING CHANGE: `client._getConnectionID()` and `client._hasConnectionID()` are removed. Read
`client.wsConnection.connectionID`, or `client.wsConnection.isOnline` where the intent was "is
the connection up" — the connection id is never cleared, so it stayed truthy through a drop.

BREAKING CHANGE: `ThreadManagerState.lastConnectionDropAt` is removed. Read
`client.wsConnection.state.lastOfflineAt`, which is written on every status transition,
including the `disconnect()` path the event is silent about.

BREAKING CHANGE: `channel.watch()` and `client.queryChannels()` now wait for a live WebSocket,
up to `connectTimeoutMs` (15s by default), instead of silently returning unwatched data, and
throw if one does not arrive. The channel is then left unwatched, offline support renders it
from the local database, and the next reconnect reloads it. An explicit `watch: false` from the
caller is still honoured. Test fixtures that fake a connected user without a live socket will
see `watch()` throw `Cannot wait for a WebSocket connection: none has been opened.`

BREAKING CHANGE: `connection.recovered` is no longer dispatched when the device network drops
while a recovery is running, because every reload in it can have failed. Work keyed off that
event will correctly stop running for recoveries that recovered nothing. Unaffected when no
network status registrar is installed.
```


src/connection/ConnectionRecoveryManager.ts
```diff
@@ -121,8 +121,29 @@ export class ConnectionRecoveryManager extends WithSubscriptions {
     if (!this.hasSubscriptions) {
       this.addUnsubscribeFunction(
         this.client.on('connection.changed', (event) => {
+          // Only the socket coming back starts a recovery. The device regaining a network is too
+          // early: there is no reconnected socket and no fresh connection ID yet, so a query fired
+          // then would not establish a watch. See `Channel.reload` / `ChannelManager.recover`.
+          if (event.connection !== 'ws') return;
```
isekovanic: This guard (and its three siblings in `offline_sync_manager`, `WSConnection.ts` and `thread_manager.ts`) silently drops any `connection.changed` that arrives without the new discriminator.

`connection` is required on the type , but `dispatchEvent` applies no default and no validation and it isn't touched by this PR, so nothing backfills it. Any dispatcher ignoring thw type like `client.dispatchEvent({ type: 'connection.changed', online: true })` now gets no recovery and no warning. Before this PR that call worked, because there was no field to omit and the handler was just if `(!event.online) return;`.

Noticed it while porting to the RN SDK through one of our mock builder dispatched without `connection`.

Maybe we can default a missing connection to `ws` in `dispatchEvent`?

src/connection/networkConnection/NetworkConnectionObserver.ts
```diff
@@ -0,0 +1,188 @@
...
+   * derivation entry point every configurable entity exposes, so the client routes a slice here and
+   * knows nothing about this class's defaults or merge semantics.
+   *
+   * Installs whatever registrar the resolved config names, falling back to the platform default. This
+   * is what makes `client.config.set({ client: { connections: { statusListenerRegistrar } } })` work.
+   */
+  public initializeConfig(config?: Partial<NetworkConnectionObserverConfig>) {
+    this.configController.initialize(config);
+    this.setStatusListenerRegistrar(
+      this.config.statusListenerRegistrar ?? getDefaultNetworkStatusListenerRegistrar(),
```
isekovanic: `setStatusListenerRegistrar()` doesn't survive a config derivation I think. Every derivation reruns this line, so an imperatively installed registrar is unsubscribed and replaced by the platform default, which is `undefined` on React Native.

So any imperative  `setStatusListenerRegistrar(r);` then any `config.set()`, `r` gets unsubscribed, config empty. For the declarative `config.set({ client: { networkConnection: { … } } })`  it works fine, which is how I integrated it into the RN SDK.

Same also happens on `disconnectUser()` and then `connectUser()`.

For me it isn't a blocker because I already implemented it with the alternative declarative route, but since it's public API it probably needs to be taken into consideration.

We can either make the setter write through to the `config` so it survives or change the `JSDoc` example to the declarative form and mark the setter as construction time only. 

Or if you're using it on React specifically perhaps we can at least write a disclaimer (or maybe even a warning since we should have this info about which SDK uses it).

src/connection/networkConnection/NetworkConnectionObserver.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /** The current resolved configuration. `Readonly` — change it through {@link updateConfig}. */
+  public get config(): Readonly<NetworkConnectionObserverConfig> {
+    return this.configState.getLatestValue();
+  }
+
+  /** Merges a partial configuration into the resolved config and notifies subscribers. */
+  public updateConfig(config: Partial<NetworkConnectionObserverConfig>) {
+    this.configController.patch(config);
```
isekovanic, line 123: `updateConfig({ statusListenerRegistrar })` writes the value into config and never installs it so the registrar is never invoked and `isOnline` stays `undefined`. Maybe we can use `onChanged` from `ConfigController` `WSConnection` gets away without one because the socket reads config live and this class doesn't.

isekovanic: Or is it the idea to invoke the registrar installation through `initializeConfig` or some other primitives ?

Do we need this at all since we already have `setStatusListenerRegistrar` ?

isekovanic: (this one works fine btw)

src/channel.ts
```diff
@@ -1758,12 +1759,19 @@ export class Channel extends ChannelApi {
...
+    //
+    // This replaces `await client.wsPromise` plus a `watch: false` downgrade, which was wrong both
+    // ways: `wsPromise` is already resolved during a socket-internal reconnect so the wait did not
+    // cover that case, and the downgrade was gated on the connection ID, which is never cleared — so
+    // during a reconnect it did not downgrade and sent `watch: true` against a dead connection.
+    //
+    // If the socket does not come back inside the timeout this throws, which is the correct existing
+    // path: the channel stays unwatched, offline support renders it from the local database, and
+    // `ConnectionRecoveryManager` reloads it on the next reconnect.
+    await waitForWSConnection(this.getClient());
```
isekovanic: This makes sense, but it's unconditional, so it runs before `combined` is built and throws even when the caller passed `watch: false`. We should honour invoking this with `watch: false` since right now it isn't

src/client.ts
```diff
@@ -1369,13 +1379,13 @@ export class StreamChat extends ChatApi {
...
-    if (!this._hasConnectionID()) {
-      defaultOptions.watch = false;
-    }
+    // Same treatment as `channel.watch()`: wait for a live socket rather than degrade, so
+    // `watch: true` always goes out and always binds to a current connection ID.
+    //
+    // The `TODO: probably serverside only thing` that guarded the downgrade is gone with it. There is
+    // no server-side surface in v10 — the constructor takes no `secret` — so there is no client that
+    // can never open a socket, which is what makes waiting safe here rather than a hang.
+    await waitForWSConnection(this);
```
isekovanic: Same here as [this comment](https://github.com/GetStream/stream-chat-js/pull/1859/changes#r4018287563):

```
await client.queryChannels({ filter_conditions: {}, watch: false });
```

would not be honoured

src/connection/wsConnection/StableWSConnection.ts
```diff
@@ -635,29 +709,71 @@ export class StableWSConnection {
...
-      if (this.isHealthy) return;
-      this.client.dispatchEvent({ type: 'connection.changed', online: this.isHealthy });
-    }, 5000);
+      if (this.isOnline) return;
+      this.client.dispatchEvent({
+        type: 'connection.changed',
+        connection: 'ws',
+        online: this.isOnline,
+      });
+    }, WS_OFFLINE_ANNOUNCE_DELAY_MS);
```
isekovanic: Since this timeout has no stored ref and `disconnect()` clears only `healthCheckTimeoutRef`/`connectionCheckTimeoutRef`, I think a a replaced socket still fires it and the closure reads the dead instance's `isOnline`, which stays `false` forever.

So like:

- drop connection 
- `openConnection()` 1s later 
- new socket up 
- at 5s the old one dispatches `connection.changed { connection: 'ws', online: false }` while `client.wsConnection.isOnline === true`, with no `online: true` to follow.

Just storing the timeout ref should be fine and should be cleared on `disconnect()` or any other failure/cleaning mechanisms we might have

isekovanic: PS: I couldn't really reproduce this as I think the WS connection needs to die, then `closeConnection()` needs to be called (i.e we do this on RN when we put the app into background) and then call `openConnection()` immediately  for it to fire wrongly (and of course the WS connection mustn't recover in the meantime)

src/connection/ConnectionRecoveryManager.ts
```diff
@@ -271,12 +297,31 @@ export class ConnectionRecoveryManager extends WithSubscriptions {
...
+    // still reaches the end. Dispatching then would tell consumers that what is on screen is fresh
+    // when none of it was refreshed — and the UI SDKs' mark-read-on-catch-up keys off this event, so
+    // it would mark messages read that were never fetched.
+    //
+    // Withholding is safe rather than stranding: a network drop guarantees a later socket reconnect,
+    // and that recovery dispatches the event. With no registrar installed `lastOfflineAt` stays
+    // `null` and `isOnline` stays `undefined`, so neither condition can fire and this behaves exactly
+    // as it did before the network signal existed.
+    const network = this.client.networkConnection.state.getLatestValue();
+    if (network.lastOfflineAt !== networkDropBefore || network.isOnline === false) {
```
isekovanic: This only covers a network drop. The socket dropping mid recovery is not an uncommon case I guess and this will still dispatch. 

Also since RN has no default registrar installation (and probably some SSR scenarios as well ?), `lastOfflineAt` is `null` and `isOnline` is `undefined`, so neither condition can ever fire. This would I guess happen if only the LLC is used (where even the network conditions would fail) for example. I don't have a good solution for this other than maybe explicitly requiring installation.

Regarding the issue of socket dropping tho, I think we can use `client.wsConnection.state.lastOfflineAt` as it is written on every connection transition and covers both.

isekovanic: Also since the earliest we can install the registrar even on the RN SDK is the `Chat` component, if this is relied on for whatever reason before the `client` is drilled down to the HOC we'd have the same issue (for network conditions)

src/connection/wsConnection/WSConnection.ts
```diff
@@ -0,0 +1,280 @@
...
+    if (this.isOnline === isOnline) return false;
+
+    const now = new Date();
+    this.state.partialNext(
+      isOnline
+        ? // The socket assigns its `connectionID` before announcing it is up, so this is current
+          // here. It is never cleared, which is why going down leaves it alone rather than
+          // pretending to.
+          { isOnline, connectionId, lastOnlineAt: now }
+        : { isOnline, lastOfflineAt: now },
```
isekovanic: `connectionId` is never cleared here and unlike the base branch nothing rebuilds the wrapper and `client.connect()` used to assign a fresh `StableWSConnection`, resetting it to `undefined` every connect. So after `disconnectUser()` the store still holds the previous user's id and sends it as `connection_id` on every request until the next hello.

Maybe we can clear it on every network down transition so that it stays consistent (and let it fix itself after when the connection is recovered) ?

src/utils/waitForWSConnection.ts
```diff
@@ -0,0 +1,119 @@
...
+    timeout = client.wsConnection.config.connectTimeoutMs,
+  }: WaitForWSConnectionOptions = {},
+): Promise<void> => {
+  // Both, not just `isOnline`. The callers need a connection **id** — it is what the server keys a
+  // watch by, and what `api-client` sends as `connection_id` — so resolving on the boolean alone let
+  // a watched query go out before the id existed, and the server rejected it with "Watch or
+  // ChatPresence requires an active websocket connection". The socket now assigns the id before
+  // announcing it is up, so these move together; requiring both means a future reordering surfaces
+  // as a wait that times out rather than as a 400.
+  const isReady = () =>
```
isekovanic: This never consults `client.networkConnection`. With `isOnline === false` this still waits the full `connectTimeoutMs`. The way I understandit, the PR adds the signal that answers it instantly and doesn't read it. On RN we can get away with this by setting offline to the channels we want to open, but still the pending promises will just sit there for some time if we load while offline (or something happens while we are connecting).

Perhaps we can introduce that too here as a short circuit as a WS connection cannot be established if we do not have a network connection ?

src/utils/waitForWSConnection.ts
```diff
@@ -0,0 +1,119 @@
...
+ * `ConnectionRecoveryManager` reloads it on the next reconnect — its `recoverableActiveChannels` is
+ * filtered on `active`, never on `watchStatus`, precisely so a channel that failed to watch is not
+ * abandoned.
+ *
+ * @internal
+ */
+export const waitForWSConnection = (
+  client: StreamChat,
+  {
+    timeout = client.wsConnection.config.connectTimeoutMs,
```
isekovanic: Should probably add support to abort the request too. `queryChannels` for example passes `requestOptions.signal` to the HTTP call but not to this wait, so an aborted search, query or w/e will holds their timer for the full budget.

src/connection/wsConnection/StableWSConnection.ts
```diff
@@ -635,29 +709,71 @@ export class StableWSConnection {
...
+   */
+  private _applyOnline(online: boolean): boolean {
+    if (online === this.isOnline) return false;
+
+    this.isOnline = online;
 
-    this.isHealthy = healthy;
+    // Optional chaining because `options.wsConnection` lets a socket be constructed before its
+    // client exists, and handed one later through `setClient`.
+    this.client?.wsConnection._setStatus({
```
isekovanic: Missing the second optional chain. The "socket built before its client exists" comment says so at least.

src/types.ts
```diff
@@ -238,8 +234,8 @@ type LocalEvent = (
         isLatestMessageSet: boolean;
       };
     })
-  | ({ type: 'connection.changed' } & { online: boolean })
-  | { type: 'connection.recovered' }
+  | ({ type: 'connection.changed' } & { connection: ConnectionType; online: boolean })
+  | ({ type: 'connection.recovered' } & { connection: ConnectionType })
```
isekovanic: Is the discriminator earning its cost here? I'm not so sure and am kind of torn on this. 

`connection.recovered` is dispatched from exactly one place in `ConnectionRecoveryManager.ts`, always 'ws'. The docs say "Do not wait for a 'network' recovery; none is dispatched." So this is a required field with one possible value, forever, that every consumer must narrow on.
The 'network' variant of `connection.changed` has exactly one consumer in the whole LLC in `WSConnection.ts`, forwarding to `_applyNetworkStatus`. That's the client event bus moving a value between two objects on the same `client`, one of which already publishes it as a store (`client.networkConnection.state`) that the other could subscribe to directly. 

For that one internal hop the cost is that every preexisting `connection.changed` handler silently changes meaning.

I am thinking about whether it might be better to introduce a separate event for this as I don't really see any benefit to conflate the two (as they mean something different and are event dependent on each other to some extent).

Why don't we just subscribe to `networkConnection.state` for example where we need it ? That should simplify this weird labelling of what type of "connection has changed" or whatever

src/connection/wsConnection/types.ts
```diff
@@ -0,0 +1,107 @@
...
+export type WSConnectionState = {
+  /**
+   * Is this client's WebSocket up.
+   *
+   * Deliberately the same field name as `client.networkConnection.state`'s — both answer the same
+   * question about a different connection, which is why the two objects are named as parallels. The
+   * types differ only where they must: this one is never `undefined`, because a socket always has a
+   * state, so `!isOnline` is safe here where the network's equivalent needs `=== false`.
+   */
+  isOnline: boolean;
```
isekovanic: Why rename `isHealthy`?  A socket is open or closed and its connection can be healthy or unheatlhy, a socket being "online" feels weird and as if it describes a device. I think this is confusing and we'll need to document and push for it very heavily. 

I think we can either keep `isHealthy` or pick something socket native like `isConnected`, `isOpen`, `isUp` etc etc.

src/connection/networkConnection/types.ts
```diff
@@ -0,0 +1,60 @@
...
+ * ```ts
+ * const netInfoRegistrar: NetworkStatusListenerRegistrar = (onStatusChange) => {
+ *   NetInfo.fetch().then((s) => onStatusChange(!!s.isConnected));
+ *   return NetInfo.addEventListener((s) => onStatusChange(!!s.isConnected));
+ * };
+ *
+ * client.networkConnection.setStatusListenerRegistrar(netInfoRegistrar);
+ * ```
+ */
+export type NetworkStatusListenerRegistrar = (
```
isekovanic: Not really in favor of the "Registrar" name for this.

`(onStatusChange) => Unsubscribe` is basically the `useSyncExternalStore` subscribe signature and the same contract as this package's own `StateStore.subscribe`, which the JSDoc also points out. `Registrar` I understand as a record keeper or something, doesn't feel like it fits. `networkConnection: { subscribe }` says the same thing with no new glossary.

isekovanic (changes requested):

src/connection/networkConnection/NetworkConnectionObserver.ts
```diff
@@ -0,0 +1,188 @@
...
+    });
+  }
+
+  /**
+   * Nothing to subscribe to beyond the registrar, which {@link initializeConfig} installs — this
+   * class has no client events to listen to, by design: deriving either status from an event would
+   * be the conflation it exists to prevent. Present because `WithSubscriptions` requires it, and so
+   * that teardown releases the platform listener.
+   */
+  public registerSubscriptions = () => {
```
isekovanic, line 245: After an `unregisterSubscriptions()`, calling this again readds only the teardown closure and it never reinstalls the registrar, so the observer goes permanently stale. `WSConnection` and `ConnectionRecoveryManager` both resubscribe properly on reregister. Probably harmless for now because nothing actually calls `unregisterSubscriptions()` (nor should it in my opinion, because network connectivity has nothing to do with the other stuff, but still)

src/connection/ConnectionRecoveryManager.ts
isekovanic, line 367: Not related to this PR, but I just noticed it. There is no return after this, so `recover()` logs the warning and then proceeds anyway. Two overlapping recoveries double reload every active channel and race on `isRecovering` (the second one clears it while the first is still running).

src/connection/networkConnection/registrars.ts
```diff
@@ -0,0 +1,46 @@
+import type { NetworkStatusListenerRegistrar } from './types';
+
+/**
+ * The built-in registrar for browsers: `navigator.onLine` plus one `online`/`offline` listener pair.
+ * Installed automatically where {@link getDefaultNetworkStatusListenerRegistrar} can pick it.
+ */
+export const browserNetworkStatusListenerRegistrar: NetworkStatusListenerRegistrar = (
```
isekovanic: This is exported but unguarded, as it dereferences `navigator.onLine` and `window.addEventListener` directly, and only the `getDefault…` wrapper feature detects. Calling the export on a non browser host throws.

Can we guard it please to save ourselves some headaches (even though it's prefixed with `browser`) ?

isekovanic (changes requested):

isekovanic (approved):

PR #1863, opened 2026-09-12
fix: keep query methods from hanging when a connection is closed during connect

### LINKED ISSUE #1122: Race condition: Closing a connection can lead to hanging promises as methods continue to await a web socket which has already been closed
In multiple methods ie `client.search()`, `client.queryChannels()` and `client.queryUsers()`, the client waits for the web socket connection like so

```javascript
await this.wsPromise
```

If `await this.wsPromise` is awaiting and some other code closes the connection at the right time, the client method, eg `client.queryChannels`, will hang forever. The promise neither resolves nor rejects. Even when a new connection is opened it doesn't resolve the original promise because a new promise is created and assigned to `this.wsPromise`. This hanging promise is especially problematic for us, as calls to `queryChannels` are in a queue and if one call doesn't resolve, effectively the whole app is in a bad state, and the list of channels won't update until the user force quits the app.

### Reproduction

[I have created a minimal reproduction demo to demonstrate the problem in action. The demo also includes a proposed fix](https://codesandbox.io/s/stream-chat-js-hanging-promise-l5db8u?file=/src/App.tsx:2299-2352). 

To manually reproduce the bug, these steps must occur:
1. One of the above listed methods, eg `client.queryChannels`, calls `await this.wsPromise` while the websocket is connecting
2. Shortly after `closeConnection()` is called. For example, in our app, we call `closeConnection` whenever the React Native app transitions into background mode. The timing of calling `closeConnection()` is very important to reproduce this race condition. It must be called right after the `new Websocket(url)` is created and before `onmessage` is fired. This is more likely to occur with poor network conditions as it takes longer for the connection to be established. 
3. The `await this.wsPromise` from step 1 is not rejected at this point, instead it continues waiting until there is an active connection. I think this makes sense as a behaviour. The alternative would be to reject the promise. I think it's fine though to keep the promise waiting for an active connection as this will lead to the best DX/UX, IMO.
4. A new connection is opened, eg once the app is reopened
5. Instead of resolving the original promise a new promise is created and the `client.queryChannels` call is left hanging forever. 

### Solution

To keep the spirit of the intended original `await this.wsPromise` code and produce the least code changes to your codebase you can turn `wsPromise` into a getter on `StreamChat`. The getter will construct a new promise and it will resolve whenever there is a healthy websocket connection.

```typescript
  get wsPromise() {
    const isConnected = (this.wsConnection?.isHealthy || this.wsFallback?.isHealthy()) && this._hasConnectionID();
    // Already connected so resolve
    if (isConnected) return Promise.resolve();

    return new Promise<void>((resolve) => {
      const { unsubscribe } = this.on('connection.changed', (event) => {
        if (event.online) {
          unsubscribe();
          resolve();
        }
      });
    });
  }
```

Let me know if you have any questions

cpruijsen:
## CLA

- [x] Code changes are tested

`openConnection` now binds each `connect()` attempt to a single pending `wsPromise`. If a previous connect has not settled, waiters keep that same promise. The latest connect is the one that resolves or rejects it. This is the behaviour the reporter asked for: keep waiting for an active connection rather than rejecting on close, without changing the public `wsPromise` field.

Fixes #1122.

`queryChannels`, `queryUsers`, `search`, `queryReactions`, and the channel `watch` / `query` / `search` paths all `await this.wsPromise` so they do not hit the API before the websocket handshake finishes. If `closeConnection()` ran during that handshake (React Native backgrounding is the reported case) and `openConnection()` later started a new connect, the original `wsPromise` was replaced. Callers still awaiting the old promise never settled, even after the new connection was healthy.

The handshake itself also never settles on close: `StableWSConnection.disconnect()` increments `wsID` before the socket closes, so `onclose` does not reject `connectionOpen`, and `_connect()` waits on that promise with no timeout. This change does not alter that internal handshake.

`connectAnonymousUser()` uses the same `openConnection` path.

`_bindWsPromise` reuses an unsettled `wsPromise` and ignores stale connect generations. Existing `await this.wsPromise` sites (client and channel) do not need to change.

The alternative to reusing the pending promise is the getter on `wsPromise` proposed on the issue, which constructs a new promise and resolves on `connection.changed` with `online: true`. A getter would break `if (!this.wsPromise)` in `userMuteStatus`, change the value `connectUser` / `openConnection` return, and break tests that assign `client.wsPromise = Promise.resolve()`. Reusing the pending promise is the same pattern `openConnection` already uses when `isConnecting` is true, extended across close/reopen.

Fine to switch to the getter if that public shape is preferred.

## Changelog

- Fix a race where `queryChannels` / `queryUsers` / `search` (and other `await this.wsPromise` callers) hung forever if the websocket was closed during connect and then reopened

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -393,6 +393,10 @@ export class StreamChat {
   wsConnection: StableWSConnection | null;
   wsFallback?: WSConnectionFallback;
   wsPromise: ConnectAPIResponse | null;
+  private _wsPromiseSettled = true;
+  private _wsConnectId = 0;
+  private _resolveWsPromise?: (value: Awaited<ConnectAPIResponse>) => void;
+  private _rejectWsPromise?: (reason?: unknown) => void;
   consecutiveFailures: number;
   insightMetrics: InsightMetrics;
   defaultWSTimeoutWithFallback: number;
@@ -852,9 +856,42 @@ export class StreamChat {
     }
 
     this.clientID = `${this.userID}--${randomId()}`;
-    this.wsPromise = this.connect();
+    this.wsPromise = this._bindWsPromise(this.connect());
     this._startCleaning();
     return this.wsPromise;
+  };
+
+  /**
+   * Keep a single pending `wsPromise` across close/reopen so callers that already
+   * `await this.wsPromise` are not stranded when `openConnection` starts a new connect.
+   */
+  private _bindWsPromise = (connectPromise: ConnectAPIResponse): ConnectAPIResponse => {
+    const connectId = ++this._wsConnectId;
+
+    let pending = this.wsPromise;
+    if (this._wsPromiseSettled || !pending) {
+      this._wsPromiseSettled = false;
+      pending = new Promise((resolve, reject) => {
+        this._resolveWsPromise = resolve;
+        this._rejectWsPromise = reject;
+      });
+      this.wsPromise = pending;
+    }
+
+    connectPromise.then(
+      (value) => {
+        if (connectId !== this._wsConnectId) return;
+        this._wsPromiseSettled = true;
+        this._resolveWsPromise?.(value);
+      },
+      (error) => {
+        if (connectId !== this._wsConnectId) return;
+        this._wsPromiseSettled = true;
+        this._rejectWsPromise?.(error);
+      },
+    );
+
+    return pending;
   };
 
   /**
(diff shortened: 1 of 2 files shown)


szuperaz:
Thanks for the PR, since it introduces an issue with consecutive user connect - disconnect calls, I made a PR building on your solution: https://github.com/GetStream/stream-chat-js/pull/1868 - we'll discuss with the team internally, and get back to you.

szuperaz:
Closing in favor of #1868. Thanks again for taking the time to open the PR, your solution is built-in to the final PR.

PR #1864, opened 2026-09-13
fix(errors): export error helpers and types from package entry point

### LINKED ISSUE #1848: Error helpers in errors.ts (isAPIError, isErrorRetryable, APIErrorCodes) are not exported from the package entry point
`src/errors.ts` exports `APIErrorCodes` (with per-code `retryable` flags), `isAPIError`, `isErrorRetryable`, and `isConnectionIDError`, but `src/index.ts` never re-exports the module, so none of them are importable from `stream-chat` (verified on 9.51.0 and current master).

Meanwhile `ErrorFromResponse` from the same module *is* publicly reachable, and `dist/types/errors.d.ts` ships in the package — so TS tooling happily autocompletes the helpers, and the failure only surfaces at build time:

```
Error: Export isAPIError doesn't exist in target module
The export isAPIError was not found in module [project]/node_modules/stream-chat/dist/esm/index.mjs [app-client] (ecmascript).
All exports of the module are statically known (It doesn't have dynamic exports).
```

(Turbopack / Next.js 16, statically analyzed ESM. Runtimes with looser CJS interop resolve the import to `undefined` instead, which is arguably worse since it fails silently.)

**Use case:** implementing an app-level send-retry policy that fails fast on non-retryable API errors (validation, message-too-long, permissions) while retrying retryable ones (rate limit, timeout). The SDK already encodes exactly this knowledge in `APIErrorCodes`, but integrators currently have to copy the table into app code to use it.

Would you accept adding `export * from './errors'` (or a curated subset) to `src/index.ts`?

Tyagiquamar:
Fixes #1848

### Summary
\rrors.ts\ defines key error helpers, types, and error code mappings (\isAPIError\, \isErrorRetryable\, \APIErrorCodes\, \isConnectionIDError\, \isWSFailure\, \isErrorResponse\, \APIError\), but they were not exported from the package entry point (\src/index.ts\).

### Changes
- Added \xport * from './errors';\ to \src/index.ts\ so SDK consumers can import error classification utilities directly from \stream-chat\.


szuperaz (changes requested):
Thanks for the PR, exporting error helpers makes sense. However adding tests that assert something is exported from the client is not something we do, can you please revert the new test?

Tyagiquamar:
Reverted the export assertions test as requested. The PR now only includes the error helper/type exports in src/index.ts.

MartinCupela:
@Tyagiquamar what is that you are missing from the errors.ts file? I believe the contents were not exported on purpose.

Tyagiquamar:
The main missing pieces are `isAPIError`, `isErrorRetryable`, and `APIErrorCodes`, as described in #1848.

The use case is an app-level retry policy: consumers can fail fast on non-retryable API errors and retry things like rate limits/timeouts using the retryability information the SDK already maintains.

Currently these helpers exist in `errors.ts`, but they aren't importable from the `stream-chat` package entry point. If exporting the whole module isn't intended, I'm happy to narrow this to only the helpers/types you consider part of the public API.

szuperaz (approved):
Looks good to me now, lets also wait for @MartinCupela's review

MartinCupela:
It looks like the list of `APIErrorCodes` is incomplete. I will create a new PR that fixes it.

Tyagiquamar:
> It looks like the list of `APIErrorCodes` is incomplete. I will create a new PR that fixes it.

Thanks, that makes sense. I'll leave this open for now so we can revisit the public exports once your `APIErrorCodes` update lands. If you'd prefer this PR to be closed in favor of your change, I'm happy to do that.

MartinCupela:
I have asked for a review by the BE team. The issue is that such a list can get our of sync pretty fast, so there is discussion, whether we could do better. If not, I will provide reviewed updated list and export it.

PR #1865, opened 2026-09-14
fix: offline db hydration race conditions

isekovanic:
## CLA

- [x] Code changes are tested

On a cold start with an offline cache, the channel list queried the wrong page and then sorted itself alphabetically by channel id. I sadly noticed this very late into the evening before heading off to PTO so could not really fix it earlier. The cause are two independent bugs in `ChannelPaginator`, both only reachable when the list mounts before the offline sync has finished - which is the normal case on a slow or flaky connection and is stretched to several seconds whenever there are queued pending tasks to replay.

### 1. The cold-start query asks for the page after the entire cache

`BasePaginator.setItems` records `offset = items.length` plus a query shape, and `ChannelPaginator` seeds the offline cache through exactly that method. A seeded window therefore looks like it has already fetched a page. `offset` is stripped from the query identity, so the next query's shape compares equal, `isFirstPageQuery` returns `false`, and a real `queryChannels` goes out at `offset = <cache size>` so asking the server for whatever follows the whole cache instead of for the list's first page.

The damage compounds since the mid list page merges into a stale cached window, `persistLoadedCids` writes the result back and the next launch seeds more and queries further out. On device the cache went 10 to 20 to 30 channels across three launches, each showing one page deeper.

Fixed in three places, all needed:

- a window that came only from the cache defers its pagination too, not just first pages
- a cache only *continuation* is rewritten into a non-destructive first-page refresh (`keepPreviousItems` + `reset: 'yes'`) **before** the defer decision, so it also applies on the early return path once the sync has completed, otherwise a next page query racing ahead of the deferred refresh still ran at the seeded offset
- the deferred refresh carries the same rewrite, so it cannot inherit `toTail`'s tailward direction

### 2. An empty sort array collapses the list into channel-id order

`get sort()` and the constructor both used `??`, which does not catch `[]`. The React Native SDK's `<ChannelList>` defaults its `sort` prop to `[]`, so the comparator was built with **no sort fields** and only its `getItemId` (cid) tiebreaker. Every comparison tied, and the first `ingestItem` re-sorted the entire list alphabetically by cid. 

`getNextQueryShape` also sent `sort: []` to the server (`[]` is truthy), so the server applied its own default while the client sorted by nothing — client and server ordering disagreed by construction.

An empty array is now treated as unset, like `undefined`. "Sort by nothing" is not something the server can do, and an existing test was already named *"sort **reset** does not reset the paginator state"*, `[]` was I guess always meant as "reset to the default".

I'm not a fan of having defaults at all as it's technically an assumption as to what the server is providing to us. Extending for now since it existed already and also I don't think we can do it unless our backend actually returns the `filters/sort` when we pass none (i.e what the server would use by default otherwise). For predefined filters this will just work, but for explicit ones it will not.

## Changelog

- Fix the channel list querying the wrong page on a cold start when the offline cache is populated
- Fix the channel list re-sorting itself by channel id when no sort is configured


src/pagination/paginators/ChannelPaginator.ts
```diff
@@ -620,18 +627,26 @@ export class ChannelPaginator extends BasePaginator<Channel, ChannelQueryShape>
...
+    const effectiveParams: PaginationQueryParams<ChannelQueryShape> =
+      this._windowIsOfflineCacheOnly && params.reset !== 'yes'
+        ? { ...params, keepPreviousItems: true, reset: 'yes' as const }
+        : params;
+
+    const queryShape = effectiveParams.queryShape ?? this.getNextQueryShape();
     const shouldDeferUntilSynced =
       !!offlineDb?.getChannelsForQuery &&
       !!this.client.user?.id &&
       !offlineDb.syncManager.syncStatus &&
```
MartinCupela: I would like to find a better name for generic `syncStatus` property. Reading the code I have no idea what is its purpose.

src/pagination/paginators/ChannelPaginator.ts
```diff
@@ -264,6 +264,13 @@ export class ChannelPaginator extends BasePaginator<Channel, ChannelQueryShape>
    * `postQueryReconcile` decides whether it should be committed (first page only).
    */
   private _pendingPredefinedFilter: ParsedPredefinedFilterResponse | undefined;
+  /**
+   * True while the loaded window came ONLY from the offline cache - seeded by `setItems`, never
+   * confirmed by a query. Such a window has no pagination position worth continuing from, so the seed
+   * records `offset = <cache size>` and a query shape, which is what makes the next query look like a
+   * continuation rather than a first page. See `executeQuery`.
+   */
+  private _windowIsOfflineCacheOnly = false;
```
MartinCupela: I would maybe change the name a bit to `_activeWindowIsOfflineCacheOnly`. But then this may  also imply there is another window that ins not offline cache only. Maybe we could change the name to make it clear, what is the paginator state.

src/pagination/paginators/ChannelPaginator.ts
isekovanic: sure we can rename it to `isSynced` or `isSyncing` (if we want it the other way around) or something like that

MartinCupela: Let me think a bit

isekovanic: for context, `syncStatus` basically means `isSynced`.

the most correct one would probably be `isSyncing` as a sync can be dropped as well in certain scenarios (meaning `isSynced`, even though serving its purpose is again misleading)

MartinCupela: `isSynced` Sounds better than `syncStatus` I think. Not sure about `isSyncing`

isekovanic: yeah I don't mind `isSynced`, for `isSyncing` we'd need to swap the value obviously as it'd mean something else. my point was, `isSynced` is *technically* not always true as what it really means is `syncingProcessHasConcluded` :D 

`isSyncing` would generally mean "I am currently running the syncing process and will turn to `false` when it finishes, regardless of success"

isekovanic: but I'll just swap it to `isSynced` as the other one anyway I'd do in a separate PR, it touches a lot of sensitive stuff

src/pagination/paginators/ChannelPaginator.ts
isekovanic: we currently don't support loading different windows from the offline DB, just the last one pretty much.

maybe we can simply not mention `window` anywhere and do something like `hasOfflineCacheItems` or whatever if you prefer it that way

MartinCupela: Or `itemsMirrorOfflineDB` / `itemsLoadedFromOfflineDB`?

MartinCupela: or `stateItemsMirrorOfflineDB` / `stateItemsLoadedFromOfflineDB`?

src/pagination/paginators/ChannelPaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1865/commits/1645ee7adf24cfc2f854b4b697629f9c2527e42a

src/pagination/paginators/ChannelPaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1865/changes/0cdaafd9fa7317a02d9146640794240a443fba1a

MartinCupela (approved):

PR #1868, opened 2026-09-14
fix: hanging wsPromise after closeConnection

### LINKED ISSUE #1122: Race condition: Closing a connection can lead to hanging promises as methods continue to await a web socket which has already been closed

szuperaz:
Builds on this PR: https://github.com/GetStream/stream-chat-js/pull/1863

closes #1122 

The original PR solves the issue described by #1122: Calling `client.queryChannels` or any other endpoint that needs `await this.wsPromise` can hang if `client.closeConnection` was called before `wsPromise` was resolved. This is a common use-case in RN where integrators call `closeConnection` when moving to the background (to be able to receive push notifications).

However, PR #1863 introduced a new issue: when calling `client.queryChannels` or any other endpoint that needs `await this.wsPromise`, and doing `disconnectUser` then `connectUser` with different credentials can result the original `queryChannels` being called with the new user's credentials (instead of the request hanging like before):
- `client.connectUser(A)`
- `client.queryChannels()`
- `client.disconnectUser`
- `client.connectUser(B)` -> `queryChannels` will be called with B's credentials

Although we ask integrators not to call `connectUser` before `disconnectUser` finishes, we should still prevent this bug from happening. This PR adds handling for this and rejects `wsPromise` when `disconnectUser` is called before the connection is set up.


src/client.ts
```diff
@@ -852,11 +862,62 @@ export class StreamChat {
...
+   * promise to the new attempt. A `disconnectUser()` drops the user, so nothing will
+   * ever resolve it.
+   */
+  private _rejectPendingWsPromise = (reason: Error) => {
+    if (this._wsPromiseSettled) return;
+
+    // invalidate in-flight attempts so a late settle cannot revive this promise
+    this._wsConnectId += 1;
+    this._wsPromiseSettled = true;
+    this._rejectWsPromise?.(reason);
```
isekovanic, line 885: This rejects a promise that might have no handler on it. `openConnection()` is sometimes used as  fire and forget and that would cause an unhandled rejection after `disconnectUser()`

Maybe something like this would be better ?

```
  this.wsPromise?.catch(() => {});
  this._rejectWsPromise?.(reason);
  this.wsPromise = null;
```

`Null`ing puts `wsPromise` back to what the constructor sets, so `userMuteStatus` and a post disconnect `queryChannels` behave like a fresh client instead of inheriting a rejected promise (and a bunch of other methods too).

src/client.ts
```diff
@@ -713,18 +717,24 @@ export class StreamChat {
...
     } catch (err) {
-      if (this.persistUserOnConnectionFailure) {
-        // cleanup client to allow the user to retry connectUser again
-        this.closeConnection();
-      } else {
-        this.disconnectUser();
+      // disconnectUser() now rejects an in-flight handshake instead of leaving it
+      // pending, so this catch can run *after* a newer connectUser() has taken over.
+      // Cleaning up then would disconnect the connection that replaced ours.
+      if (this.setUserPromise === setUserPromise) {
```
isekovanic, line 731: I think this might break anonymous users. 

`connectAnonymousUser()` doesn't set `setUserPromise` and `disconnectUser()` doesn't clear it. So this check still passes even after an anonymous connect has taken over:

```
client.connectUser({ id: 'A' }, token);  // not awaited
client.disconnectUser();
client.connectAnonymousUser();
```

`A`'s rejection lands for exampel a couple of microtasks later, the check passes, then the catch calls `disconnectUser()` and then the anonymous user is gone and so `client.userID` is `undefined` after. I know this is quite unrealistic, but it is a slight regression from `master`.

So the reject itself is right, but I'd maybe do something like:

```
if (this.setUserPromise === setUserPromise && this.userID === user.id) {
```

This also kills the redundant second `disconnectUser()` when nothing takes over at all.

isekovanic (changes requested):

src/client.ts
```diff
@@ -852,11 +862,62 @@ export class StreamChat {
...
+
+  /**
+   * Keep a single pending `wsPromise` across close/reopen so callers that already
+   * `await this.wsPromise` are not stranded when `openConnection` starts a new connect.
+   */
+  private _bindWsPromise = (connectPromise: ConnectAPIResponse): ConnectAPIResponse => {
+    const connectId = ++this._wsConnectId;
+
+    let pending = this.wsPromise;
+    if (this._wsPromiseSettled || !pending) {
```
isekovanic, line 896: `recoverState()` further down does `this.wsPromise = Promise.resolve()` directly and never touches `_wsPromiseSettled` or `_resolveWsPromise`. That was harmless on `master` because `openConnection` built a fresh promise every call, so the this never went past one call. With this reuse branch it does. You could end up with `wsPromise` already resolved while `_wsPromiseSettled` is still `false` and then the next `openConnection()` hands that resolved promise straight back. Every `await this.wsPromise` clears with no connection behind it. It's basically like an inverse #1122 , requests firing before the handshake instead of hanging after it.

I guess the least "sidequest"-y way to reproduce would be something like:

`StableWSConnection.connect()` calls `this._reconnect({ refreshToken: true })` without `await`ing it on `TOKEN_EXPIRED`, then falls through to `_waitForHealthy()`. So the outer `client.connect()` is still pending while `_reconnect()` runs `_connect()` and then finally `await this.client.recoverState()`. This will basically desync the two.

Pretty much an expired token on first connect would get you there.

Smallest fix, in `recoverState` would be:

```
if (this._wsPromiseSettled) {
  this.wsPromise = Promise.resolve();
}
this.setUserPromise = Promise.resolve();
```

Or resolve through the deferred instead:

```
_wsPromiseSettled = true;
 _resolveWsPromise?.()
```
which is probably more correct since `recoverState` only runs after `_connect()` succeeded. But either should be ok I think.

The fundamental problem imo is that `wsPromise` is public and writable and now has four private fields shadowing it. Maybe worth making `_bindWsPromise` the only writer as this would help.

szuperaz: Fixed in 11c84f4bff1bf2c362fb587891871f6e0aebc49c

src/client.ts
szuperaz: Fixed in 11c84f4bff1bf2c362fb587891871f6e0aebc49c

src/client.ts
szuperaz: Fixed in 11c84f4bff1bf2c362fb587891871f6e0aebc49c

isekovanic (approved):

PR #1870, opened 2026-09-16
feat!: openapi related clean up

szuperaz:
## Breaking changes
- Removed public types (no longer needed): `QueryFilter`, `PrimitiveFilter`, `ExtendedQueryFilter`, `ExtendedQueryFilters`, `ExtendedQueryLogicalOperators` and `Unpacked` - generated filter types should be used instead
- Removed client options: `disableCache`, `wsConnection`, `enableInsights`, `warmUp`
- Behavioral: Previously, if we had no connection id and no pending connection request either, but still called `queryChannels` with `watch: true`, the client just sent the request with `watch` set to `false`. This now throws an error.
- Behavioral: Previously, if during a reconnect we sent a `watch: true` request the request was sent with the old connection id (then most likely rejected if we had no internet connection). These requests now will wait for the reconnect to complete and only send the request once the new connection id is ready.

Detailed description below:

This PR contains various openapi-related clean up tasks. Some of them are technically breaking but shouldn't really impact SDK work or QA tests.

1. `_checkInitialized` discrepancy

I always thought that `_checkInitialized` (checking if channel data is fetched already) method is called before all channel API calls. But that's not really the case; it is called in a few places, mostly from this repo's initial commit, and then never applied to newer api calls. Since it doesn't seem to be used, I've consolidated the checks to the necessary places:
- `_checkInitialized` is called when we rely on data only available after a channel is initialized; it's only `markRead` and `markUnread` where we need to know if read events are enabled. (Although the TS compiler doesn't warn about this, because `configController`'s data is typed as always available, I think in practice it has to be async data we only get once channel data is fetched)
- In other places, a more relaxed check `_ checkHasId` is introduced, which checks if `channel.id` is defined. The use case is that for a distinct channel, we only get the ID after the channel data is fetched. Since `channel.id` is optional, the TS compiler warns us about these places. Technically, the same is true for `channel.cid` but since this field isn't typed as optional, the compiler can't warn us here.

2. Remove dead client options

- `disableCache` - was only relevant to server-side SDKs
- `wsConnection` - replaced by `WebSocketImpl`
- `enableInsights` - not documented, no chat QA tests either
- `warmUp` - not documented, no chat QA tests either

4. Move `await this.wsPromise` to `ApiClient`

We need to call `await this.wsPromise` before an API call that sets `watch`/`presence`/etc flags to `true`. With the generated API layer this would mean to manually wrap these API calls,  and wait for the promise there. Since it's easy to miss a new method when regenerating from open api spec, the logic is moved to `ApiClient` where it automatically works for new endpoints too. The same method is used in stream-feeds-js already: https://github.com/GetStream/stream-feeds-js/blob/main/packages/feeds-client/src/common/ApiClient.ts#L67

Two behavior changes:
- Previously, if we had no connection id and no pending connection request either, but still called `queryChannels` with `watch: true`, the client just sent the request with `watch` set to `false`. This now throws an error.
- Previously, if during a reconnect we sent a `watch: true` request the request was sent with the old connection id (then most likely rejected if we had no internet connection). These requests now will wait for the reconnect to complete and only send the request once the new connection id is ready.

5. Remove type tests + remove any reference to secret

This step was to remove any remaining reference to JS client aceepting API secret, these were mainly docs and unit tests. But it turns out the type tests were also working using the server-side capability of the JS client. Since type tests seem unnecessary with the generated API layer, I removed type tests altogether.

6. Remove hand-written types related to query filters

These were already unused in client.

Linear tickets:
- https://linear.app/stream/issue/REACT-1173/generator-allow-overriding-initalized-check
- https://linear.app/stream/issue/REACT-1174/check-why-we-have-secret-in-client-still
- https://linear.app/stream/issue/REACT-1109/check-queryfilter-type-in-filter-builder
- https://linear.app/stream/issue/REACT-1085/double-check-all-client-options-are-respected-in-the-new-api-client
- https://linear.app/stream/issue/REACT-1083/move-waitforws-methods-to-apiclient


src/types.ts
```diff
@@ -195,12 +176,6 @@ export type StreamChatOptions = {
    * persist even if connectUser call fails.
    */
   persistUserOnConnectionFailure?: boolean;
-  warmUp?: boolean;
-  /**
-   * Sets the instance of `StableWSConnection` on the chat client. Intended purely for testing and
-   * should not be used in production apps.
-   */
```
MartinCupela, line 202: I think this is in conflict with this PR: https://github.com/GetStream/stream-chat-js/pull/1859 where wsConnection property is kept.

MartinCupela: Or is this just about not allowing to provide own `StableWSConnection` implementation?

src/client.ts
```diff
@@ -1290,40 +1280,19 @@ export class StreamChat extends ChatApi {
...
-        });
-      }
+    this.wsConnection = new StableWSConnection({
+      client: this,
     });
+
+    try {
+      return await this.wsConnection.connect(this.defaultWSTimeout);
+    } catch (error) {
+      // The single funnel for a failed *initial* connect. `_reconnect()` failures deliberately do
```
isekovanic: I think this comment's invariant doesn't hold here.

`StableWSConnection.connect()` ends with `await this._waitForHealthy(timeout)`, whose 15s race arm
throws `'initial WS connection could not be established'`  and it throws **while** the
`_reconnect()` loop kicked off by `onclose`/`onerror` is still retrying. So a live reconnect does
land in this catch, which is exactly the case the comment says is excluded.

I think there are 2 consequences here:

The first one happens every time. Once `rejectConnectionId` runs, `settle()` clears `loadConnectionIdPromise` and the only place that calls `arm()` is `_connect()`. So during each await `sleep(interval)` in the retry loop there is no id and no deferred and so `getConnectionId()` throws "there is no WebSocket connection and none is being established". The backoff goes up to 25s, so the windows are long enough to hit. Easiest way in is calling `client.openConnection()` directly, which is what we tell integrators to do on mobile foreground for example. `connectUser()` hides it, because its catch tears the socket down anyway.

The second one is rarer but worse imo. If the reconnect resolves an id just before this catch runs, we clear a valid id on a healthy socket. Nothing rearms after that and so `invalidate()` returns early when there's no id and `_setHealth(false)` doesn't fire because the socket is fine. You end up with a healthy WS where every watch/presence request throws and it stays that way until the next connection transition.

Simplest fix is probably to only reject here when the connection is genuinely dead (nothing
pending) ? And maybe to make `rejectConnectionId` non destructive to an already resolved id. Arming across
the whole `_reconnect` loop rather than only inside `_connect()` would also close the first point I guess but not sure which approach is better.

src/connection_id_manager.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.resolve?.(connectionId);
+    this.settle();
+  };
+
+  /**
+   * Fails anything waiting on a connection id, for a connection attempt that will not be retried.
+   *
+   * @param reason - The error to reject the pending waiters with.
+   */
+  rejectConnectionId = (reason: unknown) => {
```
isekovanic, line 61: `this.connectionId = undefined` is unconditional, which makes this method unsafe to call from any
late or superseded path.

A reject that arrives after a *different*, successful attempt has already published an id will
clear that live id and because `settle()` also drops the deferred there's nothing left for a waiter
to attach to. `invalidate()` deliberately guards itself with `if (!this.connectionId) return;` so this one has the mirror-image problem.

Maybe we can only clear when the rejection actually corresponds to the attempt that's outstanding,
i.e bail early if there's no pending deferred to reject:

```ts
rejectConnectionId = (reason: unknown) => {
  if (!this.reject) return; 
  this.connectionId = undefined;
  this.reject(reason);
  this.settle();
};
```

That keeps the `reset()` flow working (it always has a deferred armed or nothing to lose since no reject) while
making a stale reject a noop.

src/connection.ts
```diff
@@ -652,6 +639,10 @@ export class StableWSConnection {
     // The server keys channel watches by connection ID, so they are gone the moment the socket is.
     // Done here rather than off the `connection.changed` event below, which is debounced by 5s.
     this.client._markActiveChannelsWatchInterrupted();
+    // Same fact, same moment, seen from the other side: the id those watches were keyed by is dead,
+    // so no further request may be sent with it. Invalidating arms a fresh deferred, so requests
+    // needing an id wait for the reconnect instead of racing ahead with the dead one.
+    this.client.connectionIdManager.invalidate();
```
isekovanic, line 642: I think this introduces a regression. If no id has been resolved and there's no in-flight `connect()` to funnel the
rejection, the deferred armed in `_connect()` is never settled and gated requests hang for the life
of the client. There was nothing to hang on before.

The cause is the `WS_CLOSED_SUCCESS` branch of `onclose`. It's the one close path that reaches neither `_setHealth(false)` nor `_reconnect()` nor the manager. It only fires for a server initiated (token revoked, user deleted, whatever), since our own
`disconnect()` bumps `wsID` first. `this.client.connectionIdManager.reset()` there fixes it.

Also, another thing I noticed is that since we leave the id in place it can be handed to a dead socket. This isn't introduced in this PR as v9 does exactly the same (`connectionID` assigned, then cleared
nowhere), so it's old.

I'm not sure how it can be fixed as it is kinda needed for offline support.

src/api-client.ts
```diff
@@ -171,6 +169,14 @@ export class ApiClient {
     data?: unknown | null,
     additionalConfig: AxiosRequestConfig = {},
   ): Promise<{ body: T; metadata: RequestMetadata }> {
+    if (requiresConnectionId(additionalConfig.params, data)) {
+      const connectionId = await this.client.connectionIdManager.getConnectionId();
```
isekovanic, line 173: Two things.

The await ignores `additionalConfig.signal`, already populated by `toAxiosRequestConfig` in
`sendRequest`. So an aborted request stays parked for the whole outage and only fails afterwards.
Cancellation is the only escape hatch from an unbounded wait, so it should work here.

src/api-client.ts
```diff
@@ -171,6 +169,14 @@ export class ApiClient {
...
+      additionalConfig = {
+        ...additionalConfig,
+        params: { ...additionalConfig.params, connection_id: connectionId },
```
isekovanic, line 176: Maybe spread last, so this overrides a caller supplied `connection_id`. 

I don't know if this is something that happens often either but it's how it worked on V9, so unless we have a good reason to do it the other way around I would keep it.

src/api-client.ts
```diff
@@ -135,8 +135,6 @@ export class ApiClient {
         // that updates existing axios instance options instead
         ...this.client.options.axiosRequestConfig?.params,
         ...additonalConfig.params,
-        connection_id:
```
isekovanic, line 138: Only some spec declared endpoints can carry a `connection_id` now and of those only the ones
`requiresConnectionId` fires for. 

The one I'm not sure about is our
`client.sync({ channel_cids, last_sync_at })` with no `watch` in `OfflineDBSyncManager` as the generator always emits the `watch` key, so
the gate returns false and `/sync` now goes out with no `connection_id` at all, right ? Or am I looking at the wrong thing here ?

Before this was always injected by default but now I'm not sure whether it would be enough

src/types.ts
MartinCupela: I will adapt my PR to yours

src/api-client.ts
```diff
@@ -249,6 +255,47 @@ export class ApiClient {
...
+  // For those cases lets first check if the request has watch/presence flag, and return false...
+  if (
+    [params, payload, requestBody].some(
+      (source) => source && ('watch' in source || 'presence' in source),
+    )
+  ) {
+    return false;
+  }
+
+  // ...return true if the request doesn't have watch/presence flag but has connection_id
```
isekovanic, line 295: This fallback makes `stopWatchingChannel` always gated abd it declares no flag so it falls through to
the `connection_id` within the params check every time.

After `closeConnection()` the manager has been `reset()`, so `getConnectionId()` throws and
`channel.stopWatching()` rejects. That's how RN handles background/foreground for example, so a `stopWatching()` in an unmount handler now throws, where v9 sent the request with `connection_id` undefined. And since
the `channel` only sets `watchStatus` on success, the status stays `Watching` for a watch that's definitely gone.

Maybe we can either short circuit to a local only update when there's no id or keep the throw and document it.

isekovanic (changes requested):

src/api-client.ts
szuperaz: This is on purpose, because really there is no use-case for an integrator to provide a connection_id. This logic is highly internal to the Stream API, so no reason for an integrator-provided value to ever override this. I don't think integrators would ever want to do this either, but I think for fixed headers/params JS client's value should always win.

src/api-client.ts
szuperaz: Yes, this is true, but it was true before my changes, because `await this.wsPromise` is always called before we send the given API request, like here: https://github.com/GetStream/stream-chat-js/blob/master/src/client.ts#L1933

But I was also thinking about how it should work, I guess the alternative is to race the abort signal and wsPromise? In the end the reason I kept the current logic, is for example, retrieving the API token is also not counted toward the signal timeout, and I think it's sort of the same logic.

src/client.ts
szuperaz: > The second one is rarer but worse imo. If the reconnect resolves an id just before this catch runs, we clear a valid id on a healthy socket.

I don't think this can happen, because the catch is called before reconnect starts a new connection because of the mandatory sleep we have before a reconnect.

> The first one happens every time. Once rejectConnectionId runs, settle() clears loadConnectionIdPromise and the only place that calls arm() is _connect(). So during each await sleep(interval) in the retry loop there is no id and no deferred and so getConnectionId() throws "there is no WebSocket connection and none is being established". 

Yes this is true, but in that same time we also reject the `wsPromise` too with `initial WS connection could not be established`, so this logic mirrors that logic, and should introduced any behavior change compared to when we waited for `wsPromise` instead of `connectionId` promise

src/connection_id_manager.ts
szuperaz: Techically this shouldn't happen, but yeah, we can add that guard anyway, I'll fix

src/connection.ts
szuperaz: > If no id has been resolved and there's no in-flight connect() to funnel the
rejection, the deferred armed in _connect() is never settled and gated requests hang for the life
of the client. There was nothing to hang on before.

It's hanging because client always starts a reconnect, either via onclose/onerror event, or when the browser sends an online event after we were online. I think it's technically correct, but it it messes with offline, we can change it, but I don't understand the issue.

szuperaz: > Also, another thing I noticed is that since we leave the id in place it can be handed to a dead socket. This isn't introduced in this PR as v9 does exactly the same (connectionID assigned, then cleared
nowhere), so it's old.

This only time we should not clear id if the "The cause is the WS_CLOSED_SUCCESS branch of onclose. It's the one close path that reaches neither _setHealth(false) nor _reconnect() nor the manager. It only fires for a server initiated (token revoked, user deleted, whatever), since our own" since it's no-op in the other parts of the client, I didn't do anything to connection id either.

src/api-client.ts
szuperaz: Yes, this is a change, but since `watch` is always false on server-side, sending `connection_id` without `watch` doesn't do anything.

src/api-client.ts
szuperaz: I'm maybe not getting the issue right, but `closeConnection` calls `_markActiveChannelsWatchInterrupted`, so that will set the `watch` flag to `false` regardless of `stopWatching`. 

In general `stopWatching` throwing when called without a `connection_id` is the correct behavior I think. However, I guess if we're offilne, we can skip the throw, because if WS connection is closed, watches are gone anyway. Lets talk about this, I'd like to understand the issue better.

src/api-client.ts
szuperaz: Double-checking if server reads it regardless of watch

szuperaz: Double-checked and `connection_id` only does anything if paired with `watch: true`

src/api-client.ts
szuperaz: This is tricky due to types: since `client` inherits from `ChatApi` I can't change the method return type. So in order to not throw, I'd have to fake a backend response, with metadata like request headers, which I'm not sure that worth the DX gain

isekovanic (approved):

PR #1875, opened 2026-09-17
feat: message pruning

isekovanic:
## CLA

- [x] Code changes are tested

A livestream channel that never stops receiving messages grows its loaded window forever. Every publish reprojects the active interval, so that projection is `O(loaded)` and the cost of each event climbs with the size of the window. The `stateThrottleMs` throttle already cut how *often* we publish; this cuts how *much* each publish costs. They compose, neither replaces the other. Additionally, on platforms like RN this leads to a ton of engaged native views, even if they are not technically rendered/mounted.

`BasePaginator.pruneTailToLimit(interval)` drops the oldest ids until `maxLoadedItems` remain. The whole thing hinges on *where* it is called from, so `ingestItem` hands it the interval that `insertItemIdIntoInterval` just returned, which is a **fresh copy that has not been committed yet**. So it mutates something nothing can observe, and the `commitInterval` a few lines down stores and publishes the already pruned interval. The window emit after that projects from the same interval. A prune therefore costs no publish of its own and it rides the ones the ingest was going to make anyway and `state.items` never even briefly exceeds the cap.

The pagination half is the part worth reviewing. Dropping the oldest messages re-opens the tailward edge, so `hasMoreTail` and `cursor.tailward` have to move with it, and publishing those separately would be exactly the second notification this design exists to avoid. They ride the window publish instead. But in production that publish is deferred and `MessagePaginator` defaults `stateThrottleMs: 500`, so `ingestItem` schedules and returns, and the actual `partialNext` happens later in `flushWindowPublish`. We need to derive this from `_prunedIntervalId` at commit time, as otherwise due to throttling it might simply lag behind or just be plain wrong.

That is really just making the prune behave like everything else here: the throttle is not a queue, nothing is ever handed to it, and `flushWindowPublish` rereads `_itemIntervals` when it fires. A cached value was the only thing in that publish that could go stale, which is why it was the only thing that did.

We also add `maxLoadedItems` on `PaginatorOptions`, modeled around `stateThrottleMs`, so it resolves through the existing layering and can be set per instance (`channel.messagePaginator.updateConfig({ maxLoadedItems: 100 })`), at construction, or declaratively — shared `messagePaginator` key with `channel` / `thread` overrides, so a thread can cap differently from the channel list.

**`setPruningSuspended(bool)`** is the UI's "hold off" trigger. A UI that knows the user is reading near the oldest loaded message says so and the window is allowed to grow past its cap until they scroll back. Essentially it's the way our UI can hold off pruning (for example if we're at the top of a list or something). 

Only an anchored interval that is both the dataset head and the active one, so a jumped away window stays stable and a logical (live-only) window, which has no pagination provenance and could never be re-fetched, is declined. Messages the server does not know about are **skipped, not stopped at** so that a failed send sorts by the time it was composed, so a positional cap would destroy it. Skipping leaves the window a couple of items over at worst and the next arrival trims it again. The cursor comes from the oldest *server-confirmed* id for the same reason, otherwise a client-generated id goes out as `id_lt`. But this is better than letting these die out.

Additionally, every id leaving the interval goes through `_itemIndex.remove`, which is `memberIds.delete` plus `store.unlink`, and the store drops content only when the last subscriber lets go. A pruned message still held by an open thread or the pinned list keeps its content and the offline DB is never touched deliberately.

## Changelog

- Added `maxLoadedItems` to bound a paginator's loaded window, settable per instance, at construction, or declaratively per surface
- Added `MessageIntervalPaginator.setPruningSuspended` so a consumer can hold pruning off while the user reads near the oldest loaded message


src/pagination/paginators/BasePaginator.ts
```diff
@@ -533,18 +533,12 @@ export abstract class BasePaginator<T, Q> {
...
-   * `state.items` publish also carries the pagination fields that move with it instead of emitting
-   * one of their own — publishing those separately would cost a second notification per prune, the
-   * one thing the window cap must not do.
-   *
-   * A flag rather than the values themselves, on purpose. The same fields are written by the query
-   * path ({@link postQueryReconcile}), so a cached snapshot draining after a `toTail()` that landed
-   * inside the same throttle window would republish a pre-merge cursor and re-fetch the page that
-   * just merged. A flag cannot go stale that way: {@link takePrunedPaginationFields} re-derives from
-   * the committed interval at publish time, so whoever publishes last publishes the truth.
+   * The interval whose tailward edge a prune ({@link pruneTailToLimit}) moved inward, so the next
```
MartinCupela: Hard to read for a human. Could it be re-written for humans with context and less abstract / condensed manner?

src/pagination/paginators/BasePaginator.ts
MartinCupela, line 1101: The comment says "so its id is a usable cursor" - maybe we should be checking for the existence of the item's id too? `!!item?.id`

MartinCupela: This comment is hard to understand. What does "query landing between the prune and this publish" mean? Why does it mention tailward query? How can be "reached the dataset start again"? Which dataset? Start is the head, right?

isekovanic: Updated here: https://github.com/GetStream/stream-chat-js/pull/1875/changes/a3027974967925b9f41d68d24b1acb9794176786. The check isn't particularly validating the id but rather validating whether this item exists pretty much.

Since we can't just check the `id` since `item` is a generic, we do it downstream. I don't think it'll hold much value here, but if we want to make sure it exists (for a general check perhaps, even though the id is not a cursor for anything other than messages I guess) we can do something like `getItemId` since it is `(item as { id: string }).id` but its `string` return type is a lie, though.

src/pagination/paginators/BasePaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1875/changes/93ac0a3eb5d46271412cf7653665ceaba6e98d45

src/pagination/paginators/BasePaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1875/changes/4decfb649e0d9f0c790759ef932b9e8ad6e9f3f1

src/pagination/paginators/BasePaginator.ts
```diff
@@ -1245,25 +1240,30 @@ export abstract class BasePaginator<T, Q> {
...
+   * Three things it deliberately leaves alone:
+   * - a window that is no longer the pruned one. `state` tracks the active interval, so once the
+   *   consumer has jumped elsewhere the prune says nothing about what is being published.
    * - `cursor.headward`, which the prune never touched — carried over verbatim rather than re-derived,
    *   so a `config.deriveCursor` hook's verdict on the head edge survives.
    * - a null tail edge (a window whose every remaining item is locally-created). Publishing `null`
    *   would read as "tailward exhausted"; the existing cursor still names a message the *server* has,
    *   so leaving it in place keeps back-pagination working.
    */
   private takePrunedPaginationFields(): Partial<PaginatorState<T>> {
```
MartinCupela: I did not understand what the method was doing, from the name. If I understand it correctly it ensured correct pagination fields for active pruned interval?

isekovanic: eh yeah the name is probably not great, let me think of something better

MartinCupela (approved):

src/pagination/paginators/BasePaginator.ts
isekovanic: https://github.com/GetStream/stream-chat-js/pull/1875/commits/fff84e02e0a63627b6a42a7e14ae1dab8e07c996

this should be a bit clearer.

It basically doesn't ensure of them but consumes them rather (i.e it t returns the state changes, so `hasMoreTail` and `cursor`). They are them committed to the state on the next publish, when the prune is committed as well so as to avoid 2 state updates of the same thing

